PR #2, opened 2019-08-08
implemented Debug for Tag to print in hex

ibaryshnikov:
Motivation: provide useful output for searching tag in dictionary.
Before `println!("{:?}", tag)` would print `Tag(32736, 16)`, after: `Tag(0x7FE0, 0x0010)`, and now it's easy to find `E { tag: Tag(0x7FE0, 0x0010), alias: "PixelData"`

### DIFF
--- a/core/src/header.rs
+++ b/core/src/header.rs
@@ -540,7 +540,7 @@ pub type ElementNumber = u16;
 /// a (group, element) pair. For this purpose, `Tag` also provides a method
 /// for converting it to a tuple. Both `(u16, u16)` and `[u16; 2]` can be
 /// efficiently converted to this type as well.
-#[derive(Debug, PartialEq, Eq, Hash, PartialOrd, Ord, Clone, Copy)]
+#[derive(PartialEq, Eq, Hash, PartialOrd, Ord, Clone, Copy)]
 pub struct Tag(pub GroupNumber, pub ElementNumber);
 
 impl Tag {
@@ -554,6 +554,12 @@ impl Tag {
     #[inline]
     pub fn element(&self) -> ElementNumber {
         self.1
+    }
+}
+
+impl fmt::Debug for Tag {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Tag(0x{:04X}, 0x{:04X})", self.0, self.1)
     }
 }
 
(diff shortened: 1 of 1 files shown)


Enet4:
I wasn't quite expecting to receive PRs at this time, but this one is isolated enough for a quick review.

One can add those formatting flags at the upper level, to obtain a similar outcome:

```rust
println!("{:04X?}", tag); \\ Tag(7FE0, 0010)
```

What we can't do without a custom Debug implementation is to pretty-print the two components of the tag without pretty-printing the whole struct, so it doesn't add the `0x` prefix without it. I also recall that the `Display` impl already provides a familiar formatting for tags: `(GGGG, EEEE)`.

With this in mind, I'm reluctant to make this change, but would like to hear from you whether the methods above solve your particular concern. The standard dictionary should also be searchable at run-time by tag, so an exact textual match might not be needed.

ibaryshnikov:
I did a lot of debugging recently while opening different files and found this addition quite useful. I don't insist of course, I can still patch it locally.
Usually I just do `dbg!()`. I like the idea with `println!("{:04X?}, tag")`, I'll try it and submit a feedback later. At first glance in playground:
```rust
#[derive(Debug)]
struct Tag(u16, u16);

#[derive(Debug)]
struct Data {
    tag: Tag,
    value: i32,
}

fn main() {
    let tag = Tag(0xFE00, 0xFFFF);
    println!("{:X?}", tag); // Tag(FE00, FFFF)
    let data = Data { tag, value: 123 }; // I'd like to see this `value` as a decimal
    println!("{:?}", data); // Data { tag: Tag(65024, 65535), value: 123 }
    println!("{:X?}", data); // Data { tag: Tag(FE00, FFFF), value: 7B }
    dbg!(&data);
    // [src/main.rs:24] &data = Data {
    // tag: Tag(
    //     65024,
    //     65535,
    // ),
    // value: 123,
}
```

core/src/header.rs
```diff
@@ -557,6 +557,12 @@ impl Tag {
     }
 }
 
+impl fmt::Debug for Tag {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Tag(0x{:04X}, 0x{:04X})", self.0, self.1)
```
Enet4: We can take advantage of pretty-printing (`#`) to add the prefixes:

```suggestion
        write!(f, "Tag({:#06X?}, {:#06X?})", self.0, self.1)
```

ibaryshnikov:
I'm happy to hear that! And your suggestion looks clearer

Issue #11, opened 2019-09-13
Handling empty date


ibaryshnikov:
Sometimes the date in DICOM file is `0000-00-00`. Is it a common case in the ecosystem when the date is empty?

neandrake:
It is fairly common for date elements to either be all zeros, empty, or not present altogether. Including SeriesDate / StudyDate.

pauldotknopf:
This kind of relates to #9. We shouldn't attempt to parse the VR's while loading the DICOM files.

IMO, all parsing should be done after the fact, for example, using ```Result<[u8]>``` for ```VR::IS```, or ```Result<DateTime>``` for ```VR::DT```.

@Enet4, what do you think about making this the "default" approach?

Enet4:
As mentioned in #9, it is reasonable to preserve textual values as strings when loading DICOM objects and only convert when requested. I am not opposed to making this the default, especially since some equipment and software out there have the poor habit of writing values which may not 100% DICOM compliant.

Regarding this issue (and others of similar nature), it helps if you can make a reproducible example, so that it can be independently reasoned with and tested. In part, this also means that there should be an easy way to build and write DICOM objects, yet to be worked on.

Enet4:
Closing this as partially resolved by #24, and for discussion to continue in #9.

Issue #9, opened 2019-09-13
Interpreting string-based values?


pauldotknopf:
I have a question about some concerns I have about pre-interpreting data of certain VR types.

For example, the ```VR::IS``` is being interpreted as an ```[i32]``` internally.

Because of this, if we were to load a dicom file, and save it without making any modifications, the saved file would be normalized, in that the ```VR::IS``` would be converted into an array of integers, and then back to a string.

This would also pose a problem with implementing validation checkers. I'd like to develop a command line tool that loads a DICOM file, and validates it's contents. I can't validate any field at all if any one field prevents me from loading the entire DICOM file.

IMO, it is important to keep values that are represented with the ```Default Character Repertoire``` as characters. We could provide utilities to re-interpret the values into their typed representation (date times, date ranges, integer strings, etc) after the DICOM file is loaded.

Enet4:
If I understand correctly, it is a planned feature to support loading a file without parsing textual values right away. There's already some code for this string-preserving parsing mechanism, although it might not be accessible at the object API, and this might also require a bit of extra code at the element struct to only parse these values on demand. At some point this should be possible indeed.

PR #12, opened 2019-09-14
Default to not interpreting string-based values.

pauldotknopf:
Addresses #11 and #9.

### DIFF
--- a/parser/src/dataset.rs
+++ b/parser/src/dataset.rs
@@ -35,6 +35,8 @@ pub struct DataSetReader<S, P, D> {
     hard_break: bool,
     /// last decoded header
     last_header: Option<DataElementHeader>,
+    /// Optionally interpret string-based values (DT, IS, etc)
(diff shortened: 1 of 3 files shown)


parser/src/dataset.rs
```diff
@@ -104,6 +114,7 @@ where
             in_sequence: false,
             hard_break: false,
             last_header: None,
+            interpret_string_values: true,
```
Enet4: I noticed that the default is `true` at this level, but `false` at the object abstraction. Is this intentional? This behavior should also be documented.

pauldotknopf: That's correct. That should be ```false```. I will document.

pauldotknopf:
Given our recent discussions, I think the new ```AttributeEntry``` would be perfect to contain the typed values.

We can close this merge request, and when I get to implementing ```AttributeEntry```, I'll make sure string-based values are representing as strings by default.

PR #13, opened 2019-09-15
Running clippy.

pauldotknopf:
Should we add these checks to the build server?


Enet4:
These look good, thank you!

Adding Clippy to CI could be done, but not without extra care, since not all toolchains may have it successfully built and available. As a noteworthy example, the Rustup `clippy` tool hasn't been available since `nightly-2019-09-14` (see [Rustup components history](https://rust-lang.github.io/rustup-components-history/)). We can include it, but a successful installation should not be mandatory.

Issue #6, opened 2019-09-09
WebAssembly support


ibaryshnikov:
Is it a goal or non-goal?
It worked out of the box before 0.1, but not supported by inventory yet.

Enet4:
Thank you for reporting. WebAssembly support is indeed planned, but it did not occur to me that the flexible transfer syntax registry would break compatibility.

If `inventory` really can't work in a wasm environment, we'll have to look for an alternate implementation.

ibaryshnikov:
Thanks for a quick reply!

There's a related issue https://github.com/amethyst/amethyst/issues/1893, but `linkme` support for wasm is not ready yet https://github.com/dtolnay/linkme/issues/6

From the other side, it's possible to have mutable statics

```rust
lazy_static! {
    static ref REGISTRY: Mutex<TransferSyntaxRegistry> = {
        Mutex::new(TransferSyntaxRegistry { m: initialize_codecs() })
    };
}
```

the drawback is that it can affect performance.

One more option is to use decoder builder, keeping a local copy of transfer syntax registry per decoder

```rust
let decoder = DecoderBuilder::default()
    .with_transfer_syntax(extend it)
    .with_transfer_syntax(extend again)
    .build()?;
let object = decoder.from_reader(input)?;
```

What do you think? May be there is a better alternative which I missed.

Enet4:
Considering that #17 has been merged, we can consider this resolved. We should strive to keep WebAssembly as a working target, although there is much to be done regarding testing. Please create a new issue if another problem emerges when using DICOM-rs in WebAssembly.

PR #17, opened 2019-10-06
Transfer Syntax index abstraction & inventory-less registry

Enet4:
This PR introduces a trait for transfer syntax dictionaries and makes it so that the main TS registry can work even without support for the `inventory` based registry. The inventory will be behind the Cargo feature `inventory-registry`, enabled by default only at the parent crate `dicom` and in end user tools such as `dcmdump`. If this is disabled, built-in TSes will still be available, but the registry will be unable to be extended. <s>We can think of ways to overcome this issue in future contributions.</s> This can be overcome by creating a new registry implementation.

This should address the main concern presented in #6, although I would be grateful to have this tested in that environment (CC @ibaryshnikov).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,4 +20,8 @@ script:
   - if [ "$CLIPPY_INSTALLED" == 1 ]; then
       cargo clippy;
     fi
+  # default features
   - cargo test --all
+  # no inventory
+  - cargo test --manifest-path encoding/Cargo.toml --no-default-features
+  - cargo test --manifest-path transfer-syntax-registry/Cargo.toml --no-default-features
--- a/dcmdump/Cargo.toml
+++ b/dcmdump/Cargo.toml
@@ -9,5 +9,8 @@ repository = "https://github.com/Enet4/dicom-rs"
 categories = ["command-line-utilities"]
 keywords = ["cli", "dicom", "dump"]
 
+[features]
+default = ['dicom/inventory-registry']
+
 [dependencies]
-dicom = { path = "../parent/", version = "0.1.0" }
+dicom = { path = "../parent/", version = "0.1.0", default-features = false }
--- a/encoding/Cargo.toml
+++ b/encoding/Cargo.toml
@@ -9,10 +9,14 @@ repository = "https://github.com/Enet4/dicom-rs"
 categories = ["encoding"]
 keywords = ["dicom"]
 
+[features]
+default = []
+inventory-registry = ['inventory']
+
 [dependencies]
 dicom-core = { path = "../core", version = "0.1.0" }
 dicom-dictionary-std = { path = "../dictionary-std", version = "0.1.0" }
 quick-error = "1.2.2"
 encoding = "0.2.33"
 byteordered = "0.4.0"
-inventory = "0.1.4"
+inventory = { version = "0.1.4", optional = true }
--- a/encoding/src/transfer_syntax/mod.rs
+++ b/encoding/src/transfer_syntax/mod.rs
@@ -3,8 +3,11 @@
 //! algorithms for decoding and encoding DICOM data in a certain transfer syntax.
 //!
 //! This crate does not host specific transfer syntaxes. Instead, they are created in
-//! other crates and registered in the global transfer syntax registry. For more
+//! other crates and registered in the global transfer syntax registry, which implements
+//! [`TransferSyntaxIndex`]. For more
 //! information, please see the `dicom-transfer-syntax-registry` crate.
+//!
+//! [`TransferSyntaxIndex`]: ./trait.TransferSyntaxIndex.html
 
 pub mod explicit_be;
 pub mod explicit_le;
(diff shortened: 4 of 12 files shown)


ibaryshnikov:
That's awesome, thanks for cc'ing, I'll be happy to test! I'm updating my prototype to fully use the code from this PR. I have one blocker though, `DataElement` at the moment only allows
```rust
    /// Retrieve the data value.
    pub fn value(&self) -> &Value<I> {
        &self.value
    }
```
Is it possible to add `pub fn into_value(self) -> Value<I> {` ? I need it to convert `PrimitiveValue` of `PixelData` into `Vec<u16>` in place. It's important, because inside a browser there's a 2Gb memory limit. As a `.dcm` file size is around 500Mb, it's easy to get out of memory error with any additional copy, especially after copying pixel data

Enet4:
> Is it possible to add `pub fn into_value(self) -> Value<I> {` ?

Sure, I just pushed PR #18, which can be merged before this one.

ibaryshnikov:
@Enet4 would you mind rebasing it on master to grab the `into_value` update?

ibaryshnikov:
I'm running into another issue trying to update my code. I want to iterate through entries using `into_iter`, which is defined for `InMemDicomObject` and does `self.entries.into_iter()`. It's required because field `self.entries` is private.

```rust
    let root_object = from_reader(&input[128..])
        .map_err(|e| Error::new(&format!("Error parsing input buffer: {:?}", e)))?;

    for data_element in root_object.into_iter() {
        
    }
```
The error is `cannot move out of dereference of RootDicomObject<InMemDicomObject<StandardDataDictionary>>`

Since field `obj` in `RootDicomObject` is private, the only way to obtain it at the moment is to use deref
```rust
impl<T> ::std::ops::Deref for RootDicomObject<T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        &self.obj
    }
}
```
I guess, having something similar with `into_value` for `RootDicomObject` may help here

ibaryshnikov:
I tested this PR with
```toml
[dependencies]
dicom-core = "0.1.0"
dicom-object = "0.1.0"
```
and having a local checkout of this PR with minor changes
```toml
[patch.crates-io.dicom-core]
path = "../../../dicom-rs/core"
[patch.crates-io.dicom-object]
path = "../../../dicom-rs/object"
```
Both `dicom-object = "0.1.0"` and `dicom-object = { version = "0.1.0", features = ["inventory-registry"] }` work well in WebAssembly. I'd also like to test that while using feature `inventory-registry` it's not possible to extend TS registry in WebAssembly and possible to do it natively (which is expected behavior). I'll share the results after more tests tomorrow.

Changes I made in a local copy:
```rust
impl<T> RootDicomObject<T> {
    // not sure about a method name
    pub fn into_object(self) -> T {
        self.obj
    }
}
```

Enet4:
Just rebased. And yet, `RootDicomObject` is still a bit thin here. I just pushed another PR, #19, which should have what you are looking for. I called the method `into_inner` instead, as often employed for wrapper values. `into_object` might also be ambiguous, since `RootDicomObject` already behaves like an object.

Thank you for the testing and feedback so far.

ibaryshnikov:
Since #19 is merged, it'll be good to rebase this PR again.
Overal, I was able to test it with a local checkout
I used
```toml
[patch.crates-io.dicom-core]
path = "../../../dicom-rs/core"
[patch.crates-io.dicom-object]
path = "../../../dicom-rs/object"
```
All three options work fine for built-in transfer syntaxes
- `dicom-object = { version = "0.1.0" }`
- `dicom-object = { version = "0.1.0", default-features = false }`
- `dicom-object = { version = "0.1.0", features = ["inventory-registry"] }`

I ran into one compiler error with this code
```toml
dicom-object = { version = "0.1.0", features = ["inventory-registry"] }
```
```rust
submit_transfer_syntax! {
    TransferSyntax::new(
        "1.2.840.10008.1.2.4.70",
        "JPEG Lossless, Non-Hierarchical, First-Order Prediction",
        Endianness::Little,
        true,
        Codec::EncapsulatedPixelData,
    )
}
```
The error is
```
error[E0282]: type annotations needed
 --> <::inventory::submit macros>:1:73
  |
1 | ($ ($ value : tt) *) => { $ crate :: r#impl :: submit ! { $ ($ value) * } }
  |                                                                         ^ cannot infer type for `A`

error: aborting due to previous error
```

ibaryshnikov:
Tested again with
```toml
[patch.crates-io.dicom-core]
git = "https://github.com/Enet4/dicom-rs"
rev = "2647055"
[patch.crates-io.dicom-object]
git = "https://github.com/Enet4/dicom-rs"
rev = "2647055"
```
Everything works fine for WebAssembly target:
- `dicom-object = { version = "0.1.0" }`
- `dicom-object = { version = "0.1.0", default-features = false }`
- `dicom-object = { version = "0.1.0", features = ["inventory-registry"] }`

Surprisingly, I even saw some speedups in decoding)

Enet4:
The main `TransferSyntax` type is generic over an adapter type `A`, which means that your code still needs to disambiguate `A`. In most cases, the default `A = DynDataRWAdapter` works, but this is one situation where the compiler will choose not to infer to the default parameter type (see [this SO question](https://stackoverflow.com/q/57824584/1233251)).

In practice, it's easier to create a static variable of the TS in a `lazy_static` context. Please see if something like the code below works.

```rust
lazy_static! {
    static ref MY_TS: TransferSyntax = TransferSyntax::new(
        "1.2.840.10008.1.2.4.70",
        "JPEG Lossless, Non-Hierarchical, First-Order Prediction",
        Endianness::Little,
        true,
        Codec::EncapsulatedPixelData,
    );
}
submit_transfer_syntax!(MY_TS);
```

(One shouldn't need to register "JPEG Lossless, Non-Hierarchical, First-Order Prediction" like this, it's already registered built-in. You would later on do this to implement pixel data encoding and decoding, which is yet to be designed at the moment.)

ibaryshnikov:
Thanks, I'll try it

> One shouldn't need to register "JPEG Lossless, Non-Hierarchical, First-Order Prediction" like this, it's already registered built-in.

I know, I just commented it away in `entries.rs` and tried to submit it externally to test the `submit_transfer_syntax` macro )

Enet4:
I have just pushed another commit which includes a way to open DICOM objects using another TS registry, which means that in theory, even in an environment without `inventory`, you could build your own collection of supported transfer syntaxes and pass that to the constructors.

ibaryshnikov:
Sorry for a delay. As default transfer syntaxes work in webassembly, I'm just testing that extention of TS registry works as expected for native target.

ibaryshnikov:
@Enet4 I tested all the cases I wanted: ts extension works fine for native target and doesn't work for webassembly (as expected). I haven't tested external TS registries though, but this change looks pretty straightforward. I think, it's good to go now. Thank you for updates!

PS, I figured out how to submit transfer syntax properly:
```rust
submit_transfer_syntax! {
    TransferSyntax::<NeverAdapter>::new(
        "1.2.840.10008.1.2.4.70",
        "JPEG Lossless, Non-Hierarchical, First-Order Prediction",
        Endianness::Little,
        true,
        Codec::EncapsulatedPixelData,
    )
}
```
The key part here is `NeverAdapter`, after I added it the error ``cannot infer type for `A` `` disappeared, therefore no need to use `lazy_static`

PR #18, opened 2019-10-07
Add DataElement.into_value

Enet4:
@ibaryshnikov I suppose this is what you had in mind? I am not opposed to the existence of this method, and at this stage, identifying useful methods and other pieces that feel missing from the API is much more important than rock-solid stability.


ibaryshnikov:
Yes, exactly this. Thanks!

PR #19, opened 2019-10-07
RootDicomObject into_inner + into_iter

Enet4:
Provides `RootDicomObject.into_inner` and makes it implement `IntoIterator`, currently by traversing the elements of the inner object (the meta table is not included).

### DIFF
--- a/object/src/lib.rs
+++ b/object/src/lib.rs
@@ -138,5 +138,17 @@ where
     }
 }
 
+impl<T> IntoIterator for RootDicomObject<T>
+where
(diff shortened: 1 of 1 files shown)


ibaryshnikov:
`impl<T> IntoIterator for RootDicomObject<T>` is actually better than what I suggested, so it's possible to go without `into_iter` now

Enet4:
Yes, that was my thought. For all intents and purposes, a root object should be usable as any other DICOM object.

PR #29, opened 2019-12-19
Add some robustness in meta file group reading

### LINKED ISSUE #28: Handling Tag(0x0002, 0x0002) in FileMetaTable
There's a number of different tags possible in file meta information, as described in http://dicom.nema.org/medical/dicom/current/output/chtml/part10/chapter_7.html#table_7.1-1
Is it possible to support additional tags in FileMetaTable, which at the moment has the following check
```rust
if elem.tag() != (0x0002, 0x0001) {
    return Err(Error::UnexpectedTag(elem.tag()));
}
```

Enet4:
Resolves #28.

- allow file meta group attributes in arbitrary order
- allow missing File Meta Information Version (the default is `(00H, 01H)`)

CC @ibaryshnikov

### DIFF
--- a/object/src/meta.rs
+++ b/object/src/meta.rs
@@ -343,12 +343,10 @@ impl FileMetaTableBuilder {
         let information_group_length = self
             .information_group_length
             .ok_or_else(|| Error::InvalidFormat)?;
-        let information_version = self
-            .information_version
-            .or_else(|| {
-                // Missing information version, will assume (00H, 01H). See #28
-                Some([0, 1])
-            }).unwrap();
+        let information_version = self.information_version.unwrap_or_else(|| {
+            // Missing information version, will assume (00H, 01H). See #28
+            [0, 1]
+        });
         let media_storage_sop_class_uid = self
             .media_storage_sop_class_uid
             .ok_or_else(|| Error::InvalidFormat)?;
--- a/object/src/meta.rs
+++ b/object/src/meta.rs
@@ -46,34 +46,6 @@ pub struct FileMetaTable {
     pub private_information: Option<Vec<u8>>,
 }
 
-/// Utility function for reading the whole DICOM element as a string, with the given tag.
-fn read_str_as_tag<'s, S: 's, D, T>(
-    source: &'s mut S,
-    decoder: &D,
-    text: &T,
-    group_length_remaining: &mut u32,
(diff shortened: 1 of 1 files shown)


object/src/meta.rs
```diff
@@ -178,57 +109,99 @@ impl FileMetaTable {
...
-                    let mut v = vec![0; elem_len];
+                    let mut v = vec![0; elem_len as usize];
                     file.read_exact(&mut v)?;
-                    group_length_remaining -= 12 + elem_len as u32;
+                    group_length_remaining -= 12 + elem_len;
                     builder.private_information(v)
                 }
                 Tag(0x0002, _) => {
                     // unknown tag, do nothing
+                    // could be an unsupported or non-standard attribute
```
ibaryshnikov: Small nit, they're reserved
`Values of all Tags (0002,xxxx) are reserved for use by this Standard and later versions of DICOM.`
which mentioned below the table on http://dicom.nema.org/medical/dicom/current/output/chtml/part10/chapter_7.html#table_7.1-1

object/src/meta.rs
```diff
@@ -372,7 +345,10 @@ impl FileMetaTableBuilder {
             .ok_or_else(|| Error::InvalidFormat)?;
         let information_version = self
             .information_version
-            .ok_or_else(|| Error::InvalidFormat)?;
+            .or_else(|| {
```
ibaryshnikov: `unwrap_or_else` may be handy here

ibaryshnikov:
Awesome work! I'm going to try this branch, will share the results soon

ibaryshnikov:
I've just tested it, #28 is fixed in this pr

object/src/meta.rs
Enet4: Indeed, thanks! For some reason my IDE did not suggest me that method.

object/src/meta.rs
Enet4: Well yes, but with this comment I am keeping two cases in mind:

- the standard is extended to admit more file meta attributes, but the library does not know them yet
- a pseudo-DICOM producer invents its own attribute within the same group

In either case, nothing happens, but in the future we might be interested in logging it differently from the case below this one.

Enet4:
Thank you for testing, @ibaryshnikov!

Issue #23, opened 2019-11-17
Support PixelData of undefined length


ibaryshnikov:
It was briefly discussed in comments for #22
The file to reproduce (at the end of the comment) https://github.com/pydicom/pydicom/issues/738#issue-359963120
Stackoverflow answer https://stackoverflow.com/a/20886301
Page in standard http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_A.4.html#table_A.4-1

PR #30, opened 2019-12-19
Reading PixelData of undefined length

ibaryshnikov:
It's my first naive implementation, consider it an early WIP. I do really need your advice in structuring the code for this pr, @Enet4. But it seems to work and the file linked in #23 can be opened successfully.

### DIFF
--- a/parser/src/stateful/decode.rs
+++ b/parser/src/stateful/decode.rs
@@ -198,6 +198,69 @@ where
     fn read_value_ob(&mut self, header: &DataElementHeader) -> Result<PrimitiveValue> {
         // TODO add support for OB value data length resolution
         // (might need to delegate pixel data reading to a separate trait)
+
+        // if there is PixelData of undefined length
+        if header.tag == Tag(0x7FE0, 0x0010) && header.len.is_undefined() {
+            let mut buf = [0; 8];
+            self.from.read_exact(&mut buf)?;
+            self.bytes_read += 8;
+
+            println!("PixelData of undefined length");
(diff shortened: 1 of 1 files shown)


parser/src/parser.rs
```diff
@@ -218,6 +218,69 @@ where
...
+            println!("Tag({:#06X}, {:#06X})", u16::from_le_bytes([buf[0], buf[1]]), u16::from_le_bytes([buf[2], buf[3]]));
+            let len = u32::from_le_bytes([buf[4], buf[5], buf[6], buf[7]]) as usize;
+            println!("Offset table len {}", len);
+
+            let mut offset_table = vec![0; len];
+            from.read_exact(&mut offset_table)?;
+            self.bytes_read += len as u64;
+
+            let mut offsets = vec![];
+            for i in 0..len / 4 {
```
Enet4: This logic looks suspicious, considering that `i` still iterates one by one from 0 to `len/4` without being adjusted accordingly. The table offset reads are overlapping.

You could probably create the list of offsets functionally, by chunking `offset_table` into pieces of 4. Maybe use [`Itertools#tuples`](https://docs.rs/itertools/0.8.2/itertools/trait.Itertools.html#method.tuples)?

Enet4:
I'm afraid that this PR has become a bit outdated with #35. I rebased the branch, but tests and changes are still pending.

@ibaryshnikov Not sure if you are able to follow up on this, but if not this makes an opportunity to take a step back and implement encapsulated pixel data at the data-set level instead.

Enet4:
Thank you for the attempt, but this feature has been succeeded by #42.

Issue #1, opened 2019-07-08
Lazy loading of Multi-frame files


ibaryshnikov:
Hi @Enet4! Thank you for creating such an awesome library!

I have a specific use case: loading big `.dcm` files (over 500Mb) in browser. At the moment most of the browsers have a limit for webassembly memory of about 2Gb, and when I run my program, it usually crashes with out of memory error.

Is there a way we can collaborate to support lazy loading?
In particular, I need to iterate over the PixelData items in Multi-frame file

Enet4:
Hello! Thank you very much for your interest in this project!

While it's needless to say that this library isn't quite ready for use yet, I do intend to thrust development further into a first release in the near future. This one should have enough features to read most DICOM files in-memory. Lazy loading has been planned from the beginning in some way, but it might not be available in the first release. This should totally be supported eventually! After 0.1.0, I wish to conceive a roadmap with more concrete goals, that will allow me to lay down independent tasks and attract contributors in the process. Then, I would be more than glad to hear more from other experts and enthusiasts to drive this project forward.

pauldotknopf:
I want to begin implementing this before completing #15.

There is a ```LazyDataSetReader``` currently implemented, but not being used. @Enet4, can you elaborate more on your intentions here? It looks like lazy loading is intended to be an "all or nothing" approach?

My thoughts:
* Lazy loading should be built in to the standard ```RootDicomObject```, regardless of the ```DicomObject``` being used (```InMemDicomObject```, or the presumed to-be-implemented ```LazyDicomObject```). So, there should be a single ```DicomObject``` implementation.
* The pivot to lazy load should be based on the input stream. Maybe create a custom trait the implements both ```Read``` and ```Seek```, but with additional properties indicating if seek is supported, and the ability to build a factory token that can be used to rebuild the trait. We would provide helper functions to open a file stream, or a stream from an in-memory buffer, etc.
* Another pivot point would be a parameter to indicate the maximum value to load into memory, all larger elements being lazy loaded. A default value should be sane. This value would determine if an element should be lazy loaded, including OB, OW, and UT, etc. The default value would mostly catch Pixel Data values, in practice, but would also catch really large strings.
* We should add a new ```PrimitiveValue::Lazy``` type that contains a ```pub enum LazyToken``` type. This token can be used to build a ```Read``` trait. This way, the values can stay unloaded, but can be partially buffered and written to another ```Write``` stream when writing to another file or network.
[...]

PR #32, opened 2020-01-05
Data set writing

Enet4:
This PR adds substantial changes to the encoding and parser crates, which then allowed me to implement DICOM data set writing.

- Changed the Encode and Decode APIs to follow triple handshaking: `Encode` and `Decode` are monomorphized over writers and readers, whereas `EncodeTo<W>` and `DecodeFrom<S>` are restrained to a specific type. This allows implementations to be generic over the source/target while enabling object safe consumption.
- Changed some methods in encoder traits to expose the concrete number of bytes that were written
- Remove erased decoder: it was not useful
- Reorganized the data set reading and writing into its own module in the parser crate: `dataset`
- Add a `Printer` type; for the lack of a better name, this one is the dual to the `DicomParser` type, and is a stateful encoder which admits proper text encoding.
- Add a data set writer type in the parser crate
- Multiple bugs in the encoders were found and fixed.
- Move the primitive value parsing module to `encoding`
- Also re-export `chrono` via `dicom_core`


pauldotknopf:
Lazy comment, but I'm hoping you got the use case covered where we are reading/writing large fields (PixelData), buffered directly between network->file streams and vice versa. We don't want to load 4gb chunks into RAM.

Enet4:
@pauldotknopf That will be something I want to reiterate after the complete writing pipeline is set up. This PR is large enough as it is, and I picture at least three things to consider here in order to achieve such a stream-based manipulation of DICOM values:

1. At the core and encoding crates, there is still no standard way to describe a "lazy" DICOM value. The API is yet to be concretely designed. I can imagine another trait, but it would need to coordinate with the remaining traits for encoding and decoding.
2. The functionality would need to be coordinated with #23, which inevitably influences the parsing process. In the case where pixel data is encapsulated with small enough chunks, we may choose to treat each chunk as an independently retrieved token, thus not having to keep it all in memory. If you have stumbled upon a DICOM file with a very large non-encapsulated pixel data, I would be grateful to know. To my knowledge, large files are usually encapsulated to benefit from image compression algorithms and an unbounded size.
3. The inverse process (data set reader) would also need to be adjusted to work with lazy DICOM values, as discussed in #1.

PR #35, opened 2020-04-22
DICOM object writing

### LINKED ISSUE #15: Writing DICOM files?
I've been working on SCU/SCP support. I've had to pause because of the missing write support.

With that said, what are you thinking with regards to writing? Do you want to stub out some traits and then I will handle the implementation/tests?

Enet4:
This is another bulky set of contributions that ultimately enable users of DICOM-rs to write objects to a file. Resolves #15 as the MVP to writing, future issues related with the writing process can be made into more specific issues.

This was accompanied by several major changes, but I will summarize them here:

- substantially changed the parser and printer abstractions to be become _stateful_ decoder and encoder abstractions, thus being easier to understand their purpose and behaviour: they are like plain encoders/decoders, but target a specific byte source/sink and are aware of changes to text encoding.
- extended the dataset module with the `IntoTokens` trait and multiple useful implementations of it (including for the in-memory DICOM object type and for the file meta table).
- new experimental `dicom_value!` macro for constructing primitive values of variable cardinality.
- the method `RootDicomObject.write_to_file(path)` lets you write a complete DICOM object to disk.

### DIFF
--- a/core/src/value.rs
+++ b/core/src/value.rs
@@ -1,7 +1,7 @@
 //! This module includes a high level abstraction over a DICOM data element's value.
 
 use crate::error::CastValueError;
-use crate::header::{Length, Tag};
+use crate::header::{HasLength, Length, Tag};
 use chrono::{Datelike, FixedOffset, Timelike};
 use itertools::Itertools;
 use smallvec::SmallVec;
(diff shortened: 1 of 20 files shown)


core/src/value.rs
```diff
@@ -210,7 +212,75 @@ pub enum PrimitiveValue {
...
+    }
+}
+
 impl PrimitiveValue {
+    /// Create a single unsigned 16-bit value.
+    pub fn new_u16(value: u16) -> Self {
+        PrimitiveValue::U16(C::from_elem(value, 1))
+    }
+
+    /// Create a single unsinged 32 value.
```
Enet4, line 274: The most unsung kind of typo.

```suggestion
    /// Create a single unsigned 32-bit value.
```

PR #44, opened 2020-05-30
Add support for OB values with unkown length

robyoung:
This PR needs a bit of rearranging but the core of it is here and works. There are two distinct parts to it.

## Add integration test for unknown OB length

While trying to parse some example DICOM files the most common issue I came across was failing because the file had an OB value of unknown length. In order to get some good high level tests around this I asked over at [pydicom](https://github.com/pydicom/pydicom/issues/1112) if I could use some of their example files. Off the back of that I have created a separate crate for providing example DICOM files called (rather unimaginatively) [dicom-test-files](https://github.com/robyoung/dicom-test-files) and created a failing integration test for a file that uses unknown OB length.

## Add support for decoding OB value of unknown length

Change `Read` to `ReadSeek` in a few places (or just add `Write`) so that we can read ahead in chunks and then seek back when we overshoot rather than reading a byte at a time. I experimented with reading with very small chunks and even though we're going over a `BufferedReader` performance took a massive hit.

Getting a `Tag` as bytes in a given endianness feels like it's in the wrong place but I'm not sure where the right place is. It feels like a trait that provides `as_bytes` would be ideal but I'm not sure
where that should live. Neither core nor parser seem quite right.

Panicking when the trailing 4 zero bytes are not found is not ideal but I'm not sure of the best error to return there.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -21,6 +21,7 @@ pub use dictionary::DataDictionary;
 pub use error::{Error, Result};
 pub use header::{DataElement, DataElementHeader, Length, Tag, VR};
 pub use value::{PrimitiveValue, Value as DicomValue};
+pub use util::ReadSeek;
 
 // re-export the chrono crate, as used in the public API
 pub use chrono;
--- a/core/src/util.rs
+++ b/core/src/util.rs
@@ -22,6 +22,7 @@ where
     }
 }
 
+/// A trait that combines io::Read and io::Seek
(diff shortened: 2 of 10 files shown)


robyoung:
I have just rebased on master and it seems the element type has changed but also the data is now completely different to what I'm expecting. I'm not sure what's going on here.

robyoung:
I realised this change had been superseded when I rebased on master. I can certainly pull the test files stuff out into a separate PR.

PR #45, opened 2020-05-30
[dcmdump] term size and primitive columns

robyoung:
- Add term_size crate to expand width for wider screens.
- Add divider between meta and body.
- Reorder Primitive fields so that tag and tag alias are together on the
  left and other fields expand out from there. The hope is that this
  makes it easier to scan the output.

### DIFF
--- a/dcmdump/src/main.rs
+++ b/dcmdump/src/main.rs
@@ -15,6 +15,8 @@ use dicom::core::VR;
 use dicom::object::mem::{InMemDicomObject, InMemElement};
 use dicom::object::{open_file, DefaultDicomObject, FileMetaTable, StandardDataDictionary};
 
+use term_size;
+
 use std::borrow::Cow;
 use std::io::{stdout, ErrorKind, Result as IoResult, Write};
 
@@ -42,9 +44,15 @@ fn dump_file(obj: DefaultDicomObject) -> IoResult<()> {
 
     let meta = obj.meta();
 
-    let width = 48;
+    let width = if let Some((width, _)) = term_size::dimensions() {
+        width as u32
+    } else {
+        48
+    };
 
     meta_dump(&mut to, &meta, width)?;
+
+    println!("{:-<40}", "");
 
     dump(&mut to, &obj, width, 0)?;
 
@@ -188,13 +196,13 @@ where
             let byte_len = value.calculate_byte_len();
             writeln!(
                 to,
-                "{} {} {:48} # {}, {} {}",
+                "{} {:35} {} ({},{:>3} bytes): {}",
                 elem.tag(),
+                tag_alias,
                 vr,
-                value_summary(&value, vr, width),
-                byte_len,
                 vm,
-                tag_alias
+                byte_len,
+                value_summary(&value, vr, width),
             )?;
         }
     }
@@ -267,18 +275,17 @@ where
     I: IntoIterator,
     I::Item: std::fmt::Display,
 {
-    let pieces = values.into_iter().take(64);
-    let max = max_characters as usize;
-    let mut o = String::with_capacity(max);
-    for piece in pieces {
-        o.push_str(&piece.to_string());
-        o.push('\\');
-        if o.len() > max {
-            break;
-        }
-    }
-    o.pop();
-    cut_str(&o, max_characters).into_owned()
+    let pieces = values
+        .into_iter()
+        .take(64)
+        .map(|piece| format!("\"{}\"", piece))
+        .collect::<Vec<String>>();
+    let pieces = if pieces.len() > 1 {
+        format!("[{}]", pieces.join(", "))
+    } else {
+        pieces.join(", ")
+    };
+    cut_str(&pieces, max_characters).into_owned()
 }
 
 fn cut_str(s: &str, max_characters: u32) -> Cow<str> {
--- a/dcmdump/Cargo.toml
+++ b/dcmdump/Cargo.toml
@@ -14,3 +14,4 @@ default = ['dicom/inventory-registry']
 
 [dependencies]
 dicom = { path = "../parent/", version = "0.2.0", default-features = false }
+term_size = "0.3.2"


dcmdump/src/main.rs
```diff
@@ -267,18 +275,17 @@ where
...
-            break;
-        }
-    }
-    o.pop();
-    cut_str(&o, max_characters).into_owned()
+    let pieces = values
+        .into_iter()
+        .take(64)
+        .map(|piece| format!("\"{}\"", piece))
+        .collect::<Vec<String>>();
```
Enet4, line 282: I also believe that you do not need to `collect` the string parts into a vector if you use [`Itertools::join`](https://docs.rs/itertools/0.9.0/itertools/trait.Itertools.html#method.join) to join the strings.

robyoung: Although, we need to know how many items there are in order to know whether to wrap in a square brackets.

robyoung:
The purpose is definitely not to make it easy to scan programmatically. I totally agree that a CSV or JSON output would be much better for that. The crux of this change for me is pulling the `tag_alias` over to the left and padding to align the fields after it. I found this much easier scan down the list of tags to find what I was interested in. This is roughly how pydicom prints things out.

With regards to point 1. I'd be happy to remove the quotes all together if that would be easier?

With regards to point 2. I'm not sure it would be useful to prevent all line wrapping. Given how far the indenting and padding can go you run the risk of reducing the amount of visible data to very little even for short fields. I feel like so long as the tag details and field information are aligned properly a bit of line wrapping will not hugely impede being able to scan your eye down the list.

Enet4:
> The purpose is definitely not to make it easy to scan programmatically. I totally agree that a CSV or JSON output would be much better for that. The crux of this change for me is pulling the tag_alias over to the left and padding to align the fields after it. I found this much easier scan down the list of tags to find what I was interested in. This is roughly how pydicom prints things out.

My initial approach to the output of this tool was to approximate the look of DCMTK's dcmdump, but I'm always interested in other ways to better present this information to the user.

> With regards to point 1. I'd be happy to remove the quotes all together if that would be easier?

It might actually be easier to make the necessary changes that distinguish binary from textual values from the current state of this PR. I can work on this after merging.

> With regards to point 2. I'm not sure it would be useful to prevent all line wrapping. Given how far the indenting and padding can go you run the risk of reducing the amount of visible data to very little even for short fields. I feel like so long as the tag details and field information are aligned properly a bit of line wrapping will not hugely impede being able to scan your eye down the list.

I suppose that this can be further investigated down the line. perhaps even exploiting vertical space to show more data about an element might become an option.

But considering this new format, we can least make a few adjustments for consistency:

- Tweak the spacing a bit so that the item line outputs to become more consistent. At least for these cases, they usually end up wrapping to the next line with redundant information, so we can just remove the trailing output entirely.
- In sequence elements (and pixel elements), also present the header with a similar format as the rest, including the number of items.

Here's the full diff which I believe resolves these two issues, feel free to experiment with these changes.

```diff
diff --git a/dcmdump/src/main.rs b/dcmdump/src/main.rs
index 0f0db7c..a97d583 100644
--- a/dcmdump/src/main.rs
+++ b/dcmdump/src/main.rs
@@ -139,20 +139,19 @@ where
         DicomValue::Sequence { items, .. } => {
             writeln!(
                 to,
-                "{} {} {:48} # {},    {}",
+                "{} {:35} {} ({} Item{})",
                 elem.tag(),
+                tag_alias,
                 elem.vr(),
-                "",
                 vm,
-                tag_alias
+                if vm == 1 { "" } else { "s" },
             )?;
             for item in items {
                 dump_item(&mut *to, item, width, depth + 2)?;
             }
             writeln!(
                 to,
-                "(FFFE, E0DD) na (SequenceDelimitationItem) {:20} # 0, SequenceDelimitationItem",
-                "",
+                "(FFFE, E0DD) SequenceDelimitationItem",
             )?;
         }
         DicomValue::PixelSequence {
@@ -164,18 +163,19 @@ where
             let num_items = 1 + fragments.len();
             writeln!(
                 to,
-                "{} {} (PixelSequence=#{}) {:29} # u/l, 1 PixelData",
+                "{} PixelData {:25} {} (PixelSequence, {} Item{})",
                 elem.tag(),
+                "",
                 vr,
                 num_items,
-                ""
+                if num_items == 1 { "" } else { "s" },
             )?;
 
             // write offset table
             let byte_len = offset_table.len();
             writeln!(
                 to,
-                "  (FFFE,E000) pi {:48} # {}, 1 Item",
+                "  (FFFE,E000) pi ({:>3} bytes, 1 Item): {:48}",
                 item_value_summary(&offset_table, width),
                 byte_len,
             )?;
@@ -185,9 +185,9 @@ where
                 let byte_len = fragment.len();
                 writeln!(
                     to,
-                    "  (FFFE,E000) pi {:48} # {}, 1 Item",
-                    item_value_summary(&fragment, width),
+                    "  (FFFE,E000) pi ({:>3} bytes, 1 Item): {:48}",
                     byte_len,
+                    item_value_summary(&fragment, width),
                 )?;
             }
         }
@@ -216,15 +216,12 @@ where
     D: DataDictionary,
 {
     let indent: String = std::iter::repeat(' ').take((depth * 2) as usize).collect();
-    let trail: String = std::iter::repeat(' ')
-        .take(usize::max(21, width as usize - 4 - indent.len()))
-        .collect();
-    writeln!(to, "{}(FFFE,E000) na Item {} # 0, 0 Item", indent, trail)?;
+    writeln!(to, "{}(FFFE,E000) na Item", indent)?;
     dump(to, item, width, depth + 1)?;
     writeln!(
         to,
-        "{}(FFFE,E00D) na (ItemDelimitationItem) {:23} # 0, 0 ItemDelimitationItem",
-        indent, ""
+        "{}(FFFE,E00D) ItemDelimitationItem",
+        indent,
     )?;
     Ok(())
 }
```

Enet4:
For now, I will take this contribution and fine-tune a bit afterwards. There are yet greater improvements and features to consider in the future, and the project will continue being open to proposing and discussing them.

Thank you!

PR #48, opened 2020-06-05
Add integration test for unknown OB length

robyoung:
This test makes use of a DICOM file that was originally from [pydicom](https://github.com/pydicom/pydicom) that contains an OB field with unknown length.

This makes use of the [dicom-test-files](https://github.com/robyoung/dicom-test-files) crate that provides access to
DICOM files that can be used for testing across different DICOM readers.

### DIFF
--- a/object/tests/integration_test.rs
+++ b/object/tests/integration_test.rs
@@ -11,7 +11,7 @@ fn test_ob_value_with_unknown_length() {
 
     match element.value() {
         Value::PixelSequence { fragments, .. } => {
-            // check the start and end of the bytes the check it looks right
+            // check if the leading and trailing bytes look right
             assert_eq!(fragments.len(), 1);
             let fragment = &fragments[0];
             assert_eq!(fragment[0..2], [255, 79]);
--- a/object/tests/integration_test.rs
+++ b/object/tests/integration_test.rs
@@ -1,0 +1,24 @@
+use dicom_core::value::Value;
+use dicom_object::open_file;
+use dicom_test_files;
+
+#[test]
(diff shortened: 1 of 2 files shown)


object/tests/integration_test.rs
```diff
@@ -0,0 +1,24 @@
...
+#[test]
+fn test_ob_value_with_unknown_length() {
+    let path =
+        dicom_test_files::path("pydicom/JPEG2000.dcm").expect("test DICOM file should exist");
+    let object = open_file(&path).unwrap();
+    let element = object.element_by_name("PixelData").unwrap();
+
+    match element.value() {
+        Value::PixelSequence { fragments, .. } => {
+            // check the start and end of the bytes the check it looks right
```
Enet4, line 14: An odd grammatical construction here, maybe this would be clearer?

```suggestion
            // check if the leading and trailing bytes look right
```

object/tests/integration_test.rs
```diff
@@ -0,0 +1,24 @@
...
+            assert_eq!(fragments.len(), 1);
+            let fragment = &fragments[0];
+            assert_eq!(fragment[0..2], [255, 79]);
+            assert_eq!(fragment[fragment.len() - 2..fragment.len()], [255, 217]);
+        },
+        _ => {
+            panic!("expected a byte value");
```
Enet4, line 21: We can make this panic slightly more useful in the event of regressions:

```rust
val => {
    panic!("expected a byte value sequence, but got {:?}", val);
}
```

Enet4:
I will take the liberty of applying minor changes later. Thank you again, @robyoung.

PR #52, opened 2020-06-13
fix parse Specific Character Set error

### LINKED ISSUE #51: Parse Specific Charset error.
Hi!
I've added support for Chinese encoding, but Chinese characters still can't be parsed correctly. At last I found the reason why it went wrong.
The Character set should be changed [here](https://github.com/Enet4/dicom-rs/blob/master/parser/src/stateful/decode.rs#L504), but it never seems to run here. Because the Specific Character code is `CS`, not `UI`. Please take a look [here](https://dicom.innolitics.com/ciods/mr-image/sop-common/00080005).

kira-96:
Fixed an error in parsing character encoding and added support for Chinese characters.

Resolves #51.

### DIFF
--- a/encoding/src/text.rs
+++ b/encoding/src/text.rs
@@ -15,7 +15,7 @@
 //! At the moment, this library supports only IR-6 and IR-192.
 
 use crate::error::{Result, TextEncodingError};
-use encoding::all::{ISO_8859_1, UTF_8};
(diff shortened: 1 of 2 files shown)


Enet4 (approved):
These are good changes, nice catch! I am preparing a PR to provide test coverage over this part of the decoder, which I can also merge after this one.

Thank you very much!

Issue #49, opened 2020-06-07
Logging foundation


Enet4:
As defined in the roadmap, some situations while working with DICOM data may not be critical to impose a soft error or panic, but should still be called out in some way. Logging comes to play as a non-invasive mechanism to report abnormal situations which do not impede the process, but may bring other issues along the way. As such, it is still relevant to provide this information to developers and to end users of CLI tools.

The two phases for fulfilling this issue would be:

- [x] Decide on a logging API for all library crates (as a façade to an actual logger implementation which can be decided by dependents). I believe this would be narrowed down to choosing between `log`, `slog`, or `tracing`.
   - `tracing` it is.
- [x] Choose a logging implementation for each existing CLI tool (`dicom-dump`, `dicom-scproxy`, and so on). One that prints to stdout/stderr would suffice.

PR #53, opened 2020-06-13
Text encoding revamp: trait changes and more text encodings

Enet4:
This is another step intended to cover some of the current issues in text encoding.

**Major changes:**

- `TextCodec` no longer implies `Debug`
   - This is not a common practice anyway, such a constraint makes more sense where it is actually needed.
- New method `TextCodec::name`, for retrieving the character set's defined term.
- `SpecificCharacterSet` is marked as non-exhaustive, so that the presence of future encodings does not break future code upon this change.

**New:**

- New character encodings: ISO-IR 100, ISO-IR 101, ISO-IR 109, ISO-IR 110, ISO-IR 144.  
- Report to stderr when an unsupported text encoding is decoded (should become a warning log line with #49).

**Misc:**

- Added a test at `parser` to check that the specific character set is updated after decoding the respective data element (verifies #52).
- Used a macro to implement existing text encodings via the `encoding` crate.
- Add baseline tests for some of the current text encodings.

### DIFF
--- a/encoding/src/text.rs
+++ b/encoding/src/text.rs
@@ -21,7 +21,17 @@ use std::fmt::Debug;
 
 /// A holder of encoding and decoding mechanisms for text in DICOM content,
 /// which according to the standard, depends on the specific character set.
-pub trait TextCodec: Debug {
+pub trait TextCodec {
+
+    /// Obtain a unique name of the text encoding,
+    /// which may be used as the value of a
+    /// Specific Character Set (0008, 0005) element to refer to this codec.
+    ///
+    /// Should contain no leading or trailing spaces.
+    /// This method may be useful for testing purposes, considering that
+    /// `TextCodec` is often used as a trait object.
+    fn name(&self) -> &'static str;
+
     /// Decode the given byte buffer as a single string. The resulting string
     /// _may_ contain backslash characters ('\') to delimit individual values,
     /// and should be split later on if required.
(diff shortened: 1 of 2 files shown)


ikneg:
@Enet4 Is it OK to ask questions here related to this PR? If it is against the rules, I will delete the post.

> Report to stderr when an unsupported text encoding is decoded (should become a warning log line with https://github.com/Enet4/dicom-rs/issues/49).

According to avobe description, unsupported text encoding will cause an error when decoded, but is the case I tried below the case where an error should occur?

```rust
let specific_character_set = self
    .object
    .element(tags::SPECIFIC_CHARACTER_SET)
    .unwrap()
    .to_multi_str();
let patient_name = self
    .object
    .element(tags::PATIENT_NAME)
    .unwrap()
    .to_multi_str();
info!("SPECIFIC_CHARACTER_SET: {:?}", specific_character_set);
info!("PATIENT_NAME:           {:?}", patient_name);
```

```text
2023-11-24T08:44:03.903759Z  INFO api::presentation::validates::analysis::dicom: SPECIFIC_CHARACTER_SET: Ok(["ISO 2022 IR 13", "ISO 2022 IR 87"])
2023-11-24T08:44:03.903783Z  INFO api::presentation::validates::analysis::dicom: PATIENT_NAME:           Ok(["Å¶ÔÏ^º³¼Þ=\u{1b}$BCf;3\u{1b}(J^\u{1b}$B9'<#\u{1b}(J"])
```

PR #64, opened 2020-08-27
Treat undefined length UN as sequence

Enet4:
Fixes not being able to parse a data set in Implicit VR Little Endian when an unknown sequence attribute is found.

### DIFF
--- a/parser/src/dataset/read.rs
+++ b/parser/src/dataset/read.rs
@@ -327,33 +327,36 @@ where
                     Some(Ok(DataToken::SequenceStart { tag, len }))
                 }
                 Ok(DataElementHeader {
-                    tag,
-                    vr: VR::UN,
-                    len,
-                }) if len.is_undefined() => {
-                    // treat undefined length UN elements
-                    // as sequences
-                    self.in_sequence = true;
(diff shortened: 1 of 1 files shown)


parser/src/dataset/read.rs
```diff
@@ -326,6 +326,18 @@ where
 
                     Some(Ok(DataToken::SequenceStart { tag, len }))
                 }
+                Ok(DataElementHeader {
+                    tag,
+                    vr: VR::UN,
```
Enet4, line 331: Just checked the relevant section of the standard recently ([part 5, 7.1.2](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/chapter_7.html#sect_7.1.2)), it is actually possible for elements of other VRs to have an undefined length. The full list is: OB, OD, OF, OL, OV, OW, SQ, and UN. The ones which are still missing and would break the parser are OW, OL, OV, OF, and OD.

I should check the implications of supporting these value representations.

Enet4: The decision was to treat such elements as data set sequences regardless of the VR. Should exceptions to this rule come around, we can re-evaluate it, but nevertheless this would be an improvement to the previous algorithm.

PR #79, opened 2020-10-03
Add take_element methods for InMemoryDicomObject

### LINKED ISSUE #78: Add take_element methods for InMemoryDicomObject
Allow the user of `dicom_object` to take the element from the object, removing it in the process.

For in-memory objects, this can translate to less data copying.

```rust
    /// <document please>
    pub fn take_element(&mut self, tag: Tag) -> Result<InMemElement<D>> {
        todo!()
    }

    /// <document please>
    pub fn take_element_by_name(&mut self, name: &str) -> Result<InMemElement<D>> {
        let tag = self.lookup_name(name)?;
        self.take_element(tag)
    }
```

pjsier:
Closes #78. Adds `take_element` and `take_element_by_name` on `InMemoryDicomObject` to take elements from an object and remove them. I based the doc comments and tests on the `element` method. Let me know if I should make any changes, thanks!

### DIFF
--- a/object/src/mem.rs
+++ b/object/src/mem.rs
@@ -336,6 +336,19 @@ where
         self.entries.insert(elt.tag(), elt)
     }
 
+    /// Removes and returns a particular DICOM element by its tag.
+    pub fn take_element(&mut self, tag: Tag) -> Result<InMemElement<D>> {
+        self.entries
+            .remove(&tag)
+            .context(NoSuchDataElementTag { tag })
+    }
+
+    /// Removes and returns a particular DICOM element by its name.
+    pub fn take_element_by_name(&mut self, name: &str) -> Result<InMemElement<D>> {
+        let tag = self.lookup_name(name)?;
+        self.take_element(tag)
+    }
+
     // private methods
 
     /// Build an object by consuming a data set parser.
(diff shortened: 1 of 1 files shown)


object/src/mem.rs
```diff
@@ -336,6 +336,19 @@ where
...
+    pub fn take_element(&mut self, tag: Tag) -> Result<InMemElement<D>> {
+        self.entries
+            .remove(&tag)
+            .context(NoSuchDataElementTag { tag })
+    }
+
+    /// Removes and returns a particular DICOM element by its name.
+    pub fn take_element_by_name(&mut self, name: &str) -> Result<InMemElement<D>> {
+        let tag = self.lookup_name(name)?;
+        self.take_element(tag)
```
Enet4, line 349: Hmm, I have reminded myself on why I chose to replicate the implementation of `element` and `element_by_name`, instead of just delegating one to the other. If `take_element` fails, it returns an error of variant `NoSuchElementTag`, which is less informative than the variant `NoSuchElementAlias`.

Please see how it was done in `element_by_name` and ensure that the error returned contains the attribute alias. Let me know if further clarification is needed.

object/src/mem.rs
```diff
@@ -570,6 +583,34 @@ mod tests {
...
+        let another_patient_name = DataElement::new(
+            Tag(0x0010, 0x0010),
+            VR::PN,
+            PrimitiveValue::Str("Doe^John".to_string()).into(),
+        );
+        let mut obj = InMemDicomObject::create_empty();
+        obj.put(another_patient_name.clone());
+        let elem1 = obj.take_element_by_name("PatientName").unwrap();
+        assert_eq!(elem1, another_patient_name);
+        assert!(obj.take_element_by_name("PatientName").is_err());
```
Enet4, line 611: Changing this assertion will allow us to verify the issue above.

```suggestion
        assert!(matches!(
            obj.take_element_by_name("PatientName"),
            Err(Error::NoSuchDataElementAlias {
                tag: Tag(0x0010, 0x0010),
                alias,
                ..
            }) if alias == "PatientName")
        );
```

object/src/mem.rs
```diff
@@ -570,6 +583,34 @@ mod tests {
...
+        let another_patient_name = DataElement::new(
+            Tag(0x0010, 0x0010),
+            VR::PN,
+            PrimitiveValue::Str("Doe^John".to_string()).into(),
+        );
+        let mut obj = InMemDicomObject::create_empty();
+        obj.put(another_patient_name.clone());
+        let elem1 = obj.take_element(Tag(0x0010, 0x0010)).unwrap();
+        assert_eq!(elem1, another_patient_name);
+        assert!(obj.take_element(Tag(0x0010, 0x0010)).is_err());
```
Enet4, line 597:
```suggestion
        assert!(matches!(
            obj.take_element(Tag(0x0010, 0x0010)),
            Err(Error::NoSuchDataElementTag {
                tag: Tag(0x0010, 0x0010),
                ..
            })
        ));
```

Enet4 (commented):
Thanks for working on this! I detected a minor concern that I would like resolved before merging, please see the details below.

object/src/mem.rs
pjsier: That makes sense, thanks for the feedback! I'll update that

object/src/mem.rs
pjsier: Thanks for these suggestions! I'm still getting adjusted to Rust and wasn't familiar with the `matches!` macro, but it seems really useful

Enet4 (approved):
Looks good, thank you!

Issue #71, opened 2020-09-27
Colored dcmdump output


Enet4:
It would be interesting to have the project's dcmdump print colored output for better readability and analysis.

This would give tag numbers, attribute names, and respective values, in different colours if the output device is a TTY.

[`termcolor`](https://crates.io/crates/termcolor) seems to be the right way to do this.

pjsier:
@Enet4 I'm interested in taking this on! To clarify, in the example you mentioned there would be 3 colors assigned, one each for numbers, attribute names, and values? And do you have any preference on the colors used?

Enet4:
Yes, the colours would differ among different components in a line.

To be honest, I did not specify concrete colours because this is likely going to be an iterative process for obtaining a good palette. Still, here's what I would picture as a good starting point for a dark background:

- a slightly darker shade of grey for the tag numbers;
- known attribute names in bold if supported by the platform
- values in a brighter colour, which could even depend on the data type and hint their validity: white or yellowish white for string values, light green for dates and times, red for invalid dates/times (this is already checked by dcmdump).

But again, we'll only know if they work well once tried out.

I will be able to test this on both Linux and Windows.

pjsier:
Makes sense, thanks for clarifying! I can take an initial stab at some of the colors and then we can try them out to see what works

pjsier:
@Enet4 I've been taking a look at `termcolor`, and it looks like the main benefit compared to something like [`colored`](https://github.com/mackwic/colored) is support for earlier versions of Windows. Is support for Windows versions before 10 a priority here? If so I can keep going forward with `termcolor`, but wanted to check because it looks like using `colored` is simpler overall

Enet4:
That does seem to be an easier API. Feel free to use it for now. Then we can evaluate the outcomes based on platform support and performance.

Enet4:
Resolved in #80

PR #80, opened 2020-10-16
Add colored output for dcmdump

pjsier:
Progress on #71. I wanted to get some feedback on this initial approach before going much further.

I'm using an enum that implements `fmt::Display` with formatting set for each option. Not sure if I've added the right options to every instance of the values, but I think the overall setup should work. Also calling `condition::set_virtual_terminal` for improving Windows 10 compatibility.

Let me know what you think!

### DIFF
--- a/dcmdump/src/main.rs
+++ b/dcmdump/src/main.rs
@@ -28,8 +28,6 @@ const ERROR_NO: i32 = -1;
 const ERROR_READ: i32 = -2;
 /// Exit code for when an error emerged while dumping the file.
 const ERROR_PRINT: i32 = -3;
-/// Exit code for failure to set OS compatibility
-const ERROR_COMPAT: i32 = -4;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 enum DumpValue<T>
@@ -42,7 +40,7 @@ where
     Str(T),
     DateTime(T),
     Invalid(T),
-    Nothing(T),
+    Nothing,
 }
 
 impl<T> fmt::Display for DumpValue<T>
@@ -57,7 +55,7 @@ where
             DumpValue::Str(v) => v.to_string().yellow(),
             DumpValue::DateTime(v) => v.to_string().green(),
             DumpValue::Invalid(v) => v.to_string().red(),
-            DumpValue::Nothing(v) => v.to_string().dimmed(),
+            DumpValue::Nothing => "(no value)".dimmed(),
         };
         write!(f, "{}", value)
     }
@@ -100,8 +98,7 @@ fn os_compatibility() -> Result<(), ()> {
 
 fn main() {
     os_compatibility().unwrap_or_else(|_| {
-        println!("Error setting OS compatibility");
-        std::process::exit(ERROR_COMPAT);
+        eprintln!("Error setting OS compatibility for colored output");
     });
 
     let filename = ::std::env::args()
@@ -407,7 +404,7 @@ fn value_summary(value: &PrimitiveValue, vr: VR, max_characters: u32) -> DumpVal
         (Str(value), _) => {
             DumpValue::Str(cut_str(&format!("\"{}\"", value), max_characters).to_string())
         }
-        (Empty, _) => DumpValue::Nothing("".to_string()),
+        (Empty, _) => DumpValue::Nothing,
     }
 }
 
--- a/dcmdump/src/main.rs
+++ b/dcmdump/src/main.rs
@@ -57,7 +57,11 @@ where
             DumpValue::Invalid(v) => v.to_string().red(),
             DumpValue::Nothing => "(no value)".dimmed(),
         };
-        write!(f, "{}", value)
+        if let Some(width) = f.width() {
+            write!(f, "{:width$}", value, width = width)
+        } else {
+            write!(f, "{}", value)
+        }
     }
 }
 
--- a/dcmdump/src/main.rs
+++ b/dcmdump/src/main.rs
@@ -8,6 +8,7 @@
 //! ```none
 //! dcmdump <file.dcm>
 //! ```
+use colored::*;
 use dicom::core::dictionary::{DataDictionary, DictionaryEntry};
 use dicom::core::header::Header;
 use dicom::core::value::{PrimitiveValue, Value as DicomValue};
@@ -18,6 +19,7 @@ use dicom::object::{open_file, DefaultDicomObject, FileMetaTable, StandardDataDi
 use dicom::transfer_syntax::TransferSyntaxRegistry;
 use snafu::ErrorCompat;
 use std::borrow::Cow;
+use std::fmt;
 use std::io::{stdout, ErrorKind, Result as IoResult, Write};
 
 /// Exit code for missing CLI arguments or --help
@@ -26,6 +28,36 @@ const ERROR_NO: i32 = -1;
 const ERROR_READ: i32 = -2;
 /// Exit code for when an error emerged while dumping the file.
 const ERROR_PRINT: i32 = -3;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum DumpValue<T>
+where
+    T: ToString,
+{
+    TagNum(T),
+    Alias(T),
+    Num(T),
+    Str(T),
+    DateTime(T),
+    Invalid(T),
+}
+
+impl<T> fmt::Display for DumpValue<T>
+where
+    T: ToString,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let value = match self {
+            DumpValue::TagNum(v) => v.to_string().dimmed(),
+            DumpValue::Alias(v) => v.to_string().bold(),
+            DumpValue::Num(v) => v.to_string().blue(),
+            DumpValue::Str(v) => v.to_string().yellow(),
+            DumpValue::DateTime(v) => v.to_string().green(),
+            DumpValue::Invalid(v) => v.to_string().red(),
+        };
+        write!(f, "{}", value)
+    }
+}
 
 fn report<E: 'static>(err: E)
 where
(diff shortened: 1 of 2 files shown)


dcmdump/src/main.rs
```diff
@@ -52,7 +84,17 @@ where
     }
 }
 
+#[cfg(windows)]
+fn os_compatibility() {
+    control::set_virtual_terminal(true);
```
Enet4, line 89: Look out, this method returns a `Result` which should be handled. Try propagating it to this function and using the existing error report function.

dcmdump/src/main.rs
```diff
@@ -263,81 +312,93 @@ where
...
+        (Strs(values), _) => DumpValue::Str(format_value_list(values, max_characters, true)),
+        (Date(values), _) => DumpValue::DateTime(format_value_list(values, max_characters, true)),
+        (Time(values), _) => DumpValue::DateTime(format_value_list(values, max_characters, true)),
+        (DateTime(values), _) => {
+            DumpValue::DateTime(format_value_list(values, max_characters, true))
+        }
+        (Str(value), _) => {
+            DumpValue::Str(cut_str(&format!("\"{}\"", value), max_characters).to_string())
+        }
+        (Empty, _) => DumpValue::Num("".to_string()),
```
Enet4, line 401: How about a `DumpValue::Nothing` variant?

Enet4 (changes requested):
All right, thank you for working on this! Here's how it look on my Windows machine.

image

The colors seem mostly OK, performance is acceptable too. The trailing pad after the attribute keyword is gone now, but maybe we can make a compromise and set it to an average that would cover most name lengths?
I would also suggest trying a lighter blue for binary values, they are a bit hard to read right now.

Colors are disabled if not on a TTY, just as intended.

Other than that, I did a quick glance at the code, and it seems OK other than the issue noted inline. The idea of keeping color information in an enum is also a good way to prevent inconsistencies.

dcmdump/src/main.rs
pjsier: It looks like the [original method returns `Result<(), ()>`](https://github.com/mackwic/colored/blob/master/src/control.rs#L28), so I wasn't sure how to use the existing `report` function since that expects an error. I added an exit code instead, but let me know if there's a better approach!

pjsier:
Thanks for the feedback! I addressed some of the inline comments and I'll look more into the trailing pad and updating the blue for binary values, agreed that it's a bit hard to read

dcmdump/src/main.rs
```diff
@@ -40,6 +42,7 @@ where
     Str(T),
     DateTime(T),
     Invalid(T),
+    Nothing(T),
```
Enet4, line 45: I was thinking more along the lines of this. If there really is no value to dump, we don't need to take a value.

```suggestion
    Nothing,
```

dcmdump/src/main.rs
```diff
@@ -54,6 +57,7 @@ where
             DumpValue::Str(v) => v.to_string().yellow(),
             DumpValue::DateTime(v) => v.to_string().green(),
             DumpValue::Invalid(v) => v.to_string().red(),
+            DumpValue::Nothing(v) => v.to_string().dimmed(),
```
Enet4, line 60: Either one would be OK (don't know of a way to color the second option without allocating though).

```suggestion
            DumpValue::Nothing => return Ok(()),
```

```suggestion
            DumpValue::Nothing => return write!("(no value)"),
```

dcmdump/src/main.rs
```diff
@@ -85,15 +89,20 @@ where
...
 #[cfg(not(windows))]
-fn os_compatibility() {}
+fn os_compatibility() -> Result<(), ()> {
+    Ok(())
+}
 
 fn main() {
-    os_compatibility();
+    os_compatibility().unwrap_or_else(|_| {
+        println!("Error setting OS compatibility");
```
Enet4, line 103:
```suggestion
        eprintln!("Error setting OS compatibility for colored output");
```

dcmdump/src/main.rs
```diff
@@ -398,7 +407,7 @@ fn value_summary(value: &PrimitiveValue, vr: VR, max_characters: u32) -> DumpVal
         (Str(value), _) => {
             DumpValue::Str(cut_str(&format!("\"{}\"", value), max_characters).to_string())
         }
-        (Empty, _) => DumpValue::Num("".to_string()),
+        (Empty, _) => DumpValue::Nothing("".to_string()),
```
Enet4, line 410:
```suggestion
        (Empty, _) => DumpValue::Nothing,
```

dcmdump/src/main.rs
```diff
@@ -85,15 +89,20 @@ where
...
-fn os_compatibility() {}
+fn os_compatibility() -> Result<(), ()> {
+    Ok(())
+}
 
 fn main() {
-    os_compatibility();
+    os_compatibility().unwrap_or_else(|_| {
+        println!("Error setting OS compatibility");
+        std::process::exit(ERROR_COMPAT);
```
Enet4, line 104: Instead of exiting, could the application continue running without colored output?

dcmdump/src/main.rs
pjsier: Would this work since it's implemented on `&str`? Still trying to get a handle on allocation and strings though

```rust
DumpValue::Nothing => "(no value)".dimmed(),
```

pjsier:
I think I implemented the changes requested here! Also fixed the trailing padding after the attribute keyword. Not sure about whether the `DumpValue::Nothing` string implementation is right though

dcmdump/src/main.rs
```diff
@@ -199,7 +251,12 @@ where
                 dump_item(&mut *to, item, width, depth + 2)?;
             }
             to.write_all(&indent)?;
-            writeln!(to, "(FFFE, E0DD) SequenceDelimitationItem",)?;
+            writeln!(
+                to,
+                "{} {}",
+                DumpValue::TagNum("(FFFE, E0DD)"),
```
Enet4, line 257: Tags are displayed without spaces, one might as well fix existing inconsistencies.
```suggestion
                DumpValue::TagNum("(FFFE,E0DD)"),
```

dcmdump/src/main.rs
```diff
@@ -199,7 +251,12 @@ where
...
+                DumpValue::Alias("ItemDelimitationItem"),
```
Enet4, line 258:
```suggestion
                DumpValue::Alias("SequenceDelimitationItem"),
```

dcmdump/src/main.rs
```diff
@@ -263,81 +322,93 @@ where
     D: DataDictionary,
 {
     let indent: String = std::iter::repeat(' ').take((depth * 2) as usize).collect();
-    writeln!(to, "{}(FFFE,E000) na Item", indent)?;
+    writeln!(
```
Enet4, line 325: The indentation should go before the tag. Things look weird when dumping sequence items.

```none
(0018,6011) SequenceOfUltrasoundRegions         SQ (2 Items)
(FFFE,E000)     na Item
      (0018,6012) RegionSpatialFormat                 US (1,  2 bytes): 1
      (0018,6014) RegionDataType                      US (1,  2 bytes): 1
      (0018,6016) RegionFlags                         UL (1,  4 bytes): 3
(FFFE,E00D)     ItemDelimitationItem
(FFFE,E000)     na Item
```

dcmdump/src/main.rs
```diff
@@ -27,6 +29,42 @@ const ERROR_READ: i32 = -2;
...
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let value = match self {
+            DumpValue::TagNum(v) => v.to_string().dimmed(),
+            DumpValue::Alias(v) => v.to_string().bold(),
+            DumpValue::Num(v) => v.to_string().cyan(),
+            DumpValue::Str(v) => v.to_string().yellow(),
+            DumpValue::DateTime(v) => v.to_string().green(),
+            DumpValue::Invalid(v) => v.to_string().red(),
+            DumpValue::Nothing => "(no value)".dimmed(),
```
Enet4, line 58: `dimmed` seems to not make a difference from the rest of the text here. I'd say we either try a darker shade or remove the call to `dimmed()`.

Capture-dicom-rs-colored-dcmdump-2

dcmdump/src/main.rs
```diff
@@ -27,6 +29,42 @@ const ERROR_READ: i32 = -2;
...
+    Num(T),
+    Str(T),
+    DateTime(T),
+    Invalid(T),
+    Nothing,
+}
+
+impl<T> fmt::Display for DumpValue<T>
+where
+    T: ToString,
```
Enet4, line 48: I feel that one could avoid constructing heap allocated strings here, but we can leave that for another time if it requires too many changes.

Enet4 (changes requested):
All right, binary data is easier to read now.

I found a few more problems that should not be too hard to fix, then it's good to merge.

dcmdump/src/main.rs
pjsier: Good to know! I switched this to `italic` since I didn't want to hard-code a darker color in case someone is using a light theme, and it still needs to return a `ColoredString`

dcmdump/src/main.rs
pjsier: I'm interested in doing this, but I'm not sure how to go about it so it may make sense for a different issue? Not entirely sure how to convert some of the non-string types to `&str` or `String` without allocating in this context, but I'm likely missing something

Enet4: This could well be a limitation of `colored`. The struct `ColoredString` works with strings internally, whereas a possible implementation of a hypothetical _`ColorisedSomething`_ could work for any type implementing `Display`, without allocating a string. Sounds like an opportunity to take this concern to the respective [repository](https://github.com/mackwic/colored).

Enet4 (approved):
Thanks again!

PR #90, opened 2020-11-20
Implementation of getter method to clean strings

josepp-moreira:
Implementation of a getter method and tests to clean a string or multiple strings from unwanted whitespaces.

### DIFF
--- a/core/src/header.rs
+++ b/core/src/header.rs
@@ -1146,11 +1146,11 @@ mod tests {
     use crate::{dicom_value, DicomValue};
 
     #[test]
-    fn to_clean_string(){
+    fn to_clean_string() {
         let p = dicom_value!(U16, [256, 0, 16]);
         let val = DicomValue::new(p);
         let element = DataElement::new(Tag(0x0028, 0x3002), VR::US, val);
-        match element.to_clean_str(){
+        match element.to_clean_str() {
             Ok(value) => assert_eq!(value, "256\\0\\16"),
             Err(_) => {}
         }
--- a/core/src/header.rs
+++ b/core/src/header.rs
@@ -1150,10 +1150,10 @@ mod tests {
         let p = dicom_value!(U16, [256, 0, 16]);
         let val = DicomValue::new(p);
         let element = DataElement::new(Tag(0x0028, 0x3002), VR::US, val);
-        match element.to_clean_str() {
-            Ok(value) => assert_eq!(value, "256\\0\\16"),
-            Err(_) => {}
-        }
+        assert_eq!(
+            element.to_clean_str().unwrap(),
+            "256\\0\\16",
+        );
     }
 
     #[test]
--- a/core/src/header.rs
+++ b/core/src/header.rs
@@ -252,6 +252,11 @@ where
         self.value.to_str()
     }
 
+    /// Retrieves the element's value as a clean string
+    pub fn to_clean_str(&self) -> Result<Cow<str>, CastValueError> {
+        self.value.to_clean_str()
+    }
+
     /// Convert the full primitive value into raw bytes.
     ///
     /// String values already encoded with the `Str` and `Strs` variants
@@ -1138,6 +1143,18 @@ impl fmt::Display for Length {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::{dicom_value, DicomValue};
+
+    #[test]
+    fn to_clean_string(){
+        let p = dicom_value!(U16, [256, 0, 16]);
+        let val = DicomValue::new(p);
+        let element = DataElement::new(Tag(0x0028, 0x3002), VR::US, val);
+        match element.to_clean_str(){
+            Ok(value) => assert_eq!(value, "256\\0\\16"),
+            Err(_) => {}
+        }
+    }
 
     #[test]
     fn tag_from_u16_pair() {
--- a/core/src/value/primitive.rs
+++ b/core/src/value/primitive.rs
@@ -604,25 +604,19 @@ impl PrimitiveValue {
         }
     }
 
-    /// Getter method to provide a clean version of a string or multiple strings,
-    /// removing unwanted whitespaces
+    /// Removes unwanted whitespaces from a string or multiple strings
+    ///
+    /// String values that have unwanted whitespaces
     pub fn to_clean_str(&self) -> Cow<str> {
         match self {
             PrimitiveValue::Str(values) => {
-                let new_value = values.trim_matches(|c| c == ' ' || c == '\u{0}');
-                Cow::from(new_value)
+                Cow::from(values.trim_end_matches(|c| c == ' ' || c == '\u{0}'))
             },
             PrimitiveValue::Strs(values) => {
                 if values.len() == 1 {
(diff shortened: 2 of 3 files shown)


core/src/value/primitive.rs
```diff
@@ -604,6 +604,31 @@ impl PrimitiveValue {
         }
     }
 
+    /// Getter method to provide a clean version of a string or multiple strings,
+    /// removing unwanted whitespaces
+    pub fn to_clean_str(&self) -> Cow<str> {
+        match self {
+            PrimitiveValue::Str(values) => {
+                let new_value = values.trim_matches(|c| c == ' ' || c == '\u{0}');
+                Cow::from(new_value)
```
Enet4, line 613: We will need to use [`trim_end_matches`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.trim_end_matches) instead, because in some DICOM value representations, [leading spaces are assumed to be significant](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html).

```suggestion
                Cow::from(values.trim_end_matches(|c| c == ' ' || c == '\u{0}'))
```

core/src/value/primitive.rs
```diff
@@ -604,6 +604,31 @@ impl PrimitiveValue {
...
+                if values.len() == 1 {
+                    let new_value = values[0].trim_matches(|c| c == ' ' || c == '\u{0}');
+                    Cow::from(new_value)
+                } else {
+                    let mut new_vec = Vec::new();
+                    for i in values{
+                        new_vec.push(i.trim_matches(|c| c == ' ' || c == '\u{0}'));
+                    }
+                    let new_value = new_vec.iter().join("\\").to_owned();
+                    Cow::from(new_value)
```
Enet4, line 625: A good rule of thumb is that whenever you find yourself creating a vector and pushing elements into it in a loop, there is likely a more idiomatic way via the iterator API.

```rust
let new_vec: Vec<_> = values.iter()
    .map(|s| s.trim_end_matches(|c| c == ' ' || c == '\0'))
    .collect();
```

In this case however, you don't really need a vector at all, since join works with iterators as well.

```suggestion
                    Cow::Owned(
                        values
                            .iter()
                            .map(|s| s.trim_end_matches(|c| c == ' ' || c == '\0'))
                            .join("\\"),
                    )
```

core/src/value/primitive.rs
```diff
@@ -604,6 +604,31 @@ impl PrimitiveValue {
         }
     }
 
+    /// Getter method to provide a clean version of a string or multiple strings,
+    /// removing unwanted whitespaces
```
Enet4, line 608: Please consider following the documentation structure found in other methods (e.g. `to_str`), by adding a small description of the conversions made, and an example.

core/src/value/primitive.rs
```diff
@@ -2540,6 +2565,15 @@ mod tests {
...
     }
 
+    #[test]
+    fn primitive_value_to_clean_str(){
+        let value = PrimitiveValue::from("1.2.345\0".to_string());
+        assert_eq!(&value.to_clean_str(), "1.2.345");
+
+        let value = dicom_value!(Strs, ["ONE", "TWO", "THREE", "SIX "]);
+        assert_eq!(&value.to_clean_str(), "ONE\\TWO\\THREE\\SIX");
+    }
```
Enet4, line 2575: These cases are good, and can be copied over to the example doctest above. On the other hand, there is an opportunity to improve coverage here. Let's also add a case with leading spaces (which should be preserved) and another one for some other binary representation (which should still work even when the trimming is unnecessary).

core/src/value/primitive.rs
```diff
@@ -604,6 +604,31 @@ impl PrimitiveValue {
...
+                } else {
+                    let mut new_vec = Vec::new();
+                    for i in values{
+                        new_vec.push(i.trim_matches(|c| c == ' ' || c == '\u{0}'));
+                    }
+                    let new_value = new_vec.iter().join("\\").to_owned();
+                    Cow::from(new_value)
+                }
+            }
+            prim => Cow::from(prim.to_string()),
```
Enet4, line 628: We can delegate to `to_str` instead, which already returns a `Cow`.

```suggestion
            prim => prim.to_str(),
```

Enet4 (changes requested):
Thank you for working on this! Please see my first review iteration and make the changes requested.

This is also a method that can be implemented in `DicomValue`, and even in `DataElement`, through conditional delegation. Please see if you can do this and reach out if you have any questions about this.

core/src/value/primitive.rs
josepp-moreira: Thank you for the feedback! 
I've made some changes to the code and even added more tests with leading spaces. Could you provide, if possible, some test examples with DicomValue and DataElement? Thanks!

Enet4: Sure, for `DicomValue` and `DataElement`, it is sufficient to replicate the primitive values already tested in the `primitive` module and encapsulate them into the other types. `From` implementations are available for `DicomValue`, and constructors exist for both types.

For instance, the following attribute:
``` 
(0028,3002) LUTDescriptor                       US (3,  6 bytes): [256, 0, 16]
```

Would be represented as a data element like this:

```rust
let p = dicom_value!(U16, [256, 0, 16]);
let val = DicomValue::new(p);
let element = DataElement::new(Tag(0x0028, 0x3002), VR::US, val);
assert_eq!(element.to_clean_str()?, "256\\0\\16");
```

core/src/value/primitive.rs
```diff
@@ -604,25 +604,19 @@ impl PrimitiveValue {
         }
     }
 
-    /// Getter method to provide a clean version of a string or multiple strings,
-    /// removing unwanted whitespaces
+    /// Removes unwanted whitespaces from a string or multiple strings
+    ///
+    /// String values that have unwanted whitespaces
```
Enet4, line 609: This documentation looks incomplete. Please consider replicating the prose found in other methods such as `to_str`, and adapting it to this operation. A [doc-test](https://doc.rust-lang.org/rustdoc/documentation-tests.html) would be very nice to have too.

Enet4 (changes requested):
We are almost done!

Please don't forget to format the new code (via `rustfmt` or `cargo fmt`) for more consistency in style.

core/src/header.rs
```diff
@@ -1138,6 +1143,18 @@ impl fmt::Display for Length {
...
+
+    #[test]
+    fn to_clean_string() {
+        let p = dicom_value!(U16, [256, 0, 16]);
+        let val = DicomValue::new(p);
+        let element = DataElement::new(Tag(0x0028, 0x3002), VR::US, val);
+        match element.to_clean_str() {
+            Ok(value) => assert_eq!(value, "256\\0\\16"),
+            Err(_) => {}
+        }
```
Enet4, line 1156: We don't want to ignore the error, because the operation should succeed. In this case we can just `unwrap` the result.

```suggestion
        assert_eq!(
            element.to_clean_str().unwrap(),
            "256\\0\\16",
        );
```

Enet4 (changes requested):

Enet4 (approved):
Great! Thank you!

PR #94, opened 2021-01-12
Re-export dicom_value macro in parent crate

Enet4:


parent/src/lib.rs
```diff
@@ -8,3 +8,7 @@ pub use dicom_encoding as encoding;
 pub use dicom_object as object;
 pub use dicom_parser as parser;
 pub use dicom_transfer_syntax_registry as transfer_syntax;
+
+// re-export dicom_value macro
+#[macro_export]
+pub use dicom_core::dicom_value;
```
Enet4, line 14: The attribute is not needed.

```suggestion
pub use dicom_core::dicom_value;
```

PR #96, opened 2021-01-15
Add write_dataset_with_ts to in-mem object

Enet4:
- add example to root object docs
- also improve documentation of `read_dataset` and `write_dataset`

### DIFF
--- a/object/src/mem.rs
+++ b/object/src/mem.rs
@@ -133,8 +133,10 @@ impl InMemDicomObject<StandardDataDictionary> {
 
     /// Read an object from a source using the given decoder.
     ///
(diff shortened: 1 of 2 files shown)


object/src/mem.rs
```diff
@@ -444,8 +450,10 @@ where
     /// with the specified encoder specifications,
     /// without preamble, magic code, nor file meta group.
     ///
-    /// Note: [`write_dataset_with_ts_cs`] may be easier to use.
+    /// Note: [`write_dataset_with_ts_cs`] and [`write_dataset_with_ts_cs`]
```
Enet4, line 453:
```suggestion
    /// Note: [`write_dataset_with_ts`] and [`write_dataset_with_ts_cs`]
```

PR #98, opened 2021-01-22
[ul] Association API

Enet4:
This is a work in progress for a major extension to the `ul` crate that provides facilities to establishing DICOM associations.

The first usable iteration will be based on synchronous IO, but the possibility of an asynchronous version is not out of the question.


ul/src/association/client.rs
```diff
@@ -0,0 +1,396 @@
...
+            | pdu @ Pdu::ReleaseRP { .. } => UnexpectedResponse { pdu }.fail(),
+            pdu @ Pdu::Unknown { .. } => UnknownResponse { pdu }.fail(),
+        }
+    }
+}
+
+/// A DICOM upper level association from the perspective
+/// of an association requester.
+#[derive(Debug)]
+pub struct ClientAssociation {
```
Enet4, line 283: Note to self: an SCU may want to use multiple presentation contexts during the same association. The API should change to support this.

PR #108, opened 2021-02-23
MoveSCU implementation

josepp-moreira:
A command line interface for C-MOVE SCU.

### DIFF
--- a/movescu/Cargo.toml
+++ b/movescu/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "dicom-movescu"
+version = "0.1.0"
+authors = ["Eduardo Pinho <enet4mikeenet@gmail.com>, José Moreira <joseppmoreira@ua.pt>"]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/Enet4/dicom-rs"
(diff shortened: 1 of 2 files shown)


Enet4 (changes requested):
Thank you for working on this. The expectations of use are still a bit too off for the tool to enter the ecosystem. I have gathered the related comments below.

To create a query DICOM object file, you can follow the [advice from DCMTK regarding findscu](https://support.dcmtk.org/docs-354/findscu.html). This text file:

```none
# query patient names and IDs
(0008,0052) CS [PATIENT]     # QueryRetrieveLevel
(0010,0010) PN []            # PatientsName
(0010,0020) LO [P123]            # PatientID
```

can be converted to a DICOM file with the DCMTK command `dump2dcm`. The file would then serve as input to this tool (in this example, to move _all_ images of patient with the ID `P123`).

Other minor concerns are that this is still missing a readme file.

I would appreciate it if you could look into this in some of your spare time. Otherwise, I can put it in my backlog.

Enet4:
I'm afraid that this has diverged too much from the base project. It will need to be rebuilt from scratch, maybe after #463 so that we can couple the move SCU with a local store SCP.

Issue #111, opened 2021-03-28
Helping out with pixel data decoding


pevers:
Hi @Enet4 !

Thanks for this awesome library! 

I'm now using this library to generate thumbnails for a bunch of JPEG lossless compressed dicoms that I have. I created a simple abstraction here: https://github.com/pevers/dicom-pixel-data, using GDCM bindings to decode pixel data into a `DynamicImage`:

Example from the tests:
```rust
let obj = open_file(dicom_test_files::path(value).unwrap()).unwrap();
let image = obj.decode_pixel_data().unwrap().to_dynamic_image().unwrap();
image
    .save(format!(
        "out.png",
    ))
    .unwrap();
```

Not everything is supported yet, but it can at least decode a bunch of example lossless JPEG dicoms from the Pydicom test set. I created a thin GDCM wrapper because I don't think there is solid support for the wide variety of dicom compression formats in Rust.

**My question:** I read on roadmap that this is a requested feature. Can I help by creating a PR to this repo? Or is this already in the making? I'm fairly new to Rust so I guess it would require some extensive reviewing but I would be glad to help out a bit.

Thanks again for your work!

Cheers

Enet4:
Hello! 👋It's great to know DICOM-rs is presenting itself useful to you!

Yes, it is in the project's plans to provide a uniform way to decode images out of DICOM objects, and I am thrilled to already see more efforts towards this functionality!

The project has so far favoured the use of implementations in pure Rust, but I suppose that the GDCM wrapper could provide a way to bootstrap pixel data decoding, and pure implementations could be added over time in future iterations. The only thing that could become a challenge is the integration with the transfer syntax concept at `dicom-encoding`, but this one will likely need to be modified to make way for this feature anyway. (Any questions about this in particular, feel free to ask.)

If you are willing to send in a PR, I would be glad to review it on the various aspects, including on writing idiomatic Rust. I also have a few notes about a vision for the image abstraction API that I could share.

Enet4:
Thank you! Now with #112 merged, we can keep track of further developments at #125.

PR #112, opened 2021-04-03
Add dicom-pixeldata

pevers:
This PR adds `dicom-pixeldata` to support handling pixel data decoding. 

It is a rough start to get early feedback about the design. I've added a simple trait `PixelDecoder` on `DefaultDicomObject` to get a raw decoded representation of pixel data. I'm using a GDCM bindings to decode the pixel data: https://github.com/pevers/gdcm-rs . I mimicked Pydicom's GDCM handler: https://github.com/pydicom/pydicom/blob/30cbc79d45cf708fd4afcdd150adc47230d2078e/pydicom/pixel_data_handlers/gdcm_handler.py#L215 .

Decoded pixel data can then be used to create a `DynamicImage`.  The user can use it like this:

```rust
use dicom_object::open_file;
use dicom_pixeldata::PixelDecoder;

let obj = open_file("pydicom/JPEG2000.dcm").unwrap();
let image = obj.decode_pixel_data().unwrap().to_dynamic_image().unwrap();
image
   .save("out.png")
   .unwrap();
```

The tests contain a list of Dicoms that are correctly converted. There is also a large list of dicoms that are still unsupported. Mostly because of having a different color space or being multi-frame dicoms.

There are a couple of things that we might want to add in the future:
- Support multi-frame dicoms
- Support conversion for different colors spaces and SamplesPerPixel > 3
- Support for `ndarray`
- Swap out GDCM for pure Rust implementation
- Convert VOI, VOI LUT and LUT Modality
- Applying WindowWidth/WindowCenter

Related issue: https://github.com/Enet4/dicom-rs/issues/111


pixeldata/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+dicom-object = { path = "../object", version = "0.3.0" }
+dicom-core = { path = "../core", version = "0.3.0" }
+dicom-encoding = { path = "../encoding", version = "0.3.0" }
+dicom-transfer-syntax-registry = { path = "../transfer-syntax-registry", version = "0.3.0" }
+snafu = "0.6.8"
+byteorder = "1.4.3"
```
pevers, line 15: I couldn't use that older version that is used in the rest of dicom-rs because it doesn't have `read_i16_into`.

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,468 @@
...
+    },
+
+    #[snafu(display("Non supported GDCM TransferSyntax: {}", ts))]
+    GDCMNonSupportedTS {
+        source: gdcm_rs::InvalidGDCMTS,
+        ts: String,
+    },
+
+    #[snafu(display("Not supported: {}", message))]
+    NotSupported { message: String },
```
Enet4, line 55: The use of a generic string message here is a sign that more information can be structurally embedded into the error type. Maybe this error variants could be expanded into multiple more specific variants?

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,468 @@
...
+
+    #[snafu(display("Invalid BitsAllocated, must be 8 or 16"))]
+    InvalidBitsAllocated,
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// Decoded pixel data
+/// NOTE: Is it possible to reuse dicom::PixelData ?
+pub struct DecodedPixelData {
```
Enet4, line 71: Hmm yes, this type would simply make the other one obsolete. It was an experiential attempt which never really served a purpose until now, so I guess we can keep this one and remove the one from the object crate.

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,468 @@
...
+fn get_samples_per_pixel(obj: &DefaultDicomObject) -> Result<u16> {
+    Ok(obj
+        .element_by_name("SamplesPerPixel")
+        .context(MissingRequiredField)?
+        .uint16()
+        .context(CastValueError)?)
+}
+
+/// Get the BitsAllocated of the Dicom
+fn get_bits_allocated(obj: &DefaultDicomObject) -> Result<u16> {
```
Enet4, line 285: The `get_` prefix is not very idiomatic in Rust. https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter. This one for example could be just `bits_allocated`.

Enet4 (commented):
Thank you very much for the contribution so far, it looks promising!

I will only be doing a full review some time after Sunday, but for now I can leave some pointers.

The `encoding` crate has an incomplete interface for codec implementations for a transfer syntax. https://github.com/Enet4/dicom-rs/blob/master/encoding/src/transfer_syntax/mod.rs#L127 With this type and the transfer syntax registry, it is possible to know which transfer syntaxes are supported, and to what extent (whether it can also decode or encode pixel data). Right now, it seems that the transfer syntax could announce lack of image decoding, without being actually true thanks to the implementation provided here.

So, another decision should be made here: whether to integrate these concepts with the existing transfer syntax system, or strip this information from the transfer syntax definition. I'll continue thinking about this later.

pixeldata/Cargo.toml
pevers: Ah, cool. I didn't realize that. I thought they were the same.

I'm using the `byteorder` `BigEndian/LIttleEndian::read_u/i16_into`, to read in a complete stream of `Vec<u8>` into the correct type. The type is derived from the TransferSyntax. Can this be done with `byteordered` ?

pixeldata/README.md
```diff
@@ -0,0 +1,22 @@
+# dicom-pixeldata
+Pixel data handler for the dicom-rs crate.
+
+## Example
+```rust
+use dicom::open_file;
+use dicom_object::open_file;
+use dicom_pixeldata::PixelDecoder;
+
+fn main() {
```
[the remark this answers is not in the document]
pevers: I have converted it and hid the boilerplate

pixeldata/src/lib.rs
pevers: Good idea. I expanded it in more precise types.

pixeldata/Cargo.toml
pevers: If we add that to `byteordered` it means we can replace these long match statements with just 1 line: https://github.com/Enet4/dicom-rs/pull/112/files#diff-bccaaeeafb2e2a338072bbffc91a719da64711a1a879009e771cec2dbcef5148R128

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+    pub samples_per_pixel: u16,
+    pub bits_allocated: u16,
+    pub bits_stored: u16,
+    pub high_bit: u16,
+    pub pixel_representation: u16,
+}
+
+impl DecodedPixelData {
+    /// Convert decoded pixel data into a DynamicImage
+    pub fn to_dynamic_image(&self) -> Result<DynamicImage> {
```
pevers, line 99: Maybe this should become `Result<Vec<DynamicImage>>` because we should eventually support multi-frame dicoms. Or something like `Result<DicomImageContainer>`. Although an `ndarray` would probably be a cleaner abstraction (just like Pydicom).

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+                _ => {
+                    // RGB, YBR, etc. color space
+                    UnsupportedSamplesPerPixel {
+                        spp: self.samples_per_pixel,
+                    }
+                    .fail()?
+                }
+            },
+            16 => {
+                let mut dest = vec![0; self.data.len() / 2];
```
Enet4, line 124: The vector created here is not used when `pixel_representation` is 1, so this other vector could be moved into the "unsigned" scope.

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+                        }
+
+                        // Normalize values between 0 - u16::MAX
+                        // In reality we need to apply LUT, WindowWidth/WindowCenter etc.
+                        dest = normalize_i16(&signed_buffer);
+                    }
+                    _ => InvalidPixelRepresentation.fail()?,
+                }
+                let image_buffer: ImageBuffer<Luma<u16>, Vec<u16>> =
+                    ImageBuffer::from_raw(self.cols, self.rows, dest).unwrap();
```
Enet4, line 160: The error coming from `image` should be gracefully handled.

Enet4 (changes requested):
All right, here is another review iteration. Please let me know if something isn't clear or you would like more assistance.

Some of the suggestions emerged from running Clippy, which you can also try out locally. `cargo clippy`

pevers:
> All right, here is another review iteration. Please let me know if something isn't clear or you would like more assistance.
> 
> Some of the suggestions emerged from running Clippy, which you can also try out locally. `cargo clippy`

Thanks for the review @Enet4 ! I'll fix the feedback this weekend.

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+        // "pydicom/rtstruct.dcm",
+        // "pydicom/test-SR.dcm",
+        // "pydicom/vlut_04.dcm",
+    ])]
+    fn test_parse_dicom_pixel_data(value: &str) {
+        let test_file = dicom_test_files::path(value).unwrap();
+        println!("Parsing pixel data for {:?}", test_file);
+        let obj = open_file(test_file).unwrap();
+        let image = obj.decode_pixel_data().unwrap().to_dynamic_image().unwrap();
+        image
```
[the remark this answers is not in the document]
pevers: Agreed. We can then built a set of PNG converted slices that we can assert on in the tests.

pixeldata/src/lib.rs
pevers: It is used at the end because we normalize everything to u16 vectors. Otherwise we can't create an `Image` from it becauses it can only handle `u8` and `u16` vectors.

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+        .map(|p| (u16::MAX as f32 * (*p as f32 - min) / (max - min)) as u16)
+        .collect()
+}
+
+pub trait PixelDecoder {
+    /// Decode compressed pixel data
+    fn decode_pixel_data(&self) -> Result<DecodedPixelData>;
+}
+
+impl PixelDecoder for DefaultDicomObject {
```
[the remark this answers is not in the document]
pevers: I'm not sure if I completely follow.  When I change it to this:

```rust
impl<D> PixelDecoder for InMemDicomObject<D> where D: DataDictionary + Clone {
```

It is not possible to get the transfer syntax anymore from a `InMemDicomObject`. Maybe I'm a bit confused with the types. Can you provide a bit more detail how to achieve this?

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+    #[snafu(display("Multi-frame dicoms are not supported"))]
+    UnsupportedMultiFrame,
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// Decoded pixel data
+/// NOTE: Is it possible to reuse dicom::PixelData ?
+pub struct DecodedPixelData {
+    pub data: Vec<u8>,
```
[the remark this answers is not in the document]
pevers: `data` is holding the decoded pixel data coming back from GDCM. That is different from the raw Dicom byte slice. So I think it is not possible. Maybe I should rename it to `decoded_pixel_data`?

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+                    rows.into(),
+                    pi_type,
+                    ts_type,
+                    samples_per_pixel,
+                    bits_allocated,
+                    bits_stored,
+                    high_bit,
+                    pixel_representation,
+                );
+                decoded_frame.unwrap().to_vec()
```
pevers, line 253: I'll fix this as well by creating normal error handling in gdcm-rs

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+        let pixel_representation = pixel_representation(self)?;
+
+        let decoded_pixel_data = match pixel_data.value() {
+            Value::PixelSequence {
+                fragments,
+                offset_table: _,
+            } => {
+                if fragments.len() > 1 {
+                    // Bundle fragments and decode multi-frame dicoms
+                    UnsupportedMultiFrame.fail()?
```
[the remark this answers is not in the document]
pevers: I tried doing this. But it seems to be hard to decode just a fragment. It looks like Pydicom is adding all fragments and then sending it to GDCM to decode: https://github.com/pydicom/pydicom/blob/4f9c53f29e3c51476e33fdb1728c6491539a2b3e/pydicom/pixel_data_handlers/gdcm_handler.py#L127

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+        // "pydicom/ExplVR_BigEndNoMeta.dcm",
+        // "pydicom/ExplVR_LitEndNoMeta.dcm",
+        // "pydicom/JPEG-LL.dcm",                       // More than 1 fragment
+        // "pydicom/MR-SIEMENS-DICOM-WithOverlays.dcm", // Error with Endianness?
+        // "pydicom/MR2_J2KI.dcm",  // Multi-frame
+        // "pydicom/MR2_J2KR.dcm",
+        // "pydicom/MR2_UNCI.dcm",
+        // "pydicom/MR2_UNCR.dcm",
+        // "pydicom/MR_small_bigendian.dcm",            // Endianness error
+        // "pydicom/MR_small_expb.dcm",                 // Endianness error
```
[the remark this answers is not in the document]
pevers: You are right. I made a false assumption that I still had to correct for endianness when GDCM converts the slice. I removed the conversion step and now these examples run fine as well.

pevers:
Hi @Enet4 . I have addressed the feedback as best as I could. I am not sure how to make the `PixelDecoder` trait more useful by making it more generic. Maybe you can help out there if you have an idea.

pixeldata/src/lib.rs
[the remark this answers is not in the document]
Enet4: Right, an alternative would be making it a `Cow<'a, [u8]>`, but the benefits may be slim. Tricks for efficiency can be tried out at a later stage.

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,480 @@
...
+    UnsupportedSamplesPerPixel { spp: u16 },
+
+    #[snafu(display("Unsupported TransferSyntax {}", ts))]
+    UnsupportedTransferSyntax { ts: String },
+
+    #[snafu(display("Multi-frame dicoms are not supported"))]
+    UnsupportedMultiFrame,
+
+    #[snafu(display("Invalid buffer when constructing ImageBuffer"))]
+    InvalildImageBuffer,
```
Enet4, line 80: Typo.
```suggestion
    InvalidImageBuffer,
```

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,480 @@
...
+            .fail()?
+        }
+
+        match self.bits_allocated {
+            8 => match self.samples_per_pixel {
+                1 => {
+                    // Single grayscale channel
+                    let image_buffer: ImageBuffer<Luma<u8>, Vec<u8>> =
+                        ImageBuffer::from_raw(self.cols, self.rows, self.data.to_owned())
+                            .context(InvalildImageBuffer)?;
```
Enet4, line 119:
```suggestion
                            .context(InvalidImageBuffer)?;
```

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,480 @@
...
+
+                        // Normalize values between 0 - u16::MAX
+                        // In reality we need to apply LUT, WindowWidth/WindowCenter etc.
+                        dest = normalize_i16(&signed_buffer);
+                    }
+                    _ => InvalidPixelRepresentation.fail()?,
+                }
+                let image_buffer: ImageBuffer<Luma<u16>, Vec<u16>> =
+                    ImageBuffer::from_raw(self.cols, self.rows, dest)
+                        .context(InvalildImageBuffer)?;
```
Enet4, line 155:
```suggestion
                        .context(InvalidImageBuffer)?;
```

Enet4 (changes requested):
We are getting closer to a baseline. Would it be OK for you to also implement `to_ndarray(&self) -> Array<IxDyn>` for `DecodedPixelData`? Otherwise, it would be time to address just a few more minor issues and split upcoming work into tasks!

pevers:
> We are getting closer to a baseline. Would it be OK for you to also implement `to_ndarray(&self) -> Array<IxDyn>` for `DecodedPixelData`? Otherwise, it would be time to address just a few more minor issues and split upcoming work into tasks!

Thanks for the review! I'll look into implementing this this weekend and then I'll fix the feedback as well.

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,480 @@
...
+// Noramlize u16 vector using min/max normalization
+fn normalize_u16(i: &[u16]) -> Vec<u16> {
+    let min = *i.iter().min().unwrap() as f32;
+    let max = *i.iter().max().unwrap() as f32;
+    i.par_iter()
+        .map(|p| (u16::MAX as f32 * (*p as f32 - min) / (max - min)) as u16)
+        .collect()
+}
+
+pub trait PixelDecoder {
```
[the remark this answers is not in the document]
pevers: Sounds good to me! That would make it easy to start implementing pure Rust versions of different transfer syntax encodings.

pevers:
Hi @Enet4 ,

I have implemented the feedback in commit: https://github.com/Enet4/dicom-rs/pull/112/commits/6ba3ab533f162ac5972c8b23d51091e8368012c1

I have added a conversion method for `ndarray` here: https://github.com/Enet4/dicom-rs/pull/112/commits/ac660ed6a4e3daacdf91d608a3bc541b70904357

I have created an enum wrapper so that the user of the library can match against the layout of the pixel data:

```
/// Different memory representations for the decoded pixel data
/// as ndarray
#[derive(Debug)]
pub enum NumArcArrayD {
    UnsignedByte(ArcArray<u8, IxDyn>),
    UnsignedWord(ArcArray<u16, IxDyn>),
    SignedByte(ArcArray<i8, IxDyn>),
    SignedWord(ArcArray<i16, IxDyn>),
}
```

And I have tested the shape and dimensions of the array by loading in the same file using Numpy and validating that the same pixel data and shape is returned.

This week I'll try to add another example for ndarray (if it looks ok). And I forgot to validate if 8-bit pixel data is in 2s complement https://github.com/Enet4/dicom-rs/pull/112/files#diff-bccaaeeafb2e2a338072bbffc91a719da64711a1a879009e771cec2dbcef5148R138. I'll try to do that coming week. Out now for some 🌞 .

Enet4 (commented):
Looking great so far!

Eh, I have been reminded of the possible output element type differences when producing an ndarray. That is one possible way of handling it. In [`nifti-rs`](https://github.com/Enet4/nifti-rs/blob/master/src/volume/ndarray.rs#L39-L41), I let the caller pick the element type directly as a parameter type, leading to numerical conversions underneath if the source data element type differs from the target type. This approach is pretty practical to the library user, albeit at the cost of added complexity for implementing the conversions at this level, and the original nature of the data being somewhat hidden. Such methods could be added in the future, but we can go with the suggested impl for now, as we understand that the set of methods is likely something that we want to continuously build upon anyway. :+1:

pevers:
> Looking great so far!
> 
> Eh, I have been reminded of the possible output element type differences when producing an ndarray. That is one possible way of handling it. In [`nifti-rs`](https://github.com/Enet4/nifti-rs/blob/master/src/volume/ndarray.rs#L39-L41), I let the caller pick the element type directly as a parameter type, leading to numerical conversions underneath if the source data element type differs from the target type. This approach is pretty practical to the library user, albeit at the cost of added complexity for implementing the conversions at this level, and the original nature of the data being somewhat hidden. Such methods could be added in the future, but we can go with the suggested impl for now, as we understand that the set of methods is likely something that we want to continuously build upon anyway. 👍

Ah that is a smart solution. It is indeed a bit hard at the moment to use the library. Automatically converting it would work for most cases I think.

pevers:
Hi @Enet4, I found it hard to come up with a nice short example. I came up with this:

```rust
use std::error::Error;
use dicom_object::open_file;
use dicom_pixeldata::{PixelDecoder, NumArrayD};

fn main() -> Result<(), Box<dyn Error>> {
    let obj = open_file("rgb_dicom.dcm")?;
    let pixel_data = obj.decode_pixel_data()?;
    let num_array_d = pixel_data.to_ndarray()?;
    
    let shape = match num_array_d {
        NumArrayD::UnsignedByte(ndarray) => ndarray.shape().to_owned(),
        NumArrayD::UnsignedWord(ndarray) => ndarray.shape().to_owned(),
        NumArrayD::SignedWord(ndarray) => ndarray.shape().to_owned(),
        _ => panic!("Unknown shape")
    };
    println!("Shape of the dicom {:?}", shape); // [100, 100, 3]
    Ok(())
}
```

It kind of shows why it is hard to use 😅 . I agree that we can continue with this implementation and then later on decide if we want to convert it automatically or leave that up to the user. Can you think of a better example or should I add this to the docs?

I removed the `SignedByte`. I don't think I've ever seen a 8-bit 2s complement dicom. But we can add that if we find these scans.

Enet4:
> It kind of shows why it is hard to use 😅 . I agree that we can continue with this implementation and then later on decide if we want to convert it automatically or leave that up to the user. Can you think of a better example or should I add this to the docs?

Ah, I understand the pain here. With this enum, we're basically reinventing a multi-value type like `PrimitiveValue`, but for _n_d arrays. For such a type to be nice to use, it requires a lot of boilerplate. Indeed, [primitive.rs](https://github.com/Enet4/dicom-rs/blob/master/core/src/value/primitive.rs) from the dicom-core crate is huge.

For now, we might be better served with methods which require the user to statically indicate the target element type: `to_ndarray_u8`, `to_ndarray_u16`, to_ndarray_f32` etc.

>
> I removed the SignedByte. I don't think I've ever seen a 8-bit 2s complement dicom. But we can add that if we find these scans.

Truth be told, I don't think I've seen one of that sort either. Regardless, we will not need the enum anymore if we follow the suggestion above.

---

Another undecided factor is how to name and design the functions which perform only one part of the pixel data processing pipeline.

- As the purpose of retrieving an _n_d-array is usually to perform further processing, applying the VOI LUT function might be undesirable for the reason that it is potentially a lossy operation. While I understand that the user may want to apply the VOI LUT function afterwards, I would still recommend that the default behavior of `to_ndarray_#` be to apply the Modality LUT function only (basic support means applying the linear operation based on rescale slope and rescale intercept), and perhaps provide helper functions to assist in the application of the VOI LUT.
- Considering that building an image value implies the intent to present it in a visible way, the current implementation of `to_image` may perform all the necessary steps up to the VOI LUT function (additional steps exist, but are use case dependent and are best left as explicit post procedures).

If followed in this manner, both methods would need to be carefully documented.

pevers:
> > It kind of shows why it is hard to use 😅 . I agree that we can continue with this implementation and then later on decide if we want to convert it automatically or leave that up to the user. Can you think of a better example or should I add this to the docs?
> 
> Ah, I understand the pain here. With this enum, we're basically reinventing a multi-value type like `PrimitiveValue`, but for _n_d arrays. For such a type to be nice to use, it requires a lot of boilerplate. Indeed, [primitive.rs](https://github.com/Enet4/dicom-rs/blob/master/core/src/value/primitive.rs) from the dicom-core crate is huge.
> 
> For now, we might be better served with methods which require the user to statically indicate the target element type: `to_ndarray_u8`, `to_ndarray_u16`, and `to_ndarray_f32` would be fair candidates for a first revision.
> 

Makes sense 👍 .

> > I removed the SignedByte. I don't think I've ever seen a 8-bit 2s complement dicom. But we can add that if we find these scans.
> 
> Truth be told, I don't think I've seen one of that sort either. Regardless, we will not need the enum anymore if we follow the suggestion above.
> 
> Another undecided factor is how to name and design the functions which perform only one part of the pixel data processing pipeline.
> 
> * As the purpose of retrieving an _n_d-array is usually to perform further processing, applying the VOI LUT function might be undesirable for the reason that it is potentially a lossy operation. While I understand that the user may want to apply the VOI LUT function afterwards, I would still recommend that the default behavior of `to_ndarray_#` be to apply the Modality LUT function only (basic support means applying the linear operation based on rescale slope and rescale intercept), and perhaps provide helper functions to assist in the application of the VOI LUT.

Applying the modality LUT could potentially make it a `f64` type. So that may be unexpected if you want to handle the raw pixel data. So maybe the Modality LUT should be a function too.

I think I'm leaning towards the nifti-rs approach you mentioned. It will make the conversion if it is needed, but is still performant if the user knows the input data. It will make the library much more usable because pixel manipulation works out of the box if you don't care that much about the memory footprint.

```
// f32 probably works for most dicoms if you want to manipulate pixel data
let ndarray: Array<f32, IxDyn> = pixel_data.to_ndarray(); // automatic conversion
...
let ndarray: Array<f32, IxDyn> = apply_modality_lut(ndarray); // Panics or errors if it doesn't fit?
```

If the user does care about optimal performance they can implement the different methods based on the `bits_allocated`?

> * Considering that building an image value implies the intent to present it in a visible way, the current implementation of `to_image` may perform all the necessary steps up to the VOI LUT function (additional steps exist, but are use case dependent and are best left as explicit post procedures).
> 
> If followed in this manner, both methods would need to be carefully documented.

+1 for helper functions for the VOI LUT. That is also how Pydicom handles it. 

Applying the Modality LUT for `to_image` makes sense because the intent is to display an image. There is already a lossy conversion. I'll add that if you are OK with that.

Enet4:
All those points are well made. The rescale slope could lead to surprising results if we're not careful (while this one could be 1 most of the time, it is not necessarily so). So yes, we can go with a process similar to that of `nifti`. Let me know if you need assistance here.

> Applying the Modality LUT for to_image makes sense because the intent is to display an image. There is already a lossy conversion.

Oh yes, certainly. It is was I meant with "all the necessary steps up to" :)

pevers:
Hi @Enet4 ,

I have modified `to_ndarray` to convert pixel data to the target type automatically using the `NumCast` trait from the `num-traits` crate in this commit: https://github.com/Enet4/dicom-rs/pull/112/commits/929d5d1fbbaedba8c0e46d7e435aa8a1400c2671 . 

Library usage is now much simpler, although at the cost of always converting regardless if this is needed or not:
```
        let ndarray = obj
            .decode_pixel_data()
            .unwrap()
            .to_ndarray::<u16>()
            .unwrap();
```

If this is OK, I'll continue by implementing the Modality LUT methods for the `to_image` method.

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+    #[snafu(display("Missing required element"))]
+    MissingRequiredField { source: dicom_object::Error },
+
+    #[snafu(display("Could not cast element"))]
+    CastValueError {
+        source: dicom_core::value::CastValueError,
+    },
+
+    #[snafu(display("Non supported GDCM PhotometricInterpretation: {}", pi))]
+    GDCMNonSupportedPI {
```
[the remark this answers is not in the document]
Enet4: Just a note to address this one too. ☝️

pixeldata/src/lib.rs
```diff
@@ -0,0 +1,597 @@
...
+    #[test]
+    fn test_to_ndarray_error() {
+        let test_file = dicom_test_files::path("pydicom/JPEG2000.dcm").unwrap();
+        let obj = open_file(test_file).unwrap();
+        let ndarray = obj.decode_pixel_data().unwrap().to_ndarray::<u8>();
+        if let Err(_) = ndarray {
+            //
+        } else {
+            panic!("should fail");
+        }
```
Enet4, line 595: A combination of `assert!` and `matches!` (or `assert_matches!` through a third-party dependency) would make the assertion look better and enable us to be more specific about what kind of error is expected.

Enet4 (commented):
> If this is OK, I'll continue by implementing the Modality LUT methods for the `to_image` method.

That is great! Please proceed with that plan. :) Very minor comments follow. Now that `to_ndarray` exists, it would be nice to have an example at the root of the crate.

On a separate subject that you may like to know, I have planned to release a new version of the project (0.4) soon, with the current upstream (merged) contributions. Significant contributions include a revamped data element dictionary, reading/writing bare datasets, and a new DICOM network API, plus new CLI tools. I predict that some things around this topic (and the one around lazy loading, which I am working on) may take multiple stages, and so releasing a new version will let users take advantage of new ready-to-go features while these major topics are worked on under less pressure.

pixeldata/src/lib.rs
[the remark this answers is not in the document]
pevers: 👍  forgot about this one

pixeldata/src/lib.rs
pevers: Ah thanks!

pevers:
> On a separate subject that you may like to know, I have planned to release a new version of the project (0.4) soon, with the current upstream (merged) contributions. Significant contributions include a revamped data element dictionary, reading/writing bare datasets, and a new DICOM network API, plus new CLI tools. I predict that some things around this topic (and the one around lazy loading, which I am working on) may take multiple stages, and so releasing a new version will let users take advantage of new ready-to-go features while these major topics are worked on under less pressure.

Cool!

pevers:
Ok I have implemented the Modality LUT for `to_dynamic_image` @Enet4 .

I'm now converting the slice to `f64` by applying the `RescaleSlope` and `RescaleIntercept`. I made them default to `1.0` and `0` respectively. After that I normalize to `u16` to convert them into a `DynamicImage`.

It seems a bit involved, but I think it is the safest option.

Enet4 (approved):
All right, next week I will have the opportunity to wrap up this pull request and advance on a few more things around this project.  

So below are the topics which have been brought up on this proof of concept,
and which I will expand into separate issues:

- Support for the photometric interpretations `MONOCHROME1` and `RGB`
  would make a great difference.
- Extending the API for the application of Modality and VOI LUT functions
  will be essential,
  in order to provide accurate results without any surprises.
  This would also make way to a more thorough data oriented test harness.
  From what I gather:
   - `to_dynamic_image` only performs the Modality LUT
   - `to_ndarray` does no LUT processing
- Integration with the `encoding` crate's `TransferSyntax` system.
  This would enable separate crates to install new transfer syntax implementations
  into end programs.
  The way I currently imagine this is by
  creating a new extension trait for DICOM objects
  and making implementations rely on this trait for retrieving 
  all the necessary properties,
  such as the dimensions, photometric interpretation, and the pixel data proper.
- Independency from GDCM.
  Even if at the cost of supporting less encodings,
  there should be a way to make this abstraction work without it.
  Support for WebAssembly comes to mind, which becomes more complicated
  when linking to other native libraries is involved.
  Putting gdcm behind a Cargo feature enabled by default seems ideal for now.
- _Support for multi-frame images._ API design decisions and implementations
  would be provided in their due time.
- There may be potential for efficiency optimizations,
  either by reducing the number of copies
  or changing how conversions between pixel data values are done.
  Still, it would be nice to have a benchmark before that
  (maybe using `criterion`).

Please let me know if you plan to change anything, otherwise I can just merge and proceed with the plans above.

pevers:
> All right, next week I will have the opportunity to wrap up this pull request and advance on a few more things around this project.
> 
> So below are the topics which have been brought up on this proof of concept, and which I will expand into separate issues:
> 
> * Support for the photometric interpretations `MONOCHROME1` and `RGB` would make a great difference.

Yep! I can pick this up in a follow-up PR. I think it is relatively easy to implement.

> * Extending the API for the application of Modality and VOI LUT functions will be essential, in order to provide accurate results without any surprises. This would also make way to a more thorough data oriented test harness. From what I gather:
>   
>   * `to_dynamic_image` only performs the Modality LUT
>   * `to_ndarray` does no LUT processing
> * Integration with the `encoding` crate's `TransferSyntax` system. This would enable separate crates to install new transfer syntax implementations into end programs. The way I currently imagine this is by creating a new extension trait for DICOM objects and making implementations rely on this trait for retrieving all the necessary properties, such as the dimensions, photometric interpretation, and the pixel data proper.
> * Independency from GDCM. Even if at the cost of supporting less encodings, there should be a way to make this abstraction work without it. Support for WebAssembly comes to mind, which becomes more complicated when linking to other native libraries is involved. Putting gdcm behind a Cargo feature enabled by default seems ideal for now.
> * _Support for multi-frame images._ API design decisions and implementations would be provided in their due time.
> * There may be potential for efficiency optimizations, either by reducing the number of copies or changing how conversions between pixel data values are done. Still, it would be nice to have a benchmark before that (maybe using `criterion`).
> 

This sounds all good to me! We can split it up in issues and I can help out with most of the issues. 

> Please let me know if you plan to change anything, otherwise I can just merge and proceed with the plans above.

I have no more pending changes. I just removed the Travis badge and the `.gitignore`. Thanks a lot for the extensive review. I learned a lot and I would be glad to help out with these additions!

pixeldata/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "dicom-pixeldata"
+version = "0.4.0"
```
Enet4, line 3:
```suggestion
version = "0.1.0"
```

Enet4 (approved):
Just a side note, I will be changing the version of the new crate to 0.1.0 to conform with the independent crate versioning currently applied in the rest of the DICOM-rs ecosystem (e.g. the latest version of `dicom-ul` is 0.3.0 even though DICOM-rs is at 0.4).

Great! This is ready to be merged, and new issues will be filed soon. Thank you very much for this contribution, @pevers! And I'm glad that you appreciated the process.

Issue #130, opened 2021-05-20
[pixeldata] Enable building for WebAssembly


Enet4:
This issue also depends on #121, but gating GDCM might be insufficient to ensure that it builds on WebAssembly. As `rayon` is used for parallel processing of pixel data, it may attempt to spawn threads, which [might be incompatible with this environment](https://rustwasm.github.io/book/reference/which-crates-work-with-wasm.html#spawning-threads).

pevers:
> This issue also depends on #121, but gating GDCM might be insufficient to ensure that it builds on WebAssembly. As `rayon` is used for parallel processing of pixel data, it may attempt to spawn threads, which [might be incompatible with this environment](https://rustwasm.github.io/book/reference/which-crates-work-with-wasm.html#spawning-threads).

Yes I think you are correct. Maybe we can configure the threadpool for rayon to just use the main thread.

Enet4:
> Yes I think you are correct. Maybe we can configure the threadpool for rayon to just use the main thread if we compile for wasm.

Ah, that too might work. I was picturing a feature gate on all of `rayon` and replace the algorithms to use plain iterators, but that would be less invasive.

pevers:
I found this crate: https://github.com/GoogleChromeLabs/wasm-bindgen-rayon

And it works!

![Screenshot 2021-05-20 at 21 46 50](https://user-images.githubusercontent.com/3428649/119040202-659ded80-b9b5-11eb-9e7c-d5e1a001ece4.png)

I did a test and managed to compile to WASM by making GDCM optional. My fork is here: https://github.com/pevers/dicom-rs/tree/pevers/gdcm-feature . I'll open a PR soon. It is work in progress.

And this is the test code: https://github.com/pevers/dicom-wasm

There is quite a list of caveats, but it doesn't require to touch the dicom-rs rayon implementation. And it should support multi-threading by webworkers 🎉 .

Issue #122, opened 2021-05-04
[pixeldata] API extension for selective application of Modality and VOI LUT functions


Enet4:
The current pixel data retrieval API does not have a clear way to specify specify whether to apply the Modality and/or VOI LUT functions.

```rust
/// Convert decoded pixel data into a DynamicImage.
/// A new <u8> or <u16> vector is created in memory
/// with normalized grayscale values after applying Modality LUT.
pub fn to_dynamic_image(&self) -> Result<DynamicImage>;

/// Convert decoded pixel data into an ndarray of a given type T.
/// The pixel data type is extracted from the bits_allocated and
/// pixel_representation, and automatically converted to the requested type T.
pub fn to_ndarray<T>(&self) -> Result<Array<T, IxDyn>>
```

- `to_dynamic_image` performs the Modality LUT function (currently only rescale and photometric interpretation adaptation for Monochrome1), followed by a non-standard min-max normalization.
- `to_ndarray` only reinterprets the pixel data element as a typed multi-dimensional array. No other transformation is applied.

This API should be extended to enable specifying the steps in the image processing pipeline to consider:

- obtain just a typed array (ndarray)
- calculate the Modality form of the data (ndarray)
- calculate the VOI form (ndarray, dynamic_image)
- perform all steps of the pipeline, with the default window level or one specified by the user (ndarray, dynamic_image)

Issue #120, opened 2021-05-04
[pixeldata] Integrate with encoding TransferSyntax


Enet4:
This would enable separate crates to install new transfer syntax implementations into end programs.
The way I currently imagine this is:

- Create a new extension trait for DICOM objects and making implementations rely on this trait for retrieving all the necessary properties, such as the dimensions, photometric interpretation, and the pixel data proper.
- Change the transfer syntax Codec concept so that it receives something that impls the trait above for retrieving/producing the image and/or frames of the image.

This is more of a stub, more details and acceptance criteria may come later.

PR #131, opened 2021-05-22
Make GDCM compilation optional

pevers:
This PR makes compiling with GDCM optional. This should enable WebAssembly support.

I tried compiling it to WebAssembly. And it works by using the [wasm-bindgen-rayon](https://github.com/GoogleChromeLabs/wasm-bindgen-rayon) crate.
Example can be found here: https://github.com/pevers/dicom-wasm

I've made GDCM a default feature. Once we implement native support we can ditch it and it shouldn't break SemVer if I'm correct.

I've changed the GitHub actions a bit to test `--no-default-features`. I have no clue if it is correct since I couldn't test it yet.

This should close: https://github.com/Enet4/dicom-rs/issues/130

### DIFF
--- a/pixeldata/Cargo.toml
+++ b/pixeldata/Cargo.toml
@@ -17,7 +17,7 @@ dicom-dictionary-std = { path = "../dictionary-std", version = "0.4.0" }
 snafu = "0.6.8"
 byteorder = "1.4.3"
 image = "0.23.14"
-gdcm-rs = "0.2.0"
+gdcm-rs = { version = "0.2.0", optional = true }
 rayon = "1.5.0"
 ndarray = "0.15.1"
 ndarray-stats = "0.5"
@@ -26,3 +26,7 @@ num-traits = "0.2.12"
 [dev-dependencies]
 dicom-test-files = "0.2.0"
 rstest = "0.6.4"
+
+[features]
+default = ["gdcm"]
+gdcm = ["gdcm-rs"]
--- a/pixeldata/Cargo.toml
+++ b/pixeldata/Cargo.toml
@@ -25,7 +25,7 @@ num-traits = "0.2.12"
 
 [dev-dependencies]
 dicom-test-files = "0.2.0"
-rstest = "0.6.4"
+rstest = "0.10.0"
 
 [features]
 default = ["gdcm"]
--- a/pixeldata/src/gdcm.rs
+++ b/pixeldata/src/gdcm.rs
@@ -65,7 +65,9 @@ where
                     high_bit,
                     pixel_representation,
                 )
-                .map_err(|_| Error::UnknownGdcmError)?;
+                .map_err(|source| Error::ImplementerError {
+                    source: Box::new(source),
+                })?;
                 decoded_frame.to_vec()
             }
             Value::Primitive(p) => {
@@ -99,7 +101,6 @@ mod tests {
     use rstest::rstest;
     use std::path::Path;
 
-    #[cfg(feature = "gdcm")]
     #[rstest(value => [
         "pydicom/693_J2KI.dcm",
         "pydicom/693_J2KR.dcm",
@@ -230,7 +231,7 @@ mod tests {
 ])]
     fn test_parse_dicom_pixel_data(value: &str) {
         let test_file = dicom_test_files::path(value).unwrap();
-        println!("Parsing pixel data for {:?}", test_file);
+        println!("Parsing pixel data for {}", test_file.display());
         let obj = open_file(test_file).unwrap();
         let image = obj.decode_pixel_data().unwrap().to_dynamic_image().unwrap();
         image
--- a/pixeldata/src/gdcm.rs
+++ b/pixeldata/src/gdcm.rs
@@ -1,0 +1,257 @@
+//! Decode pixel data using GDCM when the default features are enabled.
+
+use crate::*;
+use dicom_encoding::transfer_syntax::TransferSyntaxIndex;
+use dicom_transfer_syntax_registry::TransferSyntaxRegistry;
+use gdcm_rs::{decode_single_frame_compressed, GDCMPhotometricInterpretation, GDCMTransferSyntax};
+use std::str::FromStr;
+
+impl<D> PixelDecoder for FileDicomObject<InMemDicomObject<D>>
+where
+    D: DataDictionary + Clone,
+{
+    fn decode_pixel_data(&self) -> Result<DecodedPixelData> {
+        let pixel_data = self
+            .element(dicom_dictionary_std::tags::PIXEL_DATA)
+            .context(MissingRequiredField)?;
+        let cols = cols(self)?;
+        let rows = rows(self)?;
+
+        let photometric_interpretation = photometric_interpretation(self)?;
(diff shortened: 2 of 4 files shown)


pixeldata/Cargo.toml
```diff
@@ -26,3 +26,7 @@ num-traits = "0.2.12"
 [dev-dependencies]
 dicom-test-files = "0.2.0"
 rstest = "0.6.4"
```
Enet4, line 28: I recall that you introduced this dependency. I noticed that the latest version is at 0.10. How hard would it be to update it?

pixeldata/src/gdcm.rs
```diff
@@ -0,0 +1,257 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use dicom_object::open_file;
+    use dicom_test_files;
+    use rstest::rstest;
+    use std::path::Path;
+
+    #[cfg(feature = "gdcm")]
```
Enet4, line 102: Was this needed? The module is only checked with the feature anyway.

pixeldata/src/gdcm.rs
```diff
@@ -0,0 +1,257 @@
...
+    // "pydicom/rtdose_rle_1frame.dcm",
+    // "pydicom/rtplan.dcm",
+    // "pydicom/rtplan_truncated.dcm",
+    // "pydicom/rtstruct.dcm",
+    // "pydicom/test-SR.dcm",
+    // "pydicom/vlut_04.dcm",
+])]
+    fn test_parse_dicom_pixel_data(value: &str) {
+        let test_file = dicom_test_files::path(value).unwrap();
+        println!("Parsing pixel data for {:?}", test_file);
```
Enet4, line 233: This should work, if you wish to keep it.
```suggestion
        println!("Parsing pixel data for {}", test_file.display());
```

pixeldata/src/lib.rs
```diff
@@ -103,7 +99,7 @@ pub enum Error {
     InvalidImageBuffer,
 
     #[snafu(display("Unknown GDCM error while decoding image"))]
-    UnknownGdcmError { source: gdcm_rs::Error },
+    UnknownGdcmError,
```
Enet4, line 102: At this point, I believe this would call for an opaque error source, so that the error type does not depend on the implementation of pixel data encoders and decoders. Could you experiment and see if something like this would work?

```suggestion
    ImplementerError { source: Box<dyn std::error::Error + Send + 'static> },
```

Enet4 (commented):
Thank you for the contribution, @pevers.

> I've changed the GitHub actions a bit to test --no-default-features. I have no clue if it is correct since I couldn't test it yet.

That could work for now. There are other Cargo features to account for, but ensuring that the crates build without them is always important. In addition, back when the project was using Travis CI, I had set up a separate job for running the tests of some of the crates on a WebAssembly runtime. I will want to bring this back eventually.

Things look good in general! Please see what can be improved around the comments below.

pixeldata/src/gdcm.rs
pevers: No you are right. This can be removed.

pixeldata/src/lib.rs
pevers: Yes, that would make sense to make it an opaque error. I have changed it!

pevers:
> Thank you for the contribution, @pevers.
> 
> > I've changed the GitHub actions a bit to test --no-default-features. I have no clue if it is correct since I couldn't test it yet.
> 
> That could work for now. There are other Cargo features to account for, but ensuring that the crates build without them is always important. In addition, back when the project was using Travis CI, I had set up a separate job for running the tests of some of the crates on a WebAssembly runtime. I will want to bring this back eventually.
> 
> Things look good in general! Please see what can be improved around the comments below.

Hi @Enet4 , thanks for taking the time to review. I've processed the suggested feedback.

Enet4 (approved):
Looks good, @pevers, thanks again for your continued support!

In terms of this crate, it's a good time for me to expand on #120 and #122.

Enet4:
@desprit That appears to fail when retrieving a Rayon registry, which I believe is filled in by `wasm-bindgen-rayon` in this environment. Does the [example above](https://github.com/pevers/dicom-wasm) work for you?

If that still does not work depending on some quirk in your scenario, then we may need to also feature-gate rayon.

pevers:
> @pevers
> 
> Thanks for making GDCM optional! I was able to run it. One issue I'm facing in my WASM code is with `to_dynamic_image` method. When I call it I'm getting an error, here is the trace:
> 
> ```
> Uncaught (in promise) RuntimeError: unreachable
>     at __rust_start_panic (http://0.0.0.0:4040/0f00cc9c536f2b02b25c.module.wasm:wasm-function[14273]:0x2be244)
>     at rust_panic (http://0.0.0.0:4040/0f00cc9c536f2b02b25c.module.wasm:wasm-function[13011]:0x2b5e6e)
>     at std::panicking::rust_panic_with_hook::h9d91fa0fae16500f (http://0.0.0.0:4040/0f00cc9c536f2b02b25c.module.wasm:wasm-function[2935]:0x1bdb5f)
>     at std::panicking::begin_panic_handler::{{closure}}::h2d82b5b2db976f3f (http://0.0.0.0:4040/0f00cc9c536f2b02b25c.module.wasm:wasm-function[4765]:0x20ec83)
>     at std::sys_common::backtrace::__rust_end_short_backtrace::h24539600b7e2452c (http://0.0.0.0:4040/0f00cc9c536f2b02b25c.module.wasm:wasm-function[12687]:0x2b27ae)
>     at rust_begin_unwind (http://0.0.0.0:4040/0f00cc9c536f2b02b25c.module.wasm:wasm-function[11573]:0x2a4591)
>     at core::panicking::panic_fmt::h91d2023e5afe1929 (http://0.0.0.0:4040/0f00cc9c536f2b02b25c.module.wasm:wasm-function[12689]:0x2b2812)
>     at core::option::expect_none_failed::h203584ddbc785ca2 (http://0.0.0.0:4040/0f00cc9c536f2b02b25c.module.wasm:wasm-function[5643]:0x22b64e)
>     at core::result::Result<T,E>::expect::h9b350c7ce946dd9c (http://0.0.0.0:4040/0f00cc9c536f2b02b25c.module.wasm:wasm-function[4534]:0x2067ec)
>     at rayon_core::registry::global_registry::ha70428e6b344dc1e (http://0.0.0.0:4040/0f00cc9c536f2b02b25c.module.wasm:wasm-function[5250]:0x21f196)
> ```
> 
> I'm not unwrapping the result, just calling the function. Tested it on various dicom files from the dicom-test-files crate:
> `MR_small_jpeg_ls_lossless`
> `693_UNCI`
> `SC_rgb_16bit`
> 
> It only happens in WASM, works fine outside of it. My end goal is to make dicom-rs part work as a loader for cornerstone which has pixel data stored as `Uint16Array`.

You can enable debugging so that it will show you a valuable error instead of `unreachable`. See the `panic` hook here: https://github.com/pevers/dicom-wasm

It can only handle a small selection of scans (uncompressed). You also need a shim for `rayon`'s multi-threading. See my example: https://github.com/pevers/dicom-wasm .

Issue #55, opened 2020-06-23
Distinguish null components in date/time values


Enet4:
As per the standard, [part 5, section 6.2](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html):

> A component that is omitted from the string is termed a null component. Trailing null components of Date Time indicate that the value is not precise to the precision of those components. The YYYY component shall not be null. Non-trailing null components are prohibited. The optional suffix is not considered as a component.

As the variants of a `PrimitiveValue` currently decode date and time values into `chrono` date/time types, we no longer hold the information of which components were null once decoded from their textual form. The null components are reverted to the "first" values when decoding a date with these: january, day 1, 0 hours, 0 minutes, etc.

Since textual representations are already preserved default (this is done in #9), the DICOM parser will choose to keep DA, TM, and DT as plain text, and thus no information will be lost. Still, it is best to also have some way to store these values in a binary format, one that retains information about null components while still enabling an intuitive conversion to a standard date/time type for subsequent manipulation.

---

One acceptable way of resolving this would be:

- New types (maybe `PartialDate`) with multiple variants depending on how many components are defined. They might best be implemented as enums, and would have to have several utility methods in order to be usable.
- New methods `.to_partial_date`, `.to_partial_time`, `.to_partial_datetime` on DICOM value types.

jmlaka:
So I looked at the [dicom spec](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html) once more and
Date has a value length of 8 bytes FIXED (apart from range matching), so parsing of and incomplete Date should maybe fail ?

Time and DateTime mention the null components ...

Do you think that this problem could be solved like this : 

At the level of `deserialize.rs` 
- parsing of incomplete DA, DT, TM would fail with something like `IncompleteValue`
- that means that in case of DT and TM it would fail for any other case than a full representation (so a lot)
- new methods `.to_partial_date`, `.to_partial_time`, `.to_partial_datetime` would return a new type like `PartialDate` ..,
a tuple of two (DA, DT, TM) values, that would represent a range that takes into account the imprecision caused by null components.

For a Date it should be easy, as `"2010"` would translate to `(2010-01-01,2010-12-31)`
A Time of  `"071010.123"` would translate to `( 07:10:10.123000, 07:10:10.123999)` ?
A DateTime `"2010"` would translate to 
[...]

PR #137, opened 2021-07-07
Update deserialize.rs

jmlaka:
parse_date for YYYYMM correction + test

(this is my first pull request ever, hope it worked, thanks)


Enet4 (approved):
Improvements for this API so that it keeps information about null date date components are planned (#55), but until then, it ought to actually enable users to fetch these dates with a reasonable default.

As such, this is spot on! Thank you for your contribution. :+1:

PR #139, opened 2021-07-08
simplify UTC offset parsing

jmlaka:
Hello, so I was looking further into your library and found this UTC offtset parsing.
The original line 252 is unreachable code because of line 247
I think that the parsing was intended to do:
+1 = 01:00
+01 = Err
+011 = 01:01
I suggest simplifying it to always force a 4 digit input for the offset , as even the DICOM standard says "The optional suffix is not considered as a component." - so probably should not be shortened ?
Thanks for considering !
It is great to learn Rust from reading your code.

### DIFF
--- a/core/src/value/deserialize.rs
+++ b/core/src/value/deserialize.rs
@@ -474,5 +474,8 @@ mod tests {
         assert!(parse_datetime(b"20151130161445. +0000", default_offset).is_err());
         assert!(parse_datetime(b"20100423164000.001+3", default_offset).is_err());
         assert!(parse_datetime(b"200809112945*1000", default_offset).is_err());
+        assert!(parse_datetime(b"20171130101010.204+1", default_offset).is_err());
+        assert!(parse_datetime(b"20171130101010.204+01", default_offset).is_err());
+        assert!(parse_datetime(b"20171130101010.204+011", default_offset).is_err());
     }
 }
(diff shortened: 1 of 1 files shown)


Enet4 (commented):
Thank you for the PR, @jmlaka. You're right, I must have over-complicated the function under the assumption that some device could be providing partial timezone components. Non-standard date values are not unheard of, but this can be rewritten should we ever find edge cases worth supporting. This does not support ACR-NEMA dates anyway.

Please consider only including a few more test cases in the `tests` module in the same file, for future-proofing.

```rust
assert!(parse_datetime(b"20171130101010.204+1", default_offset).is_err());
assert!(parse_datetime(b"20171130101010.204+01", default_offset).is_err());
```

jmlaka:
Thank you for considering. Will do.

Juraj Mláka

Dňa ne 11. 7. 2021, 12:11 Eduardo Pinho ***@***.***>
napísal(a):

Enet4 (approved):
All right, things appear to be in good order! It is worth noting that there are plans to handle missing components in dates and times, since the standard explicitly mentions that they may be treated differently from "defaulted" components. This is tracked in #55. Should you ever be interested in contributing to DICOM-rs some more, this one could be an interesting task for you. 

Thanks again for working on this!

Issue #72, opened 2020-09-30
Retrieve date/time range from textual element value


Enet4:
In the context of queries, data elements with the value representations DA, TM, or DT can have two time instances separated by a dash ([standard](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html)). While no decoded value variant is available for ranges, they can be retrieved from string variants.

I picture being able to do the following:

```rust
let val = dicom_value!(Str, "20200313-20200930");
let (start, end) = val.date_range()?;
assert_eq!(start, NaiveDate::from_ymd(2020, 3, 13));
assert_eq!(end, NaiveDate::from_ymd(2020, 9, 30));
```

PR #145, opened 2021-07-23
Range parsing for da, tm, dt

jmlaka:
Hello Eduardo,
please review these methods for parsing date ranges.
They all return a tuple of 2 Options, where a None means no upper or lower range.
Hope they are Ok
Thanks

### DIFF
--- a/core/src/value/deserialize.rs
+++ b/core/src/value/deserialize.rs
@@ -29,6 +29,26 @@ pub enum Error {
     InvalidNumberToken { value: u8, backtrace: Backtrace },
     #[snafu(display("Invalid time zone sign token: got '{}', but must be '+' or '-'", *value as char))]
     InvalidTimeZoneSignToken { value: u8, backtrace: Backtrace },
+    #[snafu(display("No Range Separator Present"))]
+    NoRangeSeparator { backtrace: Backtrace },
+    #[snafu(display("End {} before Start {}", end, start))]
+    RangeInversion {
+        start: String,
+        end: String,
+        backtrace: Backtrace,
+    },
+    #[snafu(display("Start {} == End {}", start, end))]
+    RangeIsZero {
+        start: String,
+        end: String,
+        backtrace: Backtrace,
+    },
+    /*#[snafu(display("Could not parse DateTime range"))]
+    DTRangeParseError { backtrace: Backtrace },
+    #[snafu(display("Could not parse Time range"))]
+    TMRangeParseError { backtrace: Backtrace },
+    #[snafu(display("Could not parse Date range"))]
+    DARangeParseError { backtrace: Backtrace },*/
 }
 
 type Result<T, E = Error> = std::result::Result<T, E>;
@@ -46,31 +66,31 @@ pub fn parse_date(buf: &[u8]) -> Result<(NaiveDate, &[u8])> {
         }
         6 => {
             let year = read_number(&buf[0..4])?;
-            let month : u32 = read_number(&buf[4..6])?;
+            let month: u32 = read_number(&buf[4..6])?;
             let date: Result<_> =
                 NaiveDate::from_ymd_opt(year, month, 1).context(InvalidDateTimeZone);
             Ok((date?, &buf[6..]))
-        },
+        }
         len => {
             debug_assert!(len >= 8);
             let year = read_number(&buf[0..4])?;
-            let (month, day, rest)  = match buf[4] {
-                /* MM and DD not present, UTC follows*/ 
+            let (month, day, rest) = match buf[4] {
+                /* MM and DD not present, UTC follows*/
                 b'-' | b'+' => (1, 1, &buf[4..]),
                 _ => {
                     /* Attempt to parse MM */
-                    let m : u32 = read_number(&buf[4..6])?;
-                    let (d ,r) = match buf[6] {
+                    let m: u32 = read_number(&buf[4..6])?;
+                    let (d, r) = match buf[6] {
                         /* DD not present, UTC follows */
-                        b'-' | b'+' => (1, &buf[6..]) ,
-                        _ => (read_number(&buf[6..8])?, &buf[8..] ) /* Attempt to parse DD */
+                        b'-' | b'+' => (1, &buf[6..]),
+                        _ => (read_number(&buf[6..8])?, &buf[8..]), /* Attempt to parse DD */
                     };
-                    (m,d,r)
+                    (m, d, r)
                 }
             };
 
-            let date: Result<_> = NaiveDate::from_ymd_opt(year, month, day)
-                .context(InvalidDateTimeZone);
+            let date: Result<_> =
+                NaiveDate::from_ymd_opt(year, month, day).context(InvalidDateTimeZone);
             Ok((date?, rest))
         }
     }
@@ -150,14 +170,11 @@ fn parse_time_impl(buf: &[u8], for_datetime: bool) -> Result<(NaiveTime, &[u8])>
                         acc += 1;
                     }
                     (fract, &buf[n..])
-                },
+                }
                 /* no fraction, but UTC offset present, ok only for DT */
-                b'+' | b'-' if for_datetime => {
-                    (0, &buf[6..])
-                },
-                c => return UnexpectedAfterDateToken { value: c }.fail()
+                b'+' | b'-' if for_datetime => (0, &buf[6..]),
+                c => return UnexpectedAfterDateToken { value: c }.fail(),
             };
-
 
             let time =
                 naive_time_from_components(hour as u32, minute as u32, second as u32, fract)?;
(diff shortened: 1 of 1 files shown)


core/src/value/deserialize.rs
```diff
@@ -291,9 +301,146 @@ pub fn parse_datetime(buf: &[u8], dt_utc_offset: FixedOffset) -> Result<DateTime
...
+
+/**
+ *  Looks for a range separator '-'.
+ *  Returns a tuple of two Option\<DateTime\>
+ *  None means no upper or lower range is present
+ */
+pub fn parse_datetime_range(
+    buf: &[u8],
+    dt_utc_offset: FixedOffset,
+) -> Result<(Option<DateTime<FixedOffset>>, Option<DateTime<FixedOffset>>)> {
```
Enet4, line 386: Clippy is suggesting to move this one into a type alias, since it's a pretty long definition.

```rust
type DateTimeRange =  (Option<DateTime<FixedOffset>>, Option<DateTime<FixedOffset>>);
```

And I suppose we could do the same for a date range and a time range.

core/src/value/deserialize.rs
```diff
@@ -291,9 +301,146 @@ pub fn parse_datetime(buf: &[u8], dt_utc_offset: FixedOffset) -> Result<DateTime
...
+                        /* empty separator in the beginning */
+                        0 => true,
+                        /* empty separator at the end */
+                        x if *x == buf.len() - 1 => true,
+                        x if *x < buf.len() - 6 => {
+                            match buf[x + 5] {
+                                // separator present in 5 bytes, so assume this is an offset sign
+                                b'-' => false,
+                                _ => true,
+                            }
```
Enet4, line 408: This whole piece could turn into a single comparison, right?
```suggestion
                            // separator present in 5 bytes, so assume this is an offset sign
                            buf[x + 5] != b'-'
```

core/src/value/deserialize.rs
```diff
@@ -291,9 +301,146 @@ pub fn parse_datetime(buf: &[u8], dt_utc_offset: FixedOffset) -> Result<DateTime
...
+        .enumerate()
+        .find(|(i, c)| {
+            match **c == b'-' {
+                true => {
+                    match i {
+                        /* empty separator in the beginning */
+                        0 => true,
+                        /* empty separator at the end */
+                        x if *x == buf.len() - 1 => true,
+                        x if *x < buf.len() - 6 => {
```
Enet4, line 403: This code can panic if `buf.len() == 5` and the previous component is shorter than 4. (Although the date itself is invalid, the code should not panic from it.

Test case:

```rust
        assert!(parse_datetime_range("999-X".as_bytes(), o).is_err());
```

See if this change fixes the issue.

```suggestion
                        x if *x + 6 < buf.len() => {
```

core/src/value/deserialize.rs
```diff
@@ -540,4 +687,143 @@ mod tests {
         assert!(parse_datetime(b"20171130101010.204+01", default_offset).is_err());
         assert!(parse_datetime(b"20171130101010.204+011", default_offset).is_err());
     }
+    #[test]
+    fn test_parse_date_range() {
```
Enet4, line 691: Only checking for whether it parses OK gives very little confidence. Please make more specific assertions over the outcome of the operation. Where relevant, this can also be done for errors using `assert!` and `matches!` (as they are also part of the API).

Example:

```rust
assert_eq!(
    parse_date_range("1914-".as_bytes()).unwrap(),
    (NaiveDate::ymd(1914, 1, 1), None),
);
```

Enet4 (changes requested):
Thank you for working on this @jmlaka. I understand that this would be part 1 of bringing ranges (#72). Once this one is in, the next step would be to construct a high level API on top of the DICOM value interface. One thing at a time, though. :)

For now, please attend to the comments and suggestions below.

jmlaka:
Thank you for your review. Tried to correct and learned about error matching in tests.

Enet4:
Hey @jmlaka, the code formatting resulted in a conflict, but it should be easy to resolve.

core/src/value/deserialize.rs
```diff
@@ -29,10 +31,28 @@ pub enum Error {
     InvalidNumberToken { value: u8, backtrace: Backtrace },
     #[snafu(display("Invalid time zone sign token: got '{}', but must be '+' or '-'", *value as char))]
     InvalidTimeZoneSignToken { value: u8, backtrace: Backtrace },
+    #[snafu(display("No Range Separator Present"))]
```
Enet4, line 34:
```suggestion
    #[snafu(display("No range separator present"))]
```

core/src/value/deserialize.rs
```diff
@@ -29,10 +31,28 @@ pub enum Error {
...
+    NoRangeSeparator { backtrace: Backtrace },
+    #[snafu(display("End {} before Start {}", end, start))]
```
Enet4, line 36:
```suggestion
    #[snafu(display("End {} is before start {}", end, start))]
```

core/src/value/deserialize.rs
```diff
@@ -291,9 +305,140 @@ pub fn parse_datetime(buf: &[u8], dt_utc_offset: FixedOffset) -> Result<DateTime
         .context(InvalidDateTimeZone)
 }
 
+macro_rules! check_range {
+    ($s: expr, $e: expr) => {
+        if $s == $e {
+            RangeIsZero {
```
Enet4, line 311: According to [PS3.4 C.2.2.2.5](http://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_C.2.2.2.5.html), equal ends are OK, and the match is inclusive. For instance, a range such as `20210730-20210730` would simply match one date.

The error variant `RangeIsZero` should then be removed, and the tests be updated accordingly.

Enet4 (changes requested):
Aside from the conflict, I just found a few more matters worth looking into.

jmlaka:
Hello,

1. In case of two equal DT,TM,DA input values,  we shall return two equal
Ok values then ?

2. I noticed the conflict after committing, how can I avoid it in the
future?

Thanks

Juraj Mláka

Dňa pi 30. 7. 2021, 23:12 Eduardo Pinho ***@***.***>
napísal(a):

Enet4:
> Hello, 1. In case of two equal DT,TM,DA input values, we shall return two equal Ok values then ? 2. I noticed the conflict after committing, how can I avoid it in the future?

On 1, we can go with a `Result<(Option<Date>, Option<Date>)`.

Regarding 2, this happened because the branch to be merged is based on an outdated revision. You can either rebase or merge the branch against `origin/master`; and then update the branch by pushing it again. GitHub also provides a web interface for resolving conflicts, as an alternative.

jmlaka:
I managed to rebase, corrected the code.
I wanted to "squash" the commits to one, but still have to learn that one :)

jmlaka:
Corrected as you suggested.
Does "squashing" mean that you, when accepting a pull request, can make all my small commits into one ?

Enet4:
> Does "squashing" mean that you, when accepting a pull request, can make all my small commits into one ?

That is a possible option when merging, yes, but it is one that I generally to not feel the need to do.

For what it's worth, squashing commits together leads to some information loss (of the sequence of operations done by the author over time), with the main benefits of discarding mistakes in this sequence (which are not very useful for the revision process) and simplifying the revision history once merged.

For these reasons, this project currently does not have a strict commit squashing policy.

Enet4 (changes requested):
Thank you very much for the initiative of extending this PR to the value API. I only have one minor technical recommendation (see comment), but I would like to bring a major concern to the table regarding this feature.

Suppose the following date range: `2020-2021`

With the current implementation of `to_date_range`, both components are given default values for the unspecified components (month becomes January, and day becomes 1), so this range is deserialized into `(Some(2020-01-01), Some(2021-01-01))`. The standard states (and what users would naturally expect) that this range should cover all dates in the year 2021, but with what we have right now, all dates from 2021-01-02 to 2021-12-31 would be outside the range.

This means that an accurate representation of ranges will need to consider the presence of null components and actually encode them in the various core types `Date`, `Time`, `DateTime`, and respective range types (#55). Either this or use different defaults for the second member of the range, but it would not be consistent with how two elements are converted into a date, and I suspect that this would not be very reliable anyway.

I picture that a better strategy for handling ranges is to work on #55 first, and then come back to this to introduce accurate and useful implementations of date-time range matching. What do you think?

jmlaka:
Hello,

Thanks for looking at the pull request. I'm happy to learn from a
professional and maybe one day something useful will pop up from my code 😁

You are absolutely right.

I've been only using PyDicom before I stumbled upon your library. And
Pydicom has no such thing as range parsing. Which is kinda bad if you need
to handle c-find requests...

Let's see if I figure something out.

Regards

Juraj Mláka

Dňa so 14. 8. 2021, 18:20 Eduardo Pinho ***@***.***>
napísal(a):

core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "dicom-core"
-version = "0.4.0"
+version = "0.4.1"
```
Stargateur, line 3: I advice to let the owner or member of the crate handle versioning, this is more simple for the owner, since cargo package are not really link to git, the owner can merge several pull request and make the bump when wanted.

Enet4: That would be right, but in this case the upstream branch already has dicom-core at "0.4.1". This is probably some artifact from merging.

PR #153, opened 2021-08-14
Decode encapsulated pixel data frame offset table

Enet4:
- [core] modify Value prototype so that `offset_table` is a sequence of `u32`, as specified in the standard
- [encoding] add methods for encoding offset tables
- [parser] add methods for decoding and encoding offset tables
- [parser] raise error on undefined pixel fragment item instead of panicking
- [parser] add OffsetTable variant to data token
   - note that this does not apply to the lazy data set token because here the table is not read eagerly
- [parser] change dataset reader and writer so that they assume an offset table token on first item value instead of `ItemValue`
- [object] change in-memory DICOM object constructor to assume offset table data token
- [object] check offset table in integration test
- [dcmdump] update tool to present human readable offset table

### DIFF
--- a/encoding/src/encode/mod.rs
+++ b/encoding/src/encode/mod.rs
@@ -85,9 +85,14 @@ pub enum Error {
         backtrace: Backtrace,
         source: io::Error,
     },
+    #[snafu(display("Failed to write pixel data offset table"))]
+    WriteOffsetTable {
+        backtrace: Backtrace,
+        source: io::Error,
+    },
 }
 
-pub type Result<T> = std::result::Result<T, Error>;
+pub type Result<T, E = Error> = std::result::Result<T, E>;
 
 /// Type trait for an encoder of basic data properties.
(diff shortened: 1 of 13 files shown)


encoding/src/encode/mod.rs
```diff
@@ -323,6 +328,15 @@ pub trait Encode {
     fn encode_primitive<W>(&self, to: W, value: &PrimitiveValue) -> Result<usize>
     where
         W: Write;
+    
+    /// Encode and write a DICOM pixel data offset table
+    /// to the given destination.
+    ///
+    // Note that offset tables might not apply to all forms of encoding,
+    // but this method needs to exist nevertheless.
```
Enet4, line 336:
```suggestion
    /// Encode and write a DICOM pixel data offset table
    /// to the given destination.
    ///
    /// For convenience, returns the number of bytes written on success,
    /// equivalent to `offset_table.len() * 4`.
    // Note that offset tables might not apply to all forms of encoding,
    // but this method needs to exist nevertheless.
```

Issue #50, opened 2020-06-07
Best-effort resolution of transfer syntax of DICOM files without file meta group


Enet4:
The current implementation does not allow for the reading of DICOM objects without a file meta group. Although to the best of my knowledge the lack of this group is non-standard, it is known that some implementations are capable of resolving this anyway.

- Make file opening more flexible by continuing the process even without finding enough elements for a compliant file meta group table, by assuming a flexible transfer syntax.
   - As a best effort, this is something that can be reiterated as we stumble upon more files to test with, but it should go along something like trying _Explicit VR Little Endian_ first, then _Implicit VR Little Endian_ if the outcome of the former did not make sense.
- Leave a warning (see #49) in the case that the file meta group is missing.

One could start by writing tests for the test files in `dicom-test-files` without a file meta group (see example in #48), then working your way into making them pass.

bastien-solutions:
Hello,

On trying to open a file with this transfert syntax : 1.2.840.10008.1.2.1 

```rust
    let obj = open_file(&in_file)?;
```

I get : `[ERROR] Unsupported transfer syntax `1.2.840.10008.1.2.1 ``

I assume this is linked to this issue ?

bastien-solutions:
I think you got it, with dcmtk dcmdump : `W: DcmUniqueIdentifier: Element TransferSyntaxUID (0002,0010) contains one or more space characters, which were removed`.

May be we can trim this particular tag value ?

Enet4:
There are two key components in this particular interaction:

- The standard transfer syntax registry, which is prepared to receive at most a single null character at the end of the UID string, and nothing else.
- The file meta table builder, which retains the values as they were encoded. Changing this may be possible, but the consequences may need to be further evaluated.

I have filed issue #148 to host this new matter. This particular issue is for handling files which do not have a file meta group at all.

bastien-solutions:
Agree for a : 
```rust
        if uid.as_bytes().last().cloned() == Some(b'\0') || uid.ends_with(' ') {
            &uid[..uid.len() - 1]
        } else {
            &uid
        }
```
in lib.rs:68 ?

If yes I can make a PR.

Enet4:
@bastien-solutions I'd probably do it like this:

```rust
uid.trim_end_matches(|c| c.is_whitespace() || c == '\0')
```

But certainly, a pull request would be much appreciated!

bastien-solutions:
No problem, can you give me the authorization to create / push on branch ?

[...]

PR #157, opened 2021-08-23
Make transfer syntax registry more flexible against UIDs with null characters

### LINKED ISSUE #148: Cannot read transfer syntax UIDs with non-standard whitespace characters
Currently, transfer syntax indices are required to be robust to the presence of a single null character at the end of a transfer syntax UID string, which is compliant with standard DICOM. But some DICOM implementations in the wild may put other whitespace characters at the end, which is not aligned with these expectations. While UID's are not supposed to have those whitespace, DICOM-rs could be a bit more resilient.

Originally mentioned in https://github.com/Enet4/dicom-rs/issues/50#issuecomment-889124095.

I am picturing two ways of making this work, not necessarily exclusive, but each would be sufficient:

### Approach 1: make the standard transfer syntax registry more flexible

Instead of just failing to find the transfer syntax, a more flexible implementation of [`TransferSyntaxRegistry::get`](https://github.com/Enet4/dicom-rs/blob/master/transfer-syntax-registry/src/lib.rs#L65) would:

- try to trim any kind of whitespace at the end of the string;
- and emit a warning if it finds a whitespace character other than `\0` (#49 is pending though; leave a TODO comment for now).

This is the least invasive approach and would work regardless of where the UID came from.

### Approach 2 (:warning: on hold for now): automatically trim the transfer syntax UID on file meta table building

Once the transfer syntax UID is found in the file meta group, trim any kind of whitespace before saving it, and emit a warning if it contained any non-standard DICOM whitespace (leave a TODO comment for #49).
This is probably what DCMTK does, according to https://github.com/Enet4/dicom-rs/issues/50#issuecomment-889186415, and stops the original mistake from propagating to subsequently written files, but it does not handle poor transfer syntax UID strings coming from elsewhere (such as an association request PDU), and it changes the current expectations of the file meta table, since it currently does no trimming of any value.

This issue would be resolved with either one, although I am probably more inclined towards doing approach 1. Whichever decision is made here is also going to influence #70.

bastien-solutions:
Suppress trailing space and nil char in transfer syntax UID.

Resolves #148

### DIFF
--- a/transfer-syntax-registry/src/lib.rs
+++ b/transfer-syntax-registry/src/lib.rs
@@ -63,14 +63,9 @@ impl fmt::Debug for TransferSyntaxRegistryImpl {
 impl TransferSyntaxRegistryImpl {
     /// Obtain a DICOM codec by transfer syntax UID.
     fn get<U: AsRef<str>>(&self, uid: U) -> Option<&'static TransferSyntax> {
-        let ts_uid = {
-            let uid = uid.as_ref();
-            if uid.as_bytes().last().cloned() == Some(b'\0') || uid.ends_with(' ') {
-                &uid[..uid.len() - 1]
-            } else {
-                &uid
-            }
-        };
+        let ts_uid = uid
+            .as_ref()
+            .trim_end_matches(|c: char| c.is_whitespace() || c == '\0');
         self.m.get(ts_uid).copied()
     }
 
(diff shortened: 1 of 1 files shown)


Enet4:
I took the liberty of adjusting the title, as this pull request resolves issue #148 and not #50.

Other than that, this change seems correct and good to merge. Thank you!

PR #160, opened 2021-08-28
Null date components

jmlaka:
Hello @Enet4 ,
please take a look at this work in progress.
PartialDate is working (see tests)
Time and DateTime not done.
I'd like to know if I'm heading in the right direction.

Thanks


core/src/value/deserialize.rs
```diff
@@ -29,55 +50,286 @@ pub enum Error {
...
+impl Imprecision<NaiveDate> for PartialDate {
+    fn earliest(&self) -> Result<NaiveDate> {
+        let (y, m, d) = match self.components {
+            DatePrecision::Year(y) => (y, 1, 1),
+            DatePrecision::Month(y, m) => (y, m, 1),
+            DatePrecision::Day(y, m, d) => (y, m, d),
+        };
+        let date: Result<_> =
+            NaiveDate::from_ymd_opt(y as i32, m as u32, d as u32).context(InvalidDate);
+        Ok(date?)
```
Enet4, line 197: This should work. A similar pattern can be employed in the other methods.
```suggestion
        NaiveDate::from_ymd_opt(y as i32, m as u32, d as u32).context(InvalidDate)
```

core/src/value/deserialize.rs
```diff
@@ -8,6 +8,26 @@ use std::ops::{Add, Mul, Sub};
 pub enum Error {
     #[snafu(display("Unexpected end of element"))]
     UnexpectedEndOfElement { backtrace: Backtrace },
+    //TODO impl Display for DateComponent and Range or leave it like this ?
+    #[snafu(display(
+        "{:?} has invalid value:  {}. Must be in {:?}",
```
Enet4, line 13: Sticking to the use of Debug here is fine, since they already print useful and lean textual representations. Leaving just a minor suggestion for the error message. 
```suggestion
        "Invalid {:?} value {}, must be in {:?}",
```

core/src/value/deserialize.rs
```diff
@@ -29,55 +50,286 @@ pub enum Error {
...
+
+#[derive(Debug, PartialEq)]
+pub enum DatePrecision {
+    Year(u16),
+    Month(u16, u8),
+    Day(u16, u8, u8),
+}
+
+#[derive(Debug, PartialEq)]
+pub struct PartialDate {
```
Enet4, line 87: The creation of both `PartialDate` and a `DatePrecision` is a bit intriguing. Couldn't `DatePrecision` be renamed to `PartialDate` and fulfill this one's role all the same?

core/src/value/deserialize.rs
```diff
@@ -29,55 +50,286 @@ pub enum Error {
...
+        DateComponent::UTCOffset => 0..=86_399,
+    };
+
+    if range.contains(value) {
+        Ok(())
+    } else {
+        InvalidComponent {
+            component: component,
+            value: value.clone(),
+            range: range,
```
Enet4, line 242: Running `cargo clippy` every once in a while will give you tips to make the code tidier.

```suggestion
            component,
            value,
            range,
```

core/src/value/deserialize.rs
```diff
@@ -29,55 +50,286 @@ pub enum Error {
...
+    } else {
+        let year: u32 = read_number(&buf[0..4])?;
+        check_component(DateComponent::Year, &year)?;
+        let buf = &buf[4..];
+        if buf.len() < 2 {
+            Ok((
+                PartialDate {
+                    components: DatePrecision::Year(year as u16),
+                },
+                &buf,
```
Enet4, line 302:
```suggestion
                buf,
```

core/src/value/deserialize.rs
```diff
@@ -29,55 +50,286 @@ pub enum Error {
...
+            DatePrecision::Day(..) => DateComponent::Day,
+        }
+    }
+}
+
+/**
+ * This trait is implemented by partial precision
+ * Date, Time and DateTime structures.
+ */
+pub trait Imprecision<U>
```
Enet4, line 145: This seems to call for a super-trait, rather than a constraint `Self: Precision`
```suggestion
pub trait Imprecision<U>: Precision
```

Better clarify what `U` is in this context. There may also be better names for these traits... `AsTemporalRange<T>`, maybe?

core/src/value/deserialize.rs
```diff
@@ -29,55 +50,286 @@ pub enum Error {
...
+     */
+    fn latest(&self) -> Result<U>;
+
+    /**
+     * Returns a tuple of the earliest and latest possible value from a partial precision structure.
+     * 
+     */
+    fn to_range(&self) -> Result<(Option<U>, Option<U>)> {
+        if self.is_precise() {
+            ZeroRange.fail()
```
Enet4, line 169: This will depend a bit on how `is_precised` will be designed, but if the range extremities are inclusive, creating a range should succeed even if a single specific date-time is given. For `U = PartialDate`, for example, `2021-08-28` would return a range that matches the date itself.

Enet4 (commented):
Thank you @jmlaka, it looks promising overall. I mixed some questions and suggestions inline with the code.

Also, please consider moving the partial date-time component logic to a sibling module (e.g. `dicom_core::value::partial`), as `deserialize` is starting to get very big, and a good deal of the new logic is not exactly about deserialization at this point.

jmlaka:
Thanks @Enet4.

PR #161, opened 2021-08-31
[dcmdump] add option --color=(auto|always|never)

charbeljc:
Useful, so you can do, for instance

```bash
dcmdump --color=always your.dcm | less -r
```

### DIFF
--- a/dcmdump/src/main.rs
+++ b/dcmdump/src/main.rs
@@ -20,6 +20,7 @@ use dicom::transfer_syntax::TransferSyntaxRegistry;
 use snafu::ErrorCompat;
 use std::borrow::Cow;
 use std::fmt;
+use std::str::FromStr;
 use std::io::{stdout, ErrorKind, Result as IoResult, Write};
 use std::path::PathBuf;
 use structopt::StructOpt;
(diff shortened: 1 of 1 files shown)


dcmdump/src/main.rs
```diff
@@ -29,6 +30,44 @@ const ERROR_READ: i32 = -2;
...
+    Never,
+    Auto,
+    Always,
+}
+
+impl ToString for ColoringError {
+    fn to_string(&self) -> String {
+        "invalid color mode".into()
+    }
+}
```
Enet4, line 47: Error types should implement `Display`, which is also a requirement for implementing `Error`.
 
```suggestion
impl Display for ColoringError {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        f.write_str("invalid color mode")
    }
}

impl std::error::Error for Coloring {}
```

Enet4 (changes requested):
This is indeed useful, thank you!

Please have a look at the suggested changes below.

charbeljc:
> This is indeed useful, thank you!
> 
> Please have a look at the suggested changes below.

Hi @Enet4 I just did commit your suggestions. About `dcmdump`, I think we should make most of it become a library too, so we can reuse it in other contexts. Also, we should rename the command line  to something like `dicom-dump` or` dicom-dcmdump` to avoid name clash with the `dcmtk`version.

Enet4 (approved):
> Hi @Enet4 I just did commit your suggestions.

Great, I already tried it out and it's good to enter. Thanks!

> About `dcmdump`, I think we should make most of it become a library too, so we can reuse it in other contexts.

That could be done. This initiative could start by splitting the logic in main.rs to separate modules, so as to fully understand what should be in the library. It may also help track down certain constructs which should be available in the main library crates, should they be considered useful in most use cases.

> Also, we should rename the command line to something like `dicom-dump` or` dicom-dcmdump` to avoid name clash with the `dcmtk`version.

That is one thing that has been in my mind for a while. This tool might one day have the option to print DCMTK-compatible dumps, but it is still a program of its own and with its own CLI specification. I picked `dcmdump` because there are a few other projects which followed the same nomenclature (such as dcm4che), but in retrospective this collides with the binary if they are both in the system PATH, and this might not be very consistent with the rest of the DICOM-rs ecosystem.

Issue #164, opened 2021-09-03
[ul] PDataWriter sends corrupted data on large payloads


Enet4:
Due to an implementation detail of the P-Data writer's PDU dispatcher, it would write inconsistent data in the case of having a lot of data in buffer.

Related with #163 and #158.

Enet4:
Resolved by #164

PR #165, opened 2021-09-04
[ul] Solve bugs related with PDU sending and receiving

### LINKED ISSUE #163: Association API implementation does not honor remote AE's maximum PDU length
As identified in https://github.com/Enet4/dicom-rs/pull/158#discussion_r699315685, the association client is ignoring the connection acceptor's requested maximum PDU length, and as such may send PDUs which are too large for the destination to handle. This could also be happening at the server abstraction, but this is yet to be validated.

Enet4:
Originally intended to fix #163, but other issues were found #164 in `ul`, so they were all fixed here.

Summary:

- update association client and server to read the `MaxLengthReceived` user variable item, and honor that maximum length in subsequent interactions (#163)
- also update client and server to send more user variables, including implementation credentials
- Update reader module with better documentation and tweaks
   - i.e. `PDU_HEADER_SIZE` const value changed to 6, while documenting it that it refers to the length of the first 3 fields
- Fix `PDataWriter` for payloads which may cover more than 2 PDUs (#164).
 - Improve `PDataWriter` implementation so that it does less byte copying and sends each full PDU in a single `write_all` call
- [storescu] revert to using PDataWriter to send the object data

CC @charbeljc If you can, your validation of these changes would be greatly appreciated.


charbeljc:
> [...]
> 
> CC @charbeljc If you can, your validation of these changes would be greatly appreciated.

Hi @Enet4 It was quite a busy day at work, but I will try to run some tests, hopefully tomorrow !

PR #173, opened 2021-09-09
Dcmdump as a library

charbeljc:
* rename command line to `dicom-dump`
* extract library


dcmdump/src/dump.rs
```diff
@@ -70,6 +70,23 @@ pub fn dump_file(obj: DefaultDicomObject, width: u32, no_text_limit: bool) -> Io
     Ok(())
 }
 
+pub fn dump_object(obj: &InMemDicomObject<StandardDataDictionary>, no_text_limit: bool) -> IoResult<()> {
```
Enet4, line 73: I think that all `pub` functions should let the user could dump something into an arbitrary writer.

```rust
pub fn dump_object(to: impl Write, obj: &InMemDicomObject<StandardDataDictionary>, no_text_limit: bool) -> IoResult<()> {
```

This should not be too far from adding this `to` parameter adn replacing `println!` with `writeln!`.

dcmdump/src/dump.rs
```diff
@@ -0,0 +1,514 @@
...
+            "Private Information".bold(),
+            format_value_list(v.iter().map(|n| format!("{:02X}", n)), width, false)
+        )?;
+    }
+
+    writeln!(to)?;
+    Ok(())
+}
+
+pub fn dump<W, D>(
```
Enet4, line 173: Oh, and now that they are part of a library, some documentation would be nice too.

Enet4 (commented):
Hey @charbeljc, thank you for the PR!

Not sure if this is complete, but I'm leaving some feedback for now.

The part about the rename is something that can be done here or later. There will still have to be one final release of the `dcmdump` crate so as to deprecate it and direct users to the new one.

Enet4:
A summary of changes made:

- Made the `dicom-dump` crate an independent copy from `dcmdump`
- Marked `dcmdump` as deprecated
- Added `DumpOptions` builder for flags and other options to dumping
- Added a `_to` variant to all functions that print the output to an arbitrary writer
- Several tweaks to the readme files and extended documentation

Enet4 (approved):
This seems to be in good order now. Thank you for the initiative.

PR #174, opened 2021-09-09
Preliminary dicom web client

charbeljc:
This is just a POC for now. I tested it with Orthanc only
This PR was based on #173 so the revelant commits are also included.
This is rough and dirty, as I'm neither a dicom or rust expert ...

Regards


vsaase:
nice, this works for me with Orthanc
thanks for starting the JSON to Dicom decoder, I will see if I can add to this
I wonder why you are doing this JSON array parsing. I did not run into a problem with serde_json with array responses.

you should also check it with DCM4CHEE, a popular open source PACS
It can easily be started with docker_compose: https://github.com/dcm4che/dcm4chee-arc-light/wiki/Run-minimum-set-of-archive-services-on-a-single-host#use-docker-compose

Then upload some dicom files with e.g. DCMTK: `dcmsend localhost 11112 -aec DCM4CHEE +r +sd <dicom directory>` 
The dicomweb endpoint is `http://localhost:8080/dcm4chee-arc/aets/DCM4CEE/rs`

You will see that DCM4CHEE does not by default respond with the content-length in the multipart body on requesting a file with WADO. I think this is in violation of the standard [here](http://dicom.nema.org/medical/dicom/current/output/html/part18.html), but would still need a workaround since DCM4CHEE is so popular.

charbeljc:
> nice, this works for me with Orthanc thanks for starting the JSON to Dicom decoder, I will see if I can add to this
> 
> I wonder why you are doing this JSON array parsing. I did not run into a problem with serde_json with array responses. You can parse the array as a vector of Values like this: `let parsed: Vec<Value> = serde_json::from_str(jsonstring)?;`
> 
I was doing this to achieve a streaming API on json reading, to save memory and latency, but this is certainly a premature optimisation.

> you should also check it with DCM4CHEE, a popular open source PACS It can easily be started with docker_compose: https://github.com/dcm4che/dcm4chee-arc-light/wiki/Run-minimum-set-of-archive-services-on-a-single-host#use-docker-compose

> 
> Then upload some dicom files with e.g. DCMTK: `dcmsend localhost 11112 -aec DCM4CHEE +r +sd <dicom directory>` The dicomweb endpoint is `http://localhost:8080/dcm4chee-arc/aets/DCM4CEE/rs`
> 
> You will see that DCM4CHEE does not by default respond with the content-length in the multipart body on requesting a file with WADO. I think this is in violation of the standard [here](http://dicom.nema.org/medical/dicom/current/output/html/part18.html), but would still need a workaround since DCM4CHEE is so popular.

Thanks for the pointer, I will hopefully check this  when thinks settle down a bit at Qynapse. In the mean time, feel free to takeover this POC !

PR #178, opened 2021-09-21
- typo corrected

jmlaka:


Enet4 (approved):
That is correct, thank you.

I might still document these types a bit better in the future, so that this method is no longer necessary.

PR #185, opened 2021-10-02
Null components2 no precision tweaks for time

jmlaka:
This one removes some inconsistency in DicomTime - only  full 6 digits in second fraction is considered precise now. 
Added as_naive_date,time, datetime for partial values.


Enet4:
Hey @jmlaka, thank you for the pull requests! This caught me a bit off guard here: which pull request should be looked into first? Are they complementary?

jmlaka:
Sorry for this.
I think this one is an upgrade to the previous.
So please look into this one.

If you approve it in general, I will add the range parsing in this PR as
well.



Juraj Mláka

Dňa so 2. 10. 2021, 16:09 Eduardo Pinho ***@***.***>
napísal(a):

core/src/value/partial.rs
```diff
@@ -0,0 +1,873 @@
...
+        fraction,
+        precision
+    ))]
+    FractionPrecisionMismatch {
+        fraction: u32,
+        precision: u32,
+        backtrace: Backtrace,
+    },
+    #[snafu(display("Conversion Failed."))]
+    Conversion { source: std::num::TryFromIntError },
```
Enet4, line 51: Could we expand this to also include which date/time component failed to convert?

core/src/value/partial.rs
```diff
@@ -0,0 +1,873 @@
...
+    #[snafu(display(
+        "Cannot convert from an imprecise value. This value represents a date / time range."
+    ))]
+    ImpreciseValue { backtrace: Backtrace },
+}
+
+type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// Represents components of Date, Time and DateTime values.
+#[derive(Debug, PartialEq)]
```
Enet4, line 61: Can be also `Copy`, `Clone`, `Eq`, `Hash`, `PartialOrd` and `Ord`.

core/src/value/partial.rs
```diff
@@ -0,0 +1,873 @@
...
+pub enum DateComponent {
+    Year,
+    Month,
+    Day,
+    Hour,
+    Minute,
+    Second,
+    Fraction,
+    UTCOffset,
```
Enet4, line 70: Acronyms constitute a term, so it is subjected to upper camel case, according to [C-CASE](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case).
```suggestion
    UtcOffset,
```

core/src/value/partial.rs
```diff
@@ -0,0 +1,873 @@
...
+}
+
+/// Represents a Dicom Time value with a partial precision,
+/// where some time components may be missing.
+///
+/// Unlike RUSTs `chrono::NaiveTime`, this implemenation of time is DICOM compliant:  
+/// - has only 6 digit precision for fraction of a second
+/// - has no means to handle leap seconds
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum DicomTime {
```
Enet4, line 95: `DicomTime` and `DicomDate` have publicly visible variants, which can let users create inconsistent values. I would strongly recommend employing a newtype to hide the enum-based implementation.

```rust
#[derive(Debug, Clone, Copy, PartialEq)]
pub struct DicomTime(DicomTimeImpl);

#[derive(Debug, Clone, Copy, PartialEq)]
enum DicomTimeImpl {
    // ...
}
```

core/src/value/partial.rs
```diff
@@ -0,0 +1,873 @@
...
+    }
+}
+
+impl TryFrom<&NaiveDate> for DicomDate {
+    type Error = Error;
+    fn try_from(date: &NaiveDate) -> Result<Self> {
+        let year: u16 = date.year().try_into().context(Conversion)?;
+        let month: u8 = date.month().try_into().context(Conversion)?;
+        let day: u8 = date.day().try_into().context(Conversion)?;
+        Ok(DicomDate::from_ymd(year, month, day)?)
```
Enet4, line 178:
```suggestion
        DicomDate::from_ymd(year, month, day)
```

core/src/value/partial.rs
```diff
@@ -0,0 +1,873 @@
...
+        check_component(DateComponent::Minute, &minute)?;
+        check_component(DateComponent::Second, &second)?;
+        Ok(DicomTime::Second(hour, minute, second))
+    }
+
+    /**
+     * Constructs a new `DicomTime` with hour, minute, second and second fraction
+     * precision (HHMMSS.FFFFFF).
+     */
+    pub fn from_hmsf(
```
Enet4, line 227: This one feels a bit hard to use, since the given `frac_precision` could be inconsistent with `fraction`. I would suggest separating this it to one that takes milliseconds and another one that takes microseconds. If this function is to stay, it should be carefully documented to better explain how to use `frac_precision`.

core/src/value/partial.rs
```diff
@@ -0,0 +1,873 @@
...
+            .and_time(time)
+            .context(InvalidDateTime)
+    }
+}
+
+impl DicomDate {
+    /**
+     * Retrieves a `chrono::NaiveDate` if value is precise.
+     */
+    pub fn as_naive_date(self) -> Result<NaiveDate> {
```
Enet4, line 543: Revisiting [C-CONV](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv), `to_naive_date` would be more conventional (it's owned -> owned for `Copy` types). Same for `to_naive_time` and `to_chrono_datetime`.

core/src/value/primitive.rs
```diff
@@ -1974,7 +2000,174 @@ impl PrimitiveValue {
...
+    /// # Example
+    ///
+    /// ```
+    /// # use dicom_core::value::{C, PrimitiveValue};
+    /// # use smallvec::smallvec;
+    /// # use chrono::NaiveDate;
+    ///  use dicom_core::value::{AsRange, DicomDate};
+    ///
+    ///  let value = PrimitiveValue::Str("200002".into());
+    ///  let dicom_date = value.to_dicom_date().unwrap();
```
Enet4, line 2028: Examples should avoid using `unwrap`. See [C-QUESTION-MARK](https://rust-lang.github.io/api-guidelines/documentation.html?highlight=error#examples-use--not-try-not-unwrap-c-question-mark) for how to improve this.

core/src/value/primitive.rs
```diff
@@ -1974,7 +2000,174 @@ impl PrimitiveValue {
...
+    ///
+    ///  assert_eq!(
```
Enet4, line 2030: Since this is documentation, we can guide the user within the example with comments.
```suggestion
    ///  // it is not precise, day of the month is unspecified
    ///  assert_eq!(
```

Enet4 (changes requested):
Thank you very much for continuing with this, @jmlaka. This implementation seems sound, the tests provide decent coverage, and overall I believe that this would make a fine addition to the core crate.

Please attend to the comments below to improve these changes. After attending to them, I believe we can "undraft" the PR. As for range implementations and whatnot, feel free to extend these concepts in a separate PR if you find that more convenient.

core/src/value/partial.rs
```diff
@@ -0,0 +1,873 @@
...
+    Month(u16, u8),
+    Day(u16, u8, u8),
+}
+
+/// Represents a Dicom Time value with a partial precision,
+/// where some time components may be missing.
+///
+/// Unlike RUSTs `chrono::NaiveTime`, this implemenation of time is DICOM compliant:  
+/// - has only 6 digit precision for fraction of a second
+/// - has no means to handle leap seconds
```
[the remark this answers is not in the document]
jmlaka: That's not misleading, an outright error on my side.
Removed the leap second comment, adjusted range checking to include 60.

core/src/value/partial.rs
```diff
@@ -0,0 +1,873 @@
...
+pub enum DicomTime {
+    Hour(u8),
+    Minute(u8, u8),
+    Second(u8, u8, u8),
+    Fraction(u8, u8, u8, u32, u8),
+}
+
+/// Represents a Dicom DateTime value with a partial precision,
+/// where some date / time components may be missing.
+#[derive(Debug, PartialEq, Clone, Copy)] // PartialOrd ?
```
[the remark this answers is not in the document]
jmlaka: It is a lost comment.
I'd leave implementing PartialOrd for the future.
Can we mark this a todo somehow, please?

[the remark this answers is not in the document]
Enet4: Sure. You can file an issue for this and use the issue number in the code like this:

```rust
// TODO(#...) impl PartialOrd
#[derive(Debug, PartialEq, Clone, Copy)]
```

core/src/value/partial.rs
```diff
@@ -0,0 +1,873 @@
...
+        Ok(DicomDate::from_ymd(year, month, day)?)
+    }
+}
+
+impl fmt::Display for DicomDate {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            DicomDate::Year(y) => write!(f, "{:04}-MM-DD", y),
+            DicomDate::Month(y, m) => write!(f, "{:04}-{:02}-DD", y, m),
+            DicomDate::Day(y, m, d) => write!(f, "{:04}-{:02}-{:02}", y, m, d),
```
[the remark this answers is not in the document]
Enet4: I'm thinking more of the API for `DicomDate` and the others. Imagining the case where we're building a DICOM object from scratch, we need some way of converting them into encoded strings. An additional `&self`-receiving method is what comes to mind.

```rust
let study_date = PrimitiveValue::from(date.to_encoded());
```

(Although in this case there could also be a `From<DicomDate>` implementation for `PrimitiveValue`!)

jmlaka:
Hello @Enet4,

I think I was able to fix the issues you pointed out - learned some new things again, was fun.
This PR includes range parsing as well. 
It is, for the most part, isolated in the range.rs file, so will not be overly time-consuming to read the differences.

We can replace the internal representation of PrimitiveValue::Date... for these new structs as a next step, I think.

Thanks

jmlaka:
`parse_datetime_range()` is currently ignoring UTC offsets, because there is one borderline situation, where the parser cannot decide between a date and a UTC. But I decided to change this in the future and document the borderline situation.

jmlaka:
Perfect. Thank you

PR #192, opened 2021-10-18
- typo prevented build

jmlaka:


jmlaka:
Aha !

I was still using 1.50. I should have thought about that since it passed GitHub CI tests

Enet4:
All right, thank you for reporting! The project currently does not have an MSVR policy (just filed #193 for this), but having a wider range of versions that still compile with a change as simple and non-compromising as this one is fair.

jmlaka:
Need help here, I just ran clippy with the new 1.55 version and it suggested deleting '&' on myriad occasions ... 
(needles borrow)
Is there a way to run clippy and silent these warnings ?
Or shall I just obey clippy, as nobody will use the older version of Rust anyway ?

jmlaka:
Thank you,

So I will ignore these and revert back a few changes where I already
removed the &.

Juraj Mláka

Dňa po 18. 10. 2021, 17:01 Eduardo Pinho ***@***.***>
napísal(a):

PR #196, opened 2021-10-20
Remove chrono

jmlaka:
Hi @Enet4,
I changed the PrimitiveValues internals as suggested.

Please be so nice and resolve the problem of the commit being behind the master.
I'm afraid I would mess it up.
Thank you

### DIFF
--- a/core/src/header.rs
+++ b/core/src/header.rs
@@ -5,7 +5,7 @@
 use crate::value::{
     CastValueError, ConvertValueError, DicomDate, DicomDateTime, DicomTime, PrimitiveValue, Value,
 };
-use chrono::{DateTime, FixedOffset, NaiveDate, NaiveTime};
+use chrono::{DateTime, FixedOffset, NaiveTime};
 use num_traits::NumCast;
 use snafu::{Backtrace, Snafu};
 use std::borrow::Cow;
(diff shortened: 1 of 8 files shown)


Enet4 (commented):
I've been investigating how this integrates with the `dump` crate and noticed that there is currently a discrepancy in API between the `DicomValue` and the `PrimitiveValue`. `to_date` in one is `to_dicom_date` in another, for example. The `to_datetime` one is particularly confusing, because it returns a DICOM datetime in one and a chrono DateTime in another. Any idea on how to make this uniform? Ideally, we should try to give both types the same method names.

> I left fmt::Debug as is, with all the funny placeholders, don't know if you're OK with it.

That is OK. I also checked the DICOM time `Display` impl and it seems to function as intended. Once we resolve the issue above, I would be confident to bring it upstream and start preparing a new release.

jmlaka:
Yes, I see I forgot to solve the naming.

I renamed the method names returning the new `Dicom...` structures to the short ones.
Methods returning `chrono::` values contain 'naive' or 'chrono' in them.

Enet4 (approved):
Just added a fix to the documentation and now it's good to merge! Thank you for this awesome contribution, @jmlaka!

jmlaka:
@Enet4 it was really fun as I learned new things in the process.
Keep up the good work.
I hope that one day I will find this lib inside one of my devices :)

Issue #166, opened 2021-09-04
Trailing NULLs should be trimmed from UIs


troelsarvin:
I'm working with a set of million of DICOM files where I need to extract study instance UID from the fil and use the SIUID as strings  for subsequent work. For a surpringly large set of files, my subsequent use of the SIUID has been failing. It turns out the failing SIUIDs all have trailing NULL bytes. It's a nasty issue to debug, because when you print strings with NULL bytes to the console, you may not see the NULL bytes.

The DICOM standard has an (odd) requirment to pad unique identifiers (UIs) with NULLs, if they don't have an even length. As a natural consequence of this, I believe it's generally expected, that DICOM libraries trim trailing NULLs when returning UIs to applications.

I suggest dicom-rs trim trailing null bytes when returning UIs such as when using dicom_object::element_by_name("StudyInstanceUID").

Enet4:
It is true that trailing whitespaces are a common trend when working with DICOM data. The way that this is handled in the DICOM-rs library is to provide utility methods. This is already possible in v0.4, the code below should be enough for your use case:

```rust
use dicom::object::open_file;
use dicom::dictionary_std::tags;

let obj = open_file("CT0001.dcm")?
let stuid = obj.element(tags::STUDY_INSTANCE_UID)?.to_clean_str()?;
```

This could probably be handled in a different way, although I find the current approach reasonable: all values are fetched as they are, and cleaning them up is an explicit, easily available operation.

troelsarvin:
I believe the need to run to_clean_str() breaks the principle of least surprise: One wouldn't expect to have to deal with DICOM protocol details like padding when asking for the study instance UID, for example.
And I assert that other tools and libraries don't require you to handle trailing null chars when asking for UIs, but I cannot prove that assertion right now.

Also, I like the principleof the typical use being the simple use. If someone _really_ want to peek into the raw bytes of an UI, then that should be done with a special function.

troelsarvin:
I found out about dcmtk: In dcmtk,  one doesn't need to trim trailing null-chars when getting UIs.

michaelonken:
This is correct, and such a null char is added automatically when writing (if otherwise odd length UID), i.e. the user does not have to add it when inserting it into Study Instance UID (for example). 

I think there is no good reason why the user wants to know that there was a null byte in the end (or padding spaces for text VRs), unless you debug problems. For the latter case there should be low level methods or call parameters to hand over the padding to the user.

[...]

PR #205, opened 2021-12-17
[dump] Remove control code in text value

### LINKED ISSUE #187: [dump] Handle line breaks in text values
The `dcmdump` tool (recently renamed to `dicom-dump`) prints a part of the text value as it dumps the contents of each element. When it stumbles upon a  new line character, that creates an actual new line, which disrupts the dumping sequence a bit.

It would be nice that line break characters were escaped or replaced when dumping.

aaronzs:
Fix #187

Removes control code (Cc) in text value e.g. NUL, LF, CR and more.

### DIFF
--- a/dump/src/lib.rs
+++ b/dump/src/lib.rs
@@ -737,7 +737,7 @@ where
     let len = values.len();
     let mut pieces = values.take(64).map(|piece| {
         let mut piece = piece.to_string();
-        piece.retain(|c| !c.is_control());
+        piece = piece.replace(|c: char| c.is_control(), "�");
         if quoted {
             piece = piece.replace("\"", "\\\"");
             piece.insert(0, '"');
--- a/dump/src/lib.rs
+++ b/dump/src/lib.rs
@@ -736,18 +736,14 @@ where
     let values = values.into_iter();
     let len = values.len();
     let mut pieces = values.take(64).map(|piece| {
+        let mut piece = piece.to_string();
+        piece.retain(|c| !c.is_control());
         if quoted {
-            let mut piece = piece.to_string();
-            if piece.contains('\"') {
-                format!("\"{}\"", piece.replace("\"", "\\\""))
-            } else {
-                piece.insert(0, '"');
-                piece.push('"');
-                piece
-            }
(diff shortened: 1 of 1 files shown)


Enet4 (changes requested):
Thank you for your pull request!

The existing whitespace filtering mechanisms in place are only for trimming the end of text values in the file meta group. So in order to fix #187, there are two things yet to resolve:

- Any line break or control characters in the middle of the string will not be affected by trimming, so there should be an additional step to escape or replace those characters regardless of their position in the string.
- This needs to be applied to other text values aside from those in the file meta group. See line 740 for the most likely place to introduce this process.

I would greatly appreciate it if you reiterated on this PR to resolve both matters.

aaronzs:
@Enet4 I see. I'll look into both in the next few days.

aaronzs:
@Enet4 Thanks for the tips of making change to line 740. Please review my last update for this PR.

I also discard my previous commit, because I notice that #166 and #198 are dealing with "clean_str" related work.

dump/src/lib.rs
```diff
@@ -736,18 +736,14 @@ where
     let values = values.into_iter();
     let len = values.len();
     let mut pieces = values.take(64).map(|piece| {
+        let mut piece = piece.to_string();
+        piece.retain(|c| !c.is_control());
```
Enet4, line 740: It would be a bit too surprising to find a mismatching number of bytes and text length. Replacing the character would be preferable over removing it entirely.
```suggestion
        piece.replace(|c| c.is_control(), '�');
```
Later on, this could be extended with more options such as one for using DICOM-compliant character codes.

Enet4 (commented):
Thank you for this second iteration! If you attend to the concern below, I would be happy to merge this.

dump/src/lib.rs
aaronzs: I didn't find a nice way to keep the string length. Replacement with "UNKNOWN" makes sense to me 👍

Enet4 (approved):
Thank you for your contribution, @aaronzs.

Issue #181, opened 2021-09-30
[dump] Enable dumping more than one file 


Enet4:
Given a sequence of files, the dump tool should dump each of them in sequence.

Enet4:
Resolved by #207

PR #207, opened 2022-01-03
[dump] Enable dumping more than one file

natdash:
Addresses #181

### DIFF
--- a/dump/src/main.rs
+++ b/dump/src/main.rs
@@ -65,7 +65,7 @@ fn main() {
     options.no_text_limit(no_text_limit)
             .width(width)
             .color_mode(color);
-    
+    let fail_first = filenames.len() == 1 || fail_first;
     let mut errors: i32 = 0;
 
     for filename in &filenames {
@@ -76,7 +76,7 @@ fn main() {
                 if fail_first {
                     std::process::exit(ERROR_READ);
                 }
-                errors = errors + 1;
+                errors += 1;
             },
             Ok(obj) => {
                 if let Err(ref e) = options.dump_file(&obj) {
@@ -85,10 +85,10 @@ fn main() {
                     } else {
                         eprintln!("[ERROR] {}", e);
                         if fail_first {
-                        std::process::exit(ERROR_PRINT);
+                            std::process::exit(ERROR_PRINT);
                         }
                     }
-                    errors = errors + 1;
+                    errors += 1;
                 } // else all good
             }
         };
--- a/dump/src/main.rs
+++ b/dump/src/main.rs
@@ -25,8 +25,8 @@ fn os_compatibility() -> Result<(), ()> {
 /// Dump the contents of DICOM files
 #[derive(Debug, StructOpt)]
 struct App {
-    /// The DICOM file to read
-    file: PathBuf,
+    /// The DICOM file(s) to read
+    files: Vec<PathBuf>,
     /// whether text value width limit is disabled
     /// (limited to `width` by default)
     #[structopt(long = "no-text-limit")]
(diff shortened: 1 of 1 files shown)


dump/src/main.rs
```diff
@@ -46,36 +46,39 @@ fn main() {
...
+    for filename in filenames {
+        let obj = open_file(filename).unwrap_or_else(|e| {
+            report(e);
+            std::process::exit(ERROR_READ);
+        });
+        
+        match options
+            .no_text_limit(no_text_limit)
+            .width(width)
+            .color_mode(color)
```
Enet4, line 70: Might not make much of a difference, but these options only need to be set once, before traversing through `files`.

dump/src/main.rs
```diff
@@ -25,8 +25,8 @@ fn os_compatibility() -> Result<(), ()> {
 /// Dump the contents of DICOM files
 #[derive(Debug, StructOpt)]
 struct App {
-    /// The DICOM file to read
-    file: PathBuf,
+    /// The DICOM file(s) to read
+    files: Vec<PathBuf>,
```
Enet4, line 29: Let's make this argument required, so that an error is raised if the user tries to run this without passing any files.
```suggestion
    /// The DICOM file(s) to read
    #[structopt(required = true)]
    files: Vec<PathBuf>,
```

dump/src/main.rs
```diff
@@ -46,36 +46,39 @@ fn main() {
...
-            // handle broken pipe separately with a no-op
-        }
-        Err(e) => {
-            eprintln!("[ERROR] {}", e);
-            std::process::exit(ERROR_PRINT);
+
+    for filename in filenames {
+        let obj = open_file(filename).unwrap_or_else(|e| {
+            report(e);
+            std::process::exit(ERROR_READ);
```
Enet4, line 64: This `exit` here and the one in line 78 mean that the program is abruptly terminated if any of the files could not be opened, but there could be others on the list that could be dumped without issue.

I would propose only exiting immediately upon a new flag, something like `--fail-first`, otherwise it would accumulate the number of failed dump attempts and send an exit code based on that. But this is OK to be done in a separate PR if you wish.

Enet4 (commented):
Thank you for your pull request, @natdash! Changes overall look good and fulfill the given requirements. Please see if you can work on the improvements suggested inline.

natdash:
Thanks for your feedback, @Enet4! I've implemented the suggested improvements, will reply to your comments above.

dump/src/main.rs
natdash: Great point, this is done now.

dump/src/main.rs
natdash: Sure, I've addressed this.

dump/src/main.rs
natdash: I've updated this so that exiting immediately is optional, using the suggested flag `--fail-first` . Errors are being accumulated but are not currently used to send an exit code. Do you have a suggestion/preference for exit code logic?

Enet4: Good question. The current error code logic makes sense for single file dumping, but not for multiple files. So my idea is: if the user passed more than one file, let's make it so that the exit code is the number of files which failed to dump successfully (so if 2 files failed, exit with `2`). To the best of my knowledge, this is the behavior of DCMTK's dcmdump. It also doesn't collide with the single-file error codes because negative integers were used there.

natdash: That sounds great to me. I've pushed another commit to return these exit codes now.

Enet4 (commented):
Just a few more suggestions and this would be ready to merge!

natdash:
Thanks @Enet4! I've made the suggested changes now.

Enet4 (approved):
Nice! Thank you again for your contribution.

PR #209, opened 2022-01-03
Store SCP

vsaase:
Hi, I made this quick example of a C-STORE SCP. Would be happy for comments.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,5 +14,6 @@ members = [
     "scpproxy",
     "echoscu",
     "storescu",
+    "storescp",
     "pixeldata"
 ]
--- a/echoscu/src/main.rs
+++ b/echoscu/src/main.rs
@@ -119,7 +119,7 @@ fn create_echo_command(message_id: u16) -> InMemDicomObject<StandardDataDictiona
     // group length
     obj.put(DataElement::new(
         tags::COMMAND_GROUP_LENGTH,
-        VR::UI,
+        VR::UL,
         PrimitiveValue::from(8 + 18 + 8 + 2 + 8 + 2 + 8 + 2),
     ));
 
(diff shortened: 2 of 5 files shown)


vsaase:
needs some work to fill the proper File Meta Information before saving to disk

Enet4 (commented):
Thank you for this neat contribution! It has already helped me diagnose some issues in the `ul` crate.

One thing that intrigues me here: is there something missing in the [`AssociationServer` abstraction](https://docs.rs/dicom-ul/0.3.0/dicom_ul/association/server/struct.ServerAssociationOptions.html) that led you towards reading and writing PDUs on your own? It was originally pictured that SCPs would use either the `ul::association` module or something at a higher level of abstraction. It might be worth filling the gaps here rather than building the association logic from scratch.

One thing I see is that while `PDataWriter` exists, `PDataReader` does not. And it would be useful here in order to accumulate the sequence of PData values. Just filed #216. If there are other pain points, I would love to know.

vsaase:
> Thank you for this neat contribution! It has already helped me diagnose some issues in the `ul` crate.
> 
> One thing that intrigues me here: is there something missing in the [`AssociationServer` abstraction](https://docs.rs/dicom-ul/0.3.0/dicom_ul/association/server/struct.ServerAssociationOptions.html) that led you towards reading and writing PDUs on your own? It was originally pictured that SCPs would use either the `ul::association` module or something at a higher level of abstraction. It might be worth filling the gaps here rather than building the association logic from scratch.
> 
> One thing I see is that while `PDataWriter` exists, `PDataReader` does not. And it would be useful here in order to accumulate the sequence of PData values. Just filed #216. If there are other pain points, I would love to know.

I am happy you found it useful. I just didn't know about the AssociationServer abstraction at that time, I just looked at some of the other examples.

Enet4:
I have been reiterating on the store SCP, but now that my branch is rebased, I will need to commandeer this with a new pull request. Your baseline is still pretty much appreciated!

Issue #182, opened 2021-09-30
[storescu] Enable sending more than one file at once


Enet4:
Currently, `dicom-storescu` is only prepared to send one DICOM file at a time.

Given a sequence of local file paths, the tool should:

1. Traverse the data recursively to identify all valid DICOM objects
   - DICOMDIR support is nice to have, but not a strict requirement. Until better DICOMDIR support is added, the program should ignore files with this name.
2. Collect a list of requested presentation context
3. Reuse the DICOM association with the Store SCP to send each object in sequence.

Enet4:
Resolved in #211

PR #211, opened 2022-01-07
[storescu] Enable sending more than one file

natdash:
Addresses #182

### DIFF
--- a/storescu/src/main.rs
+++ b/storescu/src/main.rs
@@ -12,18 +12,20 @@ use dicom_ul::{
     pdu::{PDataValue, PDataValueType},
 };
 use smallvec::smallvec;
+use std::ffi::OsStr;
 use std::io::Write;
 use std::path::PathBuf;
 use structopt::StructOpt;
 use transfer_syntax::TransferSyntaxIndex;
+use std::collections::HashSet;
 
 /// DICOM C-STORE SCU
 #[derive(Debug, StructOpt)]
 struct App {
     /// socket address to STORE SCP (example: "127.0.0.1:104")
     addr: String,
-    /// the DICOM file to store
-    file: PathBuf,
+    /// the DICOM file(s) to store
+    files: Vec<PathBuf>,
     /// verbose mode
     #[structopt(short = "v", long = "verbose")]
     verbose: bool,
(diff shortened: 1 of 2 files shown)


storescu/src/main.rs
```diff
@@ -39,183 +41,215 @@ struct App {
...
+    for mut file in &mut dicom_files {
+        // TODO(#106) transfer syntax conversion is currently not supported
+        match check_presentation_contexts(file, scu.presentation_contexts()) {
+            Ok((pc, ts)) => {
+                file.pc_selected = Some(pc);
+                file.ts_selected = Some(ts);
+            },
+            Err(e) => {
+                if fail_first {
+                    eprintln!("Could not choose a transfer syntax: {}", e);
```
Enet4, line 131:
```suggestion
                eprintln!("Could not choose a transfer syntax: {}", e);
                if fail_first {
```
The error should still be reported to the user.

Enet4 (commented):
Thank you for working on this!

The requirement to scan the directory tree for files was included from the start in #211, because it makes the SCU much easier to work it. I would be grateful if you extended this contribution to make a recursive listing of files to send, so that the command works by passing a directory of DICOM files.

One other concern is that unless `verbose` is set, the tool is currently rather silent, giving no clue on how much work has done. Using a lib such as [`indicatif`](https://crates.io/crates/indicatif) to show the current progress should be easy to integrate and make the output much more appealing.

natdash:
Thanks for the feedback! I'll update to address and add in the missed functionality. Adding a progress bar sounds great, I'll work on that as well.

natdash:
I'm unsure why the CI check failed - it seems to be an error accessing a test DICOM file?

Enet4:
> I'm unsure why the CI check failed - it seems to be an error accessing a test DICOM file?

It's not much to worry about, every once in a while there appears to be a sporadic failure in `dicom-test-files`. Might be worth diagnosing further if it happens again.

Thank you for including the requested capabilities! I'll review and test this on my end once possible.

storescu/src/main.rs
```diff
@@ -39,183 +43,246 @@ struct App {
...
-    if verbose {
-        println!("Establishing association with '{}'...", &addr);
+        match check_file(file) {
+            Ok(dicom_file) => {
+                presentation_contexts.insert((dicom_file.sop_class_uid.to_string(),dicom_file.file_transfer_syntax.clone()));
+                dicom_files.push(dicom_file);
+            },
+            Err(e) => {
+                if verbose {
+                    println!("Error: {}", e);
```
Enet4, line 110:
```suggestion
                    println!("Could not open file {}: {}", file.display(), e);
```

Enet4 (commented):
I've been playing around and it functions pretty much as intended! For now, I have just a few more suggestions to apply before merging.

natdash:
That's great, thanks again! I've implemented the suggested improvements now.

Enet4 (approved):
Wonderful! In the meantime, I discovered some issues in the ul crate which also affect the store SCU (#215), but I'm already working on it.

Thank you for your contributions.

Issue #215, opened 2022-01-22
Client association options may lead to misguided presentation context negotiation


Enet4:
The current API for synchronous DICOM client association establishment uses an options type with these two helper methods:

- `with_abstract_syntax` will add a presentation context with the given abstract syntax (an SOP Class representing a service)
- `with_transfer_syntax` adds a transfer syntax to each presentation context defined above.

Turns out that this API has issues, because it misguides the association requester into proposing too ample presentation contexts when it needed more presentation context with a narrower scope. One simple example of this is a store SCU trying to send two DICOM files in a different transfer syntax:

- 1.dcm - MR Image Storage - Explicit VR Little Endian
- 2.dcm - X-Ray Angiographic Image Storage - JPEG Baseline (Process 1)

With the current API, one would do this:

```rust
let association = ClientAssociationOptions::new()
    .with_abstract_syntax("1.2.840.10008.5.1.4.1.1.4") // MR storage
    .with_abstract_syntax("1.2.840.10008.5.1.4.1.1.12.1") // angiographic storage
    .with_transfer_syntax("1.2.840.10008.1.2.1") // expl VR LE
    .with_transfer_syntax("1.2.840.10008.1.2.4.50") // JPEG
    .establish("storescp:104")?;
```

This creates two presentation contexts, but each of them will let the acceptor choose whichever transfer syntax. This means that it can choose the JPEG baseline transfer syntax for the presentation context designed to send the MR image, which is in Explicit VR Little Endian.

The possibility of adding multiple transfer syntaxes for all proposed presentation contexts is something which can be misleading and lead to unexpected issues caused by the lack of transcoding support. Instead, there must be a way to provide presentation contexts individually.

natdash:
Hi @Enet4 - I've been looking at this and can submit a PR to address if you want?

Enet4:
Hey @natdash, thank you for volunteering! I was already looking into this a week ago, but I did not have the time to get this done. If you are in position to send in a way to indicate presentation contexts reliably (as in, changing `ClientAssociationOptions` to have a fitting `with_presentation_context` method), I can review and propose some recommendations later.

Enet4:
Resolved by #217

PR #217, opened 2022-02-01
Update to presentation context handling

natdash:
Addresses #215


Enet4 (commented):
This looks great and to the point, thank you very much! `with_transfer_syntax` was clearly a footgun, and I would rather keep it away from the API right now. `with_abstract_syntax` was also nicely reconsidered to adopt only the two transfer syntaxes.

Please see the suggestions inline and don't forget to reformat the touched files with `rustfmt`.

Enet4 (approved):
Thank you very much!

PR #228, opened 2022-03-30
Add rust native decoder for jpeg pixeldata

### LINKED ISSUE #226: Compatibility with multiframe pixel data via GDCM
I have been looking to extract images from dicoms that are encoded via "JPEG_LOSSLESS_NON_HIERARCHICAL_FIRST_ORDER_PREDICTION".
With the gdcm feature this also worked nicely as long as it was single frame. However I also have series of images per file where I run into https://github.com/Enet4/dicom-rs/blob/master/pixeldata/src/gdcm.rs#L56.

I have been looking around how pydicom (what we currently use mostly) does this in combination with pillow and thought a similar approach could be implemented for dicom-rs (build on top of the image crate so it should also be fully rust).
https://github.com/pydicom/pydicom/blob/master/pydicom/pixel_data_handlers/pillow_handler.py
A similar decoder to https://github.com/Enet4/dicom-rs/blob/master/encoding/src/adapters/rle_lossless.rs where we point the image crate to the u8 vectors that represent each image.

Before I invest a lot further time, is this something you would consider merging if I provide an implementation? Does it need to be gated behind a feature flag (and do you have a recommendation on how to implement it)? Maybe other things that could go wrong that I might have missed and makes the above approach unfeasible?

ingwinlu:
Provides jpeg decoding for single and multiframe dicoms. Also supports frames distributed onto multiple fragments.

Still to tackle:
- [x] Multiframe images only show up with 'data' on the first frame, rest is empty
- [x] remove untested jpeg encoder settings in tranfer syntax
- [x] change test so it writes pics to dedicated out dir (so there are no clashes between tests)
- [x] make snafu context error on jpeg parsing work
- [x] implement parsing of frames split into multiple fragments
- [x] ~~color offset (lots of green in one test case~~ Might be related to some other issue. not sure that it can be handled in decoder
- [x] ~~Slow in non release mode (maybe some bounds checks?)~~ Should not be fault of jpeg decoder

### DIFF
--- a/pixeldata/src/lib.rs
+++ b/pixeldata/src/lib.rs
@@ -857,7 +857,7 @@ mod tests {
         }
 
         #[rstest]
-        #[should_panic(expected = "Could not decode jpeg in frame")]
+        #[should_panic(expected = "Format(\"JPGn(7) marker found where not allowed\")")]
         #[case("pydicom/emri_small_jpeg_ls_lossless.dcm", 10)] // crashes, but not sure if file is ok
         #[case("pydicom/color3d_jpeg_baseline.dcm", 120)]
         // TODO: figure out why slow (40s) in non release mode, maybe out of bounds checks? also very green and only first frame has a picture in it
--- a/pixeldata/src/lib.rs
+++ b/pixeldata/src/lib.rs
@@ -679,9 +679,6 @@ mod tests {
     use super::*;
     use dicom_object::open_file;
     use dicom_test_files;
-    use rstest::rstest;
-    use std::fs;
-    use std::path::Path;
 
     #[test]
     fn test_to_ndarray_rgb() {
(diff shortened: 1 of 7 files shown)


ingwinlu:
I would like to understand the `JPGn(7) marker found where not allowed` a bit better (and maybe rule out it origins from some dicom padding or similar.
However I am afraid I would need some pointers.

As a first step this already handles a good chunk of jpeg encoded transfers (also the one I originally needed).

Enet4:
Thank you for working on this, @ingwinlu! I have yet to perform a thorough review, but here are a few comments from what I could gather so far:

- It would be nice if we could avoid the extra allocations in `encoded_frames`, by putting them all in a flat byte vector.
- The performance of the pixel data decoding test against case 6 can be improved a bit by compiling some key dependencies with some optimizations.
Try this inside the root Cargo.toml:
```toml
# optimize JPEG decoder to run tests faster
[profile.dev.package."jpeg-decoder"]
opt-level = 2

# optimize deflate to run tests faster
[profile.dev.package."deflate"]
opt-level = 2
```
In any case, this one contains 120 frames to decode, so performance won't be that great. I ought to consider a way to make this a bit "lazier" eventually.

- I'm not sure what you mean with the ` JPGn(7) marker found where not allowed` message, but if this has to do with the JPEG 2000 examples, it is natural that they won't work. JPEG 2000 is a pretty distinct format from other JPEG encodings.

ingwinlu:
> I'm not sure what you mean with the JPGn(7) marker found where not allowed message, but if this has to do with the JPEG 2000 examples, it is natural that they won't work. JPEG 2000 is a pretty distinct format from other JPEG encodings.

The examples are encoded with jpeg - ls, not jpeg2000 as far as i understand:
1.2.840.10008.1.2.4.80 (JPEG - LS Lossless Image Compression)

Hence I assumed that maybe the decoder could handle them and I just feed the data incorrectly to it.

> The performance of the pixel data decoding test against case 6 can be improved a bit by compiling some key dependencies with some optimizations.

I can include this in the pr, locally i just mostly run tests against other cases and afterwards did a test with --release enabled to achive the same effect.

> It would be nice if we could avoid the extra allocations in encoded_frames, by putting them all in a flat byte vector.
I will look into it

pixeldata/src/lib.rs
```diff
@@ -865,8 +737,172 @@ mod tests {
...
+            let obj = open_file(test_file).unwrap();
+            let pixel_data = obj.decode_pixel_data().unwrap();
+            assert_eq!(pixel_data.number_of_frames, frames);
+
+            let output_dir = Path::new(
+                "../target/dicom_test_files/_out/test_parse_jpeg_encoded_dicom_pixel_data",
+            );
+            fs::create_dir_all(output_dir).unwrap();
+
+            for i in 0..pixel_data.number_of_frames {
```
Enet4, line 897: Fetch less frames to make the test faster.
```suggestion
            for i in 0..pixel_data.number_of_frames.min(16) {
```

Enet4 (changes requested):
After a deeper investigation, I realized that some of the `pixeldata` logic was at fault for the green-looking images: the original DICOM file describes a photometric interpretation of `YBR_FULL_422`, but the decoded outcome from the adapter is already in RGB! This means that the rest of the pixel data conversion logic was applying the color space conversion _again_, even though it was not needed.

I will take care of adjusting the `pixeldata` module so that this does not happen. Basically, I will document that the adapters are expected to perform this color space conversion. On the plus side, this indeed means that the JPEG decoding process appears to be working as intended.

There is only one major matter left to solve in this PR. Having the multi-frame test case `color3d_jpeg_basline.dcm` is important, but it makes the tests substantially slower. We can boost the performance with two changes.

Please add the following to the end of `Cargo.toml`:
```toml
# optimize JPEG decoder to run tests faster
[profile.dev.package."jpeg-decoder"]
opt-level = 2

# optimize deflate to run tests faster
[profile.dev.package."deflate"]
opt-level = 2
```

And in `test_parse_jpeg_encoded_dicom_pixel_data`, limit the number of decoded frames to 16, as suggested inline.

It took 2m 3s on my personal machine to run `cargo t` on a fresh project, whereas after these tweaks it took around 1m 22s, so it makes quite a difference.

pixeldata/src/lib.rs
ingwinlu: I assume you mean .take here?

Enet4 (approved):
Without any major concerns left (#229 will be tweaked to account for this), this should be ready to enter. Thank you for your continued support.

PR #229, opened 2022-04-11
Adjust photometric interpretation of decoded pixel data

Enet4:
When using native pixel data adapters, the result of decoding should always be in MONOCHROME2 or RGB, regardless of the original encoding. This tweaks `pixeldata` so that certain color space conversions are not performed redundantly.

### DIFF
--- a/pixeldata/src/lib.rs
+++ b/pixeldata/src/lib.rs
@@ -573,7 +573,7 @@ where
                 number_of_frames,
                 photometric_interpretation: new_pi,
                 samples_per_pixel,
-                planar_configuration,
+                planar_configuration: 0,
                 bits_allocated,
                 bits_stored,
                 high_bit,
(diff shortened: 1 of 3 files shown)


ingwinlu:
If #228 is merged before we can also remove the comment about offcolor output from the test.

pixeldata/src/lib.rs
```diff
@@ -884,10 +924,10 @@ mod tests {
         #[should_panic(expected = "Unsupported(SamplePrecision(12))")]
         #[case("pydicom/JPEG-lossy.dcm", 1)]
         //
-        // works, but output has coloroffset to what is produced by other tools
+        // works fine
         #[case("pydicom/color3d_jpeg_baseline.dcm", 120)]
         //
-        // Works fine
+        // wWorks fine
```
ingwinlu, line 930: typo

ingwinlu (changes requested):

ingwinlu (approved):

Issue #212, opened 2022-01-13
Incompatibility with inventory 0.2


Enet4:
The `inventory` dependency, which is used for plugging in transfer syntax implementations at compile time, has a new version which requires the injected value to be const evaluated. This is incompatible with what we currently have in DICOM-rs because `TransferSyntax<A = DynDataRWAdapter, B = DynPixelRWAdapter>` is generic, and was automatically type-erased via the `erased` method.

PR #233, opened 2022-04-18
Implement support for multiframe image decoding via gdcm

ingwinlu:
this is mostly done to test https://github.com/pevers/gdcm-rs/pull/7 for now and maybe get some visibility on how to improve it right away.

Does not crash but for the 2 test files I added it either produces black images or images with 'bad' colors.

### DIFF
--- a/pixeldata/Cargo.toml
+++ b/pixeldata/Cargo.toml
@@ -18,7 +18,7 @@ dicom-transfer-syntax-registry = { path = "../transfer-syntax-registry", version
 dicom-dictionary-std = { path = "../dictionary-std", version = "0.5.0" }
 snafu = "0.7.0"
 byteorder = "1.4.3"
-gdcm-rs = { version = "0.3.0", optional = true }
+gdcm-rs = { version = "0.4.0", optional = true }
 rayon = "1.5.0"
 ndarray = "0.15.1"
 ndarray-stats = "0.5"
--- a/pixeldata/src/gdcm.rs
+++ b/pixeldata/src/gdcm.rs
@@ -4,9 +4,11 @@ use crate::*;
 use dicom_encoding::adapters::DecodeError;
 use dicom_encoding::transfer_syntax::TransferSyntaxIndex;
 use dicom_transfer_syntax_registry::TransferSyntaxRegistry;
-use gdcm_rs::{decode_single_frame_compressed, GDCMPhotometricInterpretation, GDCMTransferSyntax};
-use std::convert::TryFrom;
-use std::str::FromStr;
+use gdcm_rs::{
+    decode_multi_frame_compressed, decode_single_frame_compressed, Error as GDCMError,
+    GDCMPhotometricInterpretation, GDCMTransferSyntax,
+};
+use std::{convert::TryFrom, str::FromStr};
 
 impl<D> PixelDecoder for FileDicomObject<InMemDicomObject<D>>
 where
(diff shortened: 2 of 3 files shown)


Enet4:
Thank you for working on this, @ingwinlu!

I must warn you though, as I am working on #122, the crate's code will be subjected to several changes with an upcoming pull request. So we may need to rebase this later.

Enet4:
Just a heads-up, #122 has been merged, so this PR will need to be updated accordingly.

I have scheduled for a v0.5.0 release very soon, but this contribution may become part of DICOM-rs in subsequent versions.

ingwinlu:
with the relese of gdcm-rs 0.4.0 this should hopefully build now

ingwinlu:
hot damn, some more support for encoded pixel data ready to go!

ingwinlu:
Closes #226

ingwinlu:
Any feedback on this?

Enet4 (commented):
The integration with `gdcm-rs` appears to be in good order. However, I noticed that this module is still only retrieving the first frame when the pixel data is native. Can you look into line 105 and remove the unwanted slicing like in lib.rs:1577?

ingwinlu:
Short update mostly for myself so I don't forget.

When looking at the above mentioned code https://github.com/Enet4/dicom-rs/blob/c08c3511b651a47dda452e24dac11c47aeb43eff/pixeldata/src/lib.rs#L1577 I could not really understand what slicing is refered to and looked into finding some way to test the behaviour with the test dicom files.

While doing this I also noticed that the branch at https://github.com/Enet4/dicom-rs/blob/c08c3511b651a47dda452e24dac11c47aeb43eff/pixeldata/src/lib.rs#L1574 apparently is not covered by tests at placing a panic! macro there does not influence test results.

Enet4:
> When looking at the above mentioned code
> 
> https://github.com/Enet4/dicom-rs/blob/c08c3511b651a47dda452e24dac11c47aeb43eff/pixeldata/src/lib.rs#L1577
> 
> I could not really understand what slicing is refered to and looked into finding some way to test the behaviour with the test dicom files.

I suspect that the slicing was originally introduced in order to let the user retrieve the first frame from a multi-frame image. Now that we want multi-frame retrieval to work, any slicing can be deferred to eventual calls in `DecodedPixelData`.

> While doing this I also noticed that the branch at
> 
> https://github.com/Enet4/dicom-rs/blob/c08c3511b651a47dda452e24dac11c47aeb43eff/pixeldata/src/lib.rs#L1574
> 
> apparently is not covered by tests at placing a panic! macro there does not influence test results.

So yes, the current implementation is not very ideal. It was probably a consequence of the changes in #152 or #212. That part is unreachable because we already check to ensure that the transfer syntax is fully supported. 

With that said, this reminds me that we are very close to supporting _Encapsulated Uncompressed Explicit VR Little Endian_ (1.2.840.10008.1.2.1.98), which is a fairly recent transfer syntax.

ingwinlu:
@Enet4 my problem is that I don't see where there is any slicing occurring? I also see no behaviour that would just return the first frame.

Enet4:
> @Enet4 my problem is that I don't see where there is any slicing occurring?

With the current setup, slicing is expected to be done after the decoding, when turning the `DecodedPixelData` into something else. Maybe at a later stage the decoding can be deferred to a later stage for efficiency.

> I also see no behaviour that would just return the first frame.

Currently, it does when feature-gated on `gdcm`:
https://github.com/Enet4/dicom-rs/blob/c08c3511b651a47dda452e24dac11c47aeb43eff/pixeldata/src/gdcm.rs#L105-L110

The code above fetches only the first frame by slicing on the first `W*H*C*bpp/8` bytes. It is an artifact of the first iteration of `dicom-pixeldata` and should be removed. Please let me know if you need further assistance, or whether you prefer that I take over the pull request.

ingwinlu:
For some reason when reading  `Can you look into line 105 and remove the unwanted slicing like in lib.rs:1577?` I only parsed the second reference and never looked into the first. And apparently was way to confused by this to even notice that lib.rs:1577 has no influence with gdcm anyway.

I pushed the change in cc12a78.

Enet4 (approved):
Great! Thank you, @ingwinlu.

PR #254, opened 2022-06-06
[parser] Increase reader robustness to stray item end

### LINKED ISSUE #253: Unexpected token ItemEnd
Hello,

The attached DICOM test file makes dicom-rs' dicom-dump fail with error message "[ERROR] Unexpected token ItemEnd". dcm4che and dcmtk are both be able to dump its metadata.

Maybe something in parsing needs to be made more tolerant or robust?

[dummy.dcm.zip](https://github.com/Enet4/dicom-rs/files/8794355/dummy.dcm.zip)

Enet4:
Resolves #253.

- skip ItemDelimitationItem header when not in a sequence
- improve data set reader validator to account for missing tokens

CC @troelsarvin

### DIFF
--- a/parser/src/dataset/read.rs
+++ b/parser/src/dataset/read.rs
@@ -419,6 +419,15 @@ where
                 Ok(DataElementHeader {
                     tag: Tag(0xFFFE, 0xE00D),
                     ..
+                }) if self.seq_delimiters.is_empty() => {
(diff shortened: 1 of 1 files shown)


troelsarvin:
I confirm that the change has made it possible for my code to parse the DICOM headers, thanks.

PR #262, opened 2022-07-03
added PersonName support to Value

jmlaka:
Hello Eduardo,
I don't know if this would be useful but I needed to handle Dicom PN.
It only uses borrowed values and family, given names are not optional but mandatory. 

Removed one typo from a previous Doc test


core/src/value/person_name.rs
```diff
@@ -0,0 +1,326 @@
+use crate::dicom_value;
```
Enet4, line 1: If this module has to be public, then a small root-level piece of documentation is recommended. It should at least explain that this aims for an implementation of DICOM values with the VR PN (person name) as per [PS3.5 sect 6.2](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html).

core/src/value/person_name.rs
```diff
@@ -0,0 +1,326 @@
...
+            family,
+            middle,
+            given,
+            suffix,
+        }
+    }
+    pub fn to_dicom_string(&self) -> String {
+        let mut name = format!("{}^{}", self.family, self.given);
+        match self.middle {
+            Some(middle) => name.push_str(&format!("^{}", middle)),
```
Enet4, line 79: Clippy is complaining about this, probably because by calling `name.push('^')` and `name.push_str(middle)`, we save a string allocation. Same for the other cases below.

core/src/value/person_name.rs
```diff
@@ -0,0 +1,326 @@
...
+    }
+    pub fn given(&self) -> &str {
+        self.given
+    }
+    pub fn middle(&self) -> Option<&str> {
+        self.middle
+    }
+
+    // exact match to dicom formatted string
+    pub fn is(&self, search: &str) -> bool {
```
Enet4, line 115: This method is a bit confusing and expensive. Maybe one should rather invite the user into converting both operands to a common type: either both `PersonName` or both strings. Then the semantics are much clearer. Otherwise, they may be lured into calling this method multiple times assuming it to be a plain component-by-component check when it actually converts `self` into a string underneath.

core/src/value/person_name.rs
```diff
@@ -0,0 +1,326 @@
...
+        self.to_dicom_string() == search
+    }
+
+    pub fn contains(&self, search: &str) -> bool {
```
Enet4, line 119: This is another method which I believe can be a heavy source of confusion, since they could think that it means "whether the name contains the component `search`", rather than "whether any component contains the text `search` anywhere". Should the person name  "Smith^Johnathan" contain "John"? If this is for a search tool, one would rather build a proper text index.

This calls for either a rethinking of this method altogether, or a very rigorous documentation.

Let's also not forget to document all public methods.

core/src/value/person_name.rs
```diff
@@ -0,0 +1,326 @@
...
+            Some(s) => s.contains(search),
+            None => false,
+        };
+
+        prefix || self.family.contains(search) || self.given.contains(search) || suffix
+    }
+
+    pub fn from_slice(slice: &'a str) -> Result<PersonName<'a>> {
+        // As per standard, the slice can contain leading "^" separators.
+        // This implementation treats family and given name as mandatory, thus this error.
```
Enet4, line 134: Please make this implementation support missing family/given names.

Enet4 (changes requested):
Hello @jmlaka! Thank you for sending another contribution in.

A person name module can indeed be helpful for systems working such names semantically. More than once have I received cases in which the system didn't bother encoding each component properly.

However, it is important for the baseline implementation to be more flexible so as to avoid issues caused by less ordinary names that could hinder the process for bystanders. I wouldn't require ideographic/phonetic character groups to be supported in an initial phase, but I would not want the _given and family names to be mandatory_, especially when the standard imposes that any of them can be an empty string.

I invite you to continue working on this module so that it accepts person names in which any of the components may be empty. It might also be more usable to employ the builder pattern to construct a person name from scratch. Feel free to follow up if you have any questions or could use some guidance.

Other minor concerns inline with the code.

jmlaka:
Please see

Enet4:
I took the liberty of doing some final touches on this pull request. A summary:

- extend and adjust documentation and tests (to also include doctests and cover the person name builder)
- make `PersonNameBuilder` non-consuming
- add `From<PersonNameBuilder>` and `From<&mut PersonNameBuilder>` impls to `PersonName`
- rename `PersonName::from_slice` to `PersonName::from_str`
- rename `DicomValue::as_person_name` to `DicomValue::to_person_name` for consistency with the rest of the API 

In any case, the module seems robust and well constructed. @jmlaka  I am likely to merge this soon, but feel free to look over the extra commits in case you have any concerns or questions.

jmlaka:
Thanks for your insights @Enet4 ! Everything is fine with me.

PR #263, opened 2022-07-06
Add items helper

ingwinlu:
Allows for easy access to sequential data for example when trying
to access data stored in CONTENT_SEQUENCE tags commonly used in
SR files.

### DIFF
--- a/core/src/header.rs
+++ b/core/src/header.rs
@@ -484,10 +484,9 @@ where
         self.value().to_multi_datetime(default_offset)
     }
 
-    /// Retrieve items stored in value
+    /// Retrieve the items stored in a sequence value.
     ///
-    /// If value is not a sequence return None.
-    ///
+    /// Returns `None` if the value is not a sequence.
     pub fn to_items(&self) -> Option<&[I]> {
         self.value().items()
     }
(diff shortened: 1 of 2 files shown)


core/src/header.rs
```diff
@@ -483,6 +483,14 @@ where
     ) -> Result<Vec<DicomDateTime>, ConvertValueError> {
         self.value().to_multi_datetime(default_offset)
     }
+
+    /// Retrieve items stored in value
+    ///
+    /// If value is not a sequence return None.
+    ///
+    pub fn to_items(&self) -> Option<&[I]> {
```
Enet4, line 491: As per the [C-CONV](https://rust-lang.github.io/api-guidelines/naming.html#c-conv) guideline. This should rather be named `items` (named as a getter) or `as_items` (thus treating as a borrowed view over the value). Otherwise, one could expect `to_items` to return an independent list of items, which is not intended here.

```suggestion
    pub fn to_items(&self) -> Option<&[I]> {
```

Enet4 (changes requested):
Thank you for the nice suggestion! We can bring it in after addressing the concerns inline.

ingwinlu:
adjusted the helper to be called "items"

Enet4 (approved):
All right, all is in order. Thank you!

PR #268, opened 2022-07-15
[object] Add a few more iterator-related methods

Enet4:
This extends some of the types in the `object` crate which provide iterators to the contents of the DICOM objects.

This might help to make the object API a bit more ergonomic.

### DIFF
--- a/object/src/lib.rs
+++ b/object/src/lib.rs
@@ -473,6 +473,18 @@ where
     }
 }
 
+impl<'a, O> IntoIterator for &'a FileDicomObject<O>
+where
+    &'a O: IntoIterator,
+{
+    type Item = <&'a O as IntoIterator>::Item;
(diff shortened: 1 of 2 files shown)


ingwinlu:
would it also be possible to get a mutable iterator?

usecase would be opening a file, iterating over it mutably and removing or modifying certain tags (think anonymization) and later saving it again.

Or would it be better to just clone + consume + build a new file in this case?

Enet4:
@ingwinlu So yes, a mutable iterator over the data elements in an object is not possible here, otherwise it could break the invariants of the object by letting the user change a DICOM tag.

A processing pipeline such as that one could be done in one of the following ways:

- Transforming the whole object into an iterator, adapting it, and collecting it back into an object;
- Using the `retain` method (now available in this PR);
- Creating a vector of tags which are present in the object and sequentially performing the intended look-ups and manipulations.

PR #274, opened 2022-08-26
[echoscu|storescu] Change SCP address parameter to enable specifying AE title

### LINKED ISSUE #204: [storescu|echoscu] Compact way to set called AE title
As a user of these tools, I would like to specify the called application entity title alongside the AE host and address, making it compact and easier to remember.

Instead of this:
```sh
dicom-storescu --called-ae-title PACS-ARCHIVE localhost:104
```

I would be able to write:

```sh
dicom-storescu PACS-ARCHIVE@localhost:104
```

Enet4:
This resolves #204, allowing users to specify the called AE title alongside the socket address with the optional syntax `called-ae-title@socketaddress`. The option `called-ae-title` overrides the AE title in the address with a warning.
README files were updated accordingly.

CC @Almeida-a

### DIFF
--- a/echoscu/README.md
+++ b/echoscu/README.md
@@ -25,10 +25,18 @@ FLAGS:
     -v, --verbose    verbose mode
 
 OPTIONS:
-        --called-ae-title <called-ae-title>      the called AE title [default: ANY-SCP]
+        --called-ae-title <called-ae-title>
+            the called Application Entity title, overrides AE title in address if present [default: ANY-SCP]
+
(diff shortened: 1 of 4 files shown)


Almeida-a:
I would just like to note that, in the help page, the file token has the optional tag '[...]' instead of the required one '<...>'.

Is it on purpose?

Enet4:
> I would just like to note that, in the help page, the file token has the optional tag '[...]' instead of the required one '<...>'.
> 
> Is it on purpose?

Yes, the `files` parameter was set up to be optional by mistake. It is not a regression from this PR, so one can take care of that separately.

PR #278, opened 2022-09-14
[encoding] Extend TextCodec with decode_to and encode_to

Enet4:
This extends the `TextCodec` trait to also have `encode_to` and `decode_to` methods, which take a heap allocated buffer value instead of returning a new one. This makes it possible to reduce the number of allocations needed in some cases.

It is a breaking change because the trait is not sealed and implementations are required to provide definitions for these new methods.

### DIFF
--- a/encoding/src/text.rs
+++ b/encoding/src/text.rs
@@ -71,15 +71,36 @@ pub trait TextCodec {
     /// `TextCodec` is often used as a trait object.
     fn name(&self) -> &'static str;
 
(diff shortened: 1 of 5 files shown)


Enet4:
I will put this on hold until attempts are made to integrate with new text encoding libraries.

Enet4:
Closing because I'm not sure this is the right direction for text encoding/decoding, and it's gone far behind by other changes.

PR #299, opened 2022-10-22
- fix person_name whitespace trim

jmlaka:
Building PersonName from slice requires whitespace trimming


Enet4 (approved):
Indeed, this change is sound and now covered! Thank you @jmlaka. :+1:

Issue #286, opened 2022-10-02
Encoding to JPEG transfer syntaxes


Enet4:
Decoding JPEG-based transfer syntaxes is already supported, but support for encoding a native DICOM object to would be good to have.

In a first phase, this amounts to implementing the `encode` method of the `PixelRWAdapter` implementation for `JPEGAdapter`.

Almeida-a:
I think I can take this issue.

Is there a way to test this new implementation after it is finished?

Enet4:
> Is there a way to test this new implementation after it is finished?

By lack of a proper transcoding module (efforts are ongoing!), we can test the adapter in isolation by encoding one of the test files, reading the output as a JPEG file using the decoder and checking whether some meta-properties are correct (width, height...). Pixel data check could be done too, but we would need to allow an error margin to account for loss.

PR #303, opened 2022-10-25
Encoding to JPEG transfer syntaxes

Almeida-a:
This is a first version of an implementation regarding issue #286.

It it not tested yet, however, any suggestion regarding that, or other mistakes, would be appreciated!

---

Depends on #361

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -615,6 +615,7 @@ dependencies = [
  "byteordered",
  "dicom-core",
  "dicom-encoding",
+ "dicom-test-files",
  "jpeg-decoder",
  "jpeg-encoder",
  "lazy_static",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -616,6 +616,7 @@ dependencies = [
  "dicom-core",
  "dicom-encoding",
  "jpeg-decoder",
+ "jpeg-encoder",
  "lazy_static",
  "tracing",
 ]
@@ -1095,6 +1096,12 @@ checksum = "bc0000e42512c92e31c2252315bda326620a4e034105e900c98ec492fa077b3e"
 dependencies = [
  "rayon",
 ]
+
+[[package]]
+name = "jpeg-encoder"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2cf3affe27ffd9f1992690ec7575568b222abe9cb39738f6531968aca8e64906"
 
 [[package]]
 name = "js-sys"
--- a/transfer-syntax-registry/Cargo.toml
+++ b/transfer-syntax-registry/Cargo.toml
@@ -18,7 +18,7 @@ inventory-registry = ['dicom-encoding/inventory-registry']
 # natively implemented image encodings
 native = ["jpeg", "rle"]
 # native JPEG support
-jpeg = ["jpeg-decoder"]
+jpeg = ["jpeg-decoder", "jpeg-encoder"]
 # native RLE lossless support
 rle = []
 rayon = ["jpeg-decoder?/rayon"]
@@ -32,6 +32,10 @@ tracing = "0.1.34"
 
 [dependencies.jpeg-decoder]
 version = "0.3.0"
+optional = true
+
+[dependencies.jpeg-encoder]
+version = "0.5.1"
 optional = true
 
 [package.metadata.docs.rs]
--- a/transfer-syntax-registry/Cargo.toml
+++ b/transfer-syntax-registry/Cargo.toml
@@ -40,3 +40,6 @@ optional = true
 
 [package.metadata.docs.rs]
 features = ["native"]
+
+[dev-dependencies]
+dicom-test-files = "0.2.1"
--- a/transfer-syntax-registry/src/adapters/jpeg.rs
+++ b/transfer-syntax-registry/src/adapters/jpeg.rs
@@ -1,8 +1,10 @@
 //! Support for JPG image decoding.
 
+use dicom_core::ops::{AttributeAction, AttributeOp};
+use dicom_core::Tag;
 use dicom_encoding::adapters::{
-    decode_error, DecodeResult, EncodeOptions, EncodeResult, PixelDataObject, PixelDataReader,
-    PixelDataWriter,
+    decode_error, encode_error, DecodeResult, EncodeOptions, EncodeResult, PixelDataObject,
+    PixelDataReader, PixelDataWriter,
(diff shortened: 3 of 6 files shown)


encoding/src/adapters/jpeg.rs
```diff
@@ -1,15 +1,54 @@
...
+
+    /// Encode DICOM image data into jpeg encoding
+    fn encode(
+            &self,
+            src: &dyn PixelDataObject,
+            options: EncodeOptions,
+            dst: &mut Vec<u8>,
+        ) -> EncodeResult<()> {
+
+            let cols = src.cols().unwrap();
```
Enet4, line 23: We must not `unwrap` here. Passing an object without pixel data would be enough for the whole thread to panic, with little opportunity for a graceful recovery.

For the lack of more specific error variants, it should be possible to do this (requires `use snafu::ResultExt`):

```suggestion
            let cols = src.cols()
                .whatever_context("Missing image DICOM attribute Columns");
```

The equivalent change should be done for all other `unwrap()` calls.

encoding/src/adapters/jpeg.rs
```diff
@@ -1,15 +1,54 @@
...
+
+            let nr_frames = src.number_of_frames().unwrap_or(1) as usize;
+
+            let quality = options.quality.unwrap_or(100);
+        
+            let frame_size = (
+                cols * rows * samples_per_pixel * (bits_allocated / 8)
+            ) as usize;
+
+            let color_type = ColorType::Luma;
```
Enet4, line 36: It is not right to assume this color type. We need to check the image's photometric interpretation and proceed based on that and the number of samples per pixel.

encoding/src/adapters/jpeg.rs
```diff
@@ -1,15 +1,54 @@
...
+
+            for i in 0..nr_frames {
+                // Encode the data
+                let frame_uncompressed = src.fragment(i).unwrap();
+                let frame_encoded = Vec::with_capacity(frame_size);
+                let encoder = Encoder::new(frame_uncompressed, quality);
+                encoder.encode(&frame_encoded, cols, rows, color_type).unwrap();
```
Enet4, line 43: The two frame inputs are swapped. `Encoder::new` receives the writer and `Encoder::encode` takes the data to encode.

```suggestion
                let encoder = Encoder::new(&mut frame_encoded, quality);
                encoder.encode(&frame_uncompressed, cols, rows, color_type)
                    .whatever_context("could not encode frame to JPEG")?;
```

encoding/src/adapters/jpeg.rs
```diff
@@ -1,15 +1,54 @@
...
+            let frame_size = (
+                cols * rows * samples_per_pixel * (bits_allocated / 8)
+            ) as usize;
+
+            let color_type = ColorType::Luma;
+
+            for i in 0..nr_frames {
+                // Encode the data
+                let frame_uncompressed = src.fragment(i).unwrap();
+                let frame_encoded = Vec::with_capacity(frame_size);
```
Enet4, line 41: Or maybe we could use `dst` directly, thus skipping the allocation of a new vector on each frame?

Enet4 (changes requested):
Thank you for working on this. Please follow the suggestions inline to resolve the current issues.

Also, the format does not look standard. Running `cargo fmt` on the file should fix this.

---

As for the testing, we are missing a `PixelDataObject` implementation in this context. Let's devise a struct that keeps the necessary attributes in memory, then we can think of a way to construct them from test files.

encoding/src/adapters/jpeg.rs
Almeida-a: The PixelDataObject variable doesn't provide that attribute. Is there another way to extract it?

Enet4: So it is... 🤔 One other gap to attend to in future versions. We can use `SamplesPerPixel` to validate the inputs for now.

encoding/src/adapters/jpeg.rs
Almeida-a: This does not seem to work, unless I'm missing something, because there is an unsatisfied trait bound for the EncoderError struct, `FromString` I think from snafu.

I could instead return an EncodeError::CustomEncodeError instead (and maybe a MissingAttributeSnafu in the future, if it's appropriate).

Enet4: So you're right, some attributes were missing in the `CustomEncodeError` variant, so it still doesn't work with `whatever_context`. With #251 this should work more smoothly. So we can either try to get this implementation to work under the current limitations or we can base this contribution on #251. Now that v0.5.3 is out, it may be a good time to start pushing all breaking changes in for an eventual v0.6.0.

Almeida-a: It seems to be working now. I will push these changes after I finish the unit tests.

Almeida-a:
Right now I just need to finish the unit tests, but since I am encountering some errors and I am still a beginner in rust, I'm taking longer to complete the rest of this pr (sorry for that); but if at any point there is a need to more quickly finish this, I can commit my work up until this point.

Enet4:
Hey @Almeida-a, don't worry. As I intend to change the pixel data adapter API soon, it might be worth stalling this a bit more while I set up the new design, then we can work together on migrating the implementations. I hope that this works for you.

Enet4:
Update: #361 is what I have in mind. When you have the time, please rebase this branch against it and try to adjust the implementation. Feel free to ask if you need any help.

Enet4:
I took the liberty of rebasing this PR on top of #361 to start trying things out. I also took care of error handling in the encoder. Next I believe that this would call for a test suite to cover converting some pixel data to JPEG and back.

Enet4:
New stuff:

- Fixed some calculations being done at `u16` instead of `usize`, which would lead to overflow in medium/large resolutions.
- Added a test suite for writing and reading JPEG directly with adapters.

I found a few more things worth addressing at #361, but the changes to be made here should be minimal. I will take care of rebasing afterwards.

Enet4:
@Almeida-a Can you have a look and see if the new changes make sense to you? It should hopefully be ready to merge soon.

Almeida-a:
On it!

Enet4 (approved):
It looks good to me now. I will just wait for your signal. :+1:

Almeida-a:
Looks good, thank you for your assistance!

Enet4:
Great! Thank you for the initiative @Almeida-a!

PR #307, opened 2022-11-09
Findscu typo

jmlaka:
Fixed a typo
now negotiated transfer syntax is used to read / write datasets instead of IMPL_VR_LE


Enet4 (approved):
That's right! Thank you for spotting this. :+1:

PR #308, opened 2022-11-20
ClientAssociation release() and abort() take &mut self instead of mut self

jmlaka:
Hello,
I stumbled upon this while incorporating `ClientAssociation` into some higher-level API.
Can we change this?
Makes working with the ClientAssociation easier.


Enet4:
> Makes working with the ClientAssociation easier.

Could you share some details about your use case? This API was designed to consume the association because the connection should be dropped afterwards (AFAIK any further interactions would be non-standard).
If we are to change these methods to only make the release exchange or send the abort message without dropping the connection, one would also need to advise consumers not to continue using it afterwards.

PR #310, opened 2022-11-22
- replace deprecated chrono calls

jmlaka:
Replaces all `chrono` calls inside `dicom::core::` to be deprecated in chrono 0.4.23
One change in `dump` crate

Leaves  `parser::stateful::decode::StatefulDecoder::new` unchanged


Enet4 (commented):
Looks OK to me. Some of the changes just become less readable, but I suppose the latest changes to the API intend to favour explicit error handling.

There were two other lines still using `FixedOffset::east`, but I took the liberty of updating the two here.

Enet4 (approved):
It seems to be in order now. Thank you. :+1:

PR #312, opened 2022-11-25
Pixel encapsulation

dougyau:
Add the function to encapsulate a vector of frames. The function will return a PixelSequence which can be used with DataElement::new() to add the tag to the dataset.

```
let mut frames: Vec<Vec<u8>> = ...

let encapsulated_pixels = dicom::pixeldata::encaspulation::encapsulate(&mut frames, 1)?;
dcm.put(DataElement::new(dicom::dictionary_std::tags::PIXEL_DATA, dicom::core::VR::OB, encapsulated_pixels);
```


dougyau:
I have done a refactor, to allow different kind of scenarios.

In the first scenario, you could use the simplest call. The bad part of this is that at some point, the input frames and the output frames are allocated.
```
let frames: Vec<Vec<u8>> = ...

let encapsulated_pixels = dicom::pixeldata::encapsulation::encapsulate(&mut frames, 1);
dcm.put(DataElement::new(dicom::dictionary_std::tags::PIXEL_DATA, dicom::core::VR::OB, encapsulated_pixels);
```

Now if memory is an issue (mostly when dealing with multiframe images) the frames can be processed individually.
```
let frames: Vec<Vec<u8>> = ...
let encapsulated_pixels = frames.
    .into_iter()
    .map(|frame| ...do some encoding)
    .for_each(|encoded_frame| encapsulated_pixels.add_frame(encoded_frame));

dcm.put(DataElement::new(dicom::dictionary_std::tags::PIXEL_DATA, dicom::core::VR::OB, encapsulated_pixels);
```

And the last way is useful when using rayon
```
let frames: Vec<Vec<u8>> = ...
let fragments = frames.
    .into_par_iter()
    .map(|frame| ...do some encoding)
    .flat_map(|encoded_frame| dicom::pixeldata::encapsulation::fragment_frame(encoded_frame, 1))
    .collect::<Vec<Vec<u8>>>;
let encapsulated_pixels = dicom::pixeldata::encapsulation::EncapsulatedPixels::from(fragments);

dcm.put(DataElement::new(dicom::dictionary_std::tags::PIXEL_DATA, dicom::core::VR::OB, encapsulated_pixels);
```

pixeldata/src/encapsulation.rs
```diff
@@ -0,0 +1,207 @@
...
+    D: DataDictionary,
+    D: Clone,
+{
+    fn into(self) -> Value<InMemDicomObject<D>, InMemFragment> {
+        Value::PixelSequence {
+            offset_table: self.bot,
+            fragments: self.fragments,
+        }
+    }
+}
```
Enet4, line 54: This is not a good idea. The general recommendation is to implement `From` instead.

```suggestion
impl<D> From<EncapsulatedPixels> for Value<InMemDicomObject<D>, InMemFragment>
where
    D: DataDictionary,
    D: Clone,
{
    fn from(value: EncapsulatedPixels) -> Self {
        Value::PixelSequence {
            offset_table: value.bot,
            fragments: value.fragments,
        }
    }
}
```

pixeldata/src/encapsulation.rs
```diff
@@ -0,0 +1,207 @@
...
+use dicom_core::DataDictionary;
+use dicom_object::mem::InMemFragment;
+use dicom_object::InMemDicomObject;
+
+#[derive(Debug)]
+pub struct EncapsulatedPixels {
+    bot: C<u32>,
+    current_offset: u32,
+    fragments: C<Vec<u8>>,
+    number_of_fragments: u32,
```
Enet4, line 11: It is not clear what this field is for. Is it the number of fragments per frame? I feel that the frame splitting criterion should rather be based on a maximum fragment size in bytes.

Enet4 (changes requested):
Thank you for continuing this contribution. I like the idea of introducing a more substantial API! But I also feel that there is a lot to improve and fix before it is ready. Please attend to the comments inline.

pixeldata/src/encapsulation.rs
dougyau: Great suggestion, I have implemented the suggested change.

pixeldata/src/encapsulation.rs
```diff
@@ -0,0 +1,207 @@
...
+{
+    fn into(self) -> Value<InMemDicomObject<D>, InMemFragment> {
+        Value::PixelSequence {
+            offset_table: self.bot,
+            fragments: self.fragments,
+        }
+    }
+}
+
+impl From<Vec<Vec<u8>>> for EncapsulatedPixels {
```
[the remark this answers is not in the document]
dougyau: Renamed the function as `from_fragments` as it is intended to import a list of fragments, like when you generate the fragments when using `fragment_frame` as part of an iterator

```
let frames: Vec<Vec<u8>> = ...
let fragments = frames.
    .into_par_iter()
    .map(|frame| ...do some encoding)
    .flat_map(|encoded_frame| dicom::pixeldata::encapsulation::fragment_frame(encoded_frame, 0))
    .collect::<Vec<Vec<u8>>>;
let encapsulated_pixels = dicom::pixeldata::encapsulation::EncapsulatedPixels::from_fragments(fragments);

dcm.put(DataElement::new(dicom::dictionary_std::tags::PIXEL_DATA, dicom::core::VR::OB, encapsulated_pixels);
```

dougyau:
@Enet4 I have improved the code based on the suggestions. Please let me know if you see more things that can be improved.

pixeldata/src/encapsulation.rs
```diff
@@ -0,0 +1,347 @@
...
+///     .collect();
+///
+/// let encapsulated_pixels = EncapsulatedPixels::from_frame_fragments(encapsulated_data)
+///     .expect("For multi frame data only 1 fragment per frame is allowed");
+///
+/// # let mut dcm = dicom_object::InMemDicomObject::new_empty();
+/// dcm.put(DataElement::new(PIXEL_DATA, OB, encapsulated_pixels));
+/// ```
+#[derive(Debug, Default)]
+pub struct EncapsulatedPixels {
```
Enet4, line 90: The upstream version (`master` branch) of `dicom_core` now features a pixel data fragment sequence type `dicom_core::value::PixelFragmentSequence`, so we want to build up our frames-to-fragments abstraction from this. `EncapsulatedPixels` would be removed and `PixelFragmentSequence` would be given an impl `From<FrameFragments>`.

If there are any good methods that would be important to have in `PixelFragmentSequence`, it should be possible to accommodate them there as well.

Enet4 (commented):
Hey @dougyau! I have been doing some changes to most of the crates in the DICOM-rs project. In particular, I felt that we would be better served with the pixel data sequence type being declared in the core crate itself, which may subvert some of the decisions made here.
Nevertheless, it is still important and useful to have some mechanisms to turn frames into fragments and vice versa (which we do not have yet), so I do not want to put this pull request to waste! :) Please see the inline comment and let me know if you have any other plans here.

core/src/value/fragments.rs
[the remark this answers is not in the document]
Enet4: Sounds like an interesting approach! It appears that the example creates a vector of `Fragments`. Is this for the case where a frame is composed of multiple fragments? Would we always have one `Fragments` per frame in 1:1 transfer syntaxes such as in the JPEG family of TSes?

I would strongly recommend that `Fragments` is well documented so that it is clear when its use is most fitting. An example as a doc-test would be awesome as well.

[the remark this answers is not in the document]
dougyau: Yes, this approach is to cover the case when a single frame DICOM is composed by multiple fragments, and for multi-frame files it is validated that a single fragment per frame is used.

@Enet4 what do you think, do I leave this part of the code on dicom_core and leave on pixeldata only the high level calls add_frame and encapsulate (under the hood the part that on dicom_core would be used)?

dougyau:
@Enet4 I have updated the tests and the doc blocks on the PR.

pixeldata/src/encapsulation.rs
```diff
@@ -0,0 +1,80 @@
+//! DICOM Pixel encapsulation
+//!
+//! This module implements encapsulation for pixel data.
+use dicom_core::value::fragments::Fragments;
+use dicom_core::value::Value;
+
+/// Encapsulate the pixel data of the frames. If fragment_size > 0 it will use 1 fragment per frame.
+/// This parameter is ignored for multi frame data, as 1 fragment per frame is required.
```
ingwinlu, line 8: isn't that depended on the transfer syntax used?
https://dicom.nema.org/dicom/2013/output/chtml/part05/sect_A.4.html

Enet4 (commented):
I made a few more suggestions. Primarily, I think it makes sense to split `encapsulate` for the two key scenarios (unless there is a reason I haven't grasped yet). `encapsulate_single_frame` would receive a single frame and a maximum fragment size, whereas the other `encapsulate_frames` would give each frame its own fragment and provide a basic offset table accordingly.

pixeldata/src/encapsulation.rs
dougyau: I think it is better to have 1 function for this, otherwise when you intend to encapsulate you need to verify if it is single or multi frame, then call the corresponding function.

dougyau:
The changes have been implemented.

pixeldata/src/encapsulation.rs
```diff
@@ -1,80 +1,90 @@
...
+//! This module implements encapsulation for pixel data.
+use dicom_core::value::fragments::Fragments;
+use dicom_core::value::Value;
+use std::vec;
+
+/// Encapsulate the pixel data of a list of frames.
+///
+/// Check [Fragments] in case more control over the processing is required.
+///
+/// #Example
```
Enet4, line 12: This needs a space, otherwise it is not rendered as a heading.
```suggestion
/// # Example
```

pixeldata/src/encapsulation.rs
```diff
@@ -1,80 +1,90 @@
...
+        .map(|frame| Fragments::new(frame, 0))
+        .collect::<Vec<Fragments>>();
+
+    Value::PixelSequence(fragments.into())
+}
+
+/// Encapsulate the pixel data of a single frame. If `fragment_size` is zero then `frame.len()` will
+/// be used instead.
+///
+/// #Example
```
Enet4, line 38:
```suggestion
/// # Example
```

Enet4 (approved):
OK, we are in condition to bring this upstream. Thank you @dougyau for your commitment.

PR #313, opened 2022-11-26
Add SopClassExtendedNegotiationSubItem for Pdu

jmlaka:
This adds a `SopClassExtendedNegotiationSubItem` to the Pdu's `UserVariableItem `enum.
It will be useful for extended negotiation options between AEs.

Fixes test for PDU UserVariables that never failed.

### DIFF
--- a/ul/src/pdu/mod.rs
+++ b/ul/src/pdu/mod.rs
@@ -325,6 +325,7 @@ pub enum UserVariableItem {
     MaxLength(u32),
     ImplementationClassUID(String),
     ImplementationVersionName(String),
+    SopClassExtendedNegotiationSubItem(String, Vec<u8>)
 }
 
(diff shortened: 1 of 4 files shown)


Enet4 (approved):
I had the opportunity to review this, and it appears to be an accurate translation of the SOP Class Extended Negotiation Sub-Item as per [PS3.7 D.3.3.5](https://dicom.nema.org/medical/dicom/2022c/output/chtml/part07/sect_D.3.3.5.html). It's fine to enter by me! Thank you!

Considering that there are other kinds of sub-items not yet specified here, I also considered marking `PduVariableItem` as `non_exhaustive`, but this would only hide the breaking change of turning unknown item structures into future definitions, so we are in position to introduce more semver breaking changes here until all known sub-items are defined and supported.

PR #326, opened 2023-01-28
Add attribute operations API

Enet4:
This brings a new common API for applying specific changes to attributes in a DICOM object or object-like type. This has been implemented for all DICOM object types, namely `InMemDicomObject`, `FileDicomObject` and `FileMetaTable`.

Example of use:

```rust
use dicom_encoding::ops::{ApplyOp, AttributeAction, AttributeOp};

// given some DICOM object
let mut obj = open_file("I_00001.dcm")?;

// apply patient name change
obj.apply(AttributeOp {
  tag: tags::PATIENT_NAME,
  action: AttributeAction::ReplaceStr("Patient^Anonymous".into())
})?;
assert_eq!(
    obj.get(tags::PATIENT_NAME).unwrap().value().to_str().unwrap(),
    "Patient^Anonymous",
);
```

### Summary

- [core] Add `DataElement::into_parts`
- [core] Add more `PrimitiveValue::extend_*` methods for other data types
- [encoding] New attribute operations module `ops`, with `AttributeOp`, `AttributeAction`, and `ApplyOp`
- [object] impl `ApplyOp` for current object types
- [object] add infallible methods `get` and `take` for `InMemDicomObject`, return an option instead of a result.

### DIFF
--- a/core/src/value/primitive.rs
+++ b/core/src/value/primitive.rs
@@ -3662,7 +3662,7 @@ impl PrimitiveValue {
     /// use dicom_core::dicom_value;
     ///
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
-    /// let mut value = dicom_value!(U16, [1, 2]);
+    /// let mut value = dicom_value!(I16, [1, 2]);
     /// value.extend_i16([-5])?;
     /// assert_eq!(value.to_multi_int::<i16>()?, vec![1, 2, -5]);
     ///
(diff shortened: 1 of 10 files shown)


core/src/value/primitive.rs
```diff
@@ -3729,6 +3945,215 @@ impl PrimitiveValue {
...
+                elements.extend(numbers.into_iter().map(|n| n as f64));
+                Ok(())
+            }
+            PrimitiveValue::Tags(_)
+            | PrimitiveValue::Date(_)
+            | PrimitiveValue::DateTime(_)
+            | PrimitiveValue::Time(_) => Err(IncompatibleNumberTypeSnafu {
+                original: self.value_type(),
+            }
+            .build()),
```
Enet4, line 4055:
```suggestion
            | PrimitiveValue::Time(_) => IncompatibleNumberTypeSnafu {
                original: self.value_type(),
            }
            .fail(),
```

Issue #333, opened 2023-03-27
Error while converting decimal string(DS) to float64


lalberge:
Hi,

We found a bug while converting decimal strings to float64, it happened while reading the RescaleIntercept attribute with the value
 " -1024".
As stated in the DICOM norm, decimal strings should ignore trimming and **leading** spaces:
"A string of characters representing either a fixed point number or a floating point number. A fixed point number shall contain only the characters 0-9 with an optional leading "+" or "-" and an optional "." to mark the decimal point. A floating point number shall be conveyed as defined in ANSI X3.9, with an "E" or "e" to indicate the start of the exponent. Decimal Strings may be padded with leading or trailing spaces. Embedded spaces are not allowed."

The bug could be easily fixed by adding trim `.trim_start()` before casting to float64. 
https://github.com/Enet4/dicom-rs/blob/c1644cf355f160946f947273bb89283546be2c98/core/src/value/primitive.rs#L1654

I would be happy to contribute to that fix if you accept external contributions.

Enet4:
@lalberge Hello! Thank you for reaching out with this bug! It may indeed be an oversight on what is allowed in values with the DS and IS value representation.
I would gladly take a pull request for this (see also the [contribution guidelines](https://github.com/Enet4/dicom-rs/blob/master/CONTRIBUTING.md)).

lalberge:
@Enet4 
Hello. 
I have implemented the bug fix and its corresponding test cases.
The branch is ready locally, but I am not allowed to push it and create the associated PR.
I could not find in the contributing readme how to do that :) 

Léo

Enet4:
Hello @lalberge! You can create your own fork here on GitHub and [create a pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork) from it. See the link for more details and let me know if you need more assistance.

lalberge:
Hi @Enet4,

The PR is ready to be reviewed here: https://github.com/Enet4/dicom-rs/pull/335

PR #335, opened 2023-03-31
Leading whitespaces in VR DS and IS  

lalberge:
Bug:
Leading whitespaces in DS or IS would cause errors in conversion from a string primitive. 
The issue is described here: https://github.com/Enet4/dicom-rs/issues/333

Cause:
Strings were not trimmed with leading spaces.

Fix:
Added `.trim_start()` before conversion to int or float.

Verification:
Added unit tests to ensure theses cases are now adequately handled.

### DIFF
--- a/core/src/value/primitive.rs
+++ b/core/src/value/primitive.rs
@@ -964,8 +964,7 @@ impl PrimitiveValue {
     {
         match self {
             PrimitiveValue::Str(s) => {
-                s.trim_end()
-                    .trim_start()
+                s.trim()
                     .parse()
                     .context(ParseIntegerSnafu)
                     .map_err(|err| ConvertValueError {
@@ -975,8 +974,7 @@ impl PrimitiveValue {
                     })
             }
             PrimitiveValue::Strs(s) if !s.is_empty() => s[0]
-                .trim_end()
-                .trim_start()
+                .trim()
                 .parse()
                 .context(ParseIntegerSnafu)
                 .map_err(|err| ConvertValueError {
@@ -1305,8 +1303,7 @@ impl PrimitiveValue {
     pub fn to_float32(&self) -> Result<f32, ConvertValueError> {
         match self {
             PrimitiveValue::Str(s) => {
-                s.trim_end()
-                    .trim_start()
+                s.trim()
                     .parse()
                     .context(ParseFloatSnafu)
                     .map_err(|err| ConvertValueError {
@@ -1316,8 +1313,7 @@ impl PrimitiveValue {
                     })
             }
             PrimitiveValue::Strs(s) if !s.is_empty() => s[0]
-                .trim_end()
-                .trim_start()
+                .trim()
                 .parse()
                 .context(ParseFloatSnafu)
                 .map_err(|err| ConvertValueError {
@@ -1657,8 +1653,7 @@ impl PrimitiveValue {
     pub fn to_float64(&self) -> Result<f64, ConvertValueError> {
         match self {
             PrimitiveValue::Str(s) => {
-                s.trim_end()
-                    .trim_start()
+                s.trim()
                     .parse()
                     .context(ParseFloatSnafu)
                     .map_err(|err| ConvertValueError {
@@ -1668,8 +1663,7 @@ impl PrimitiveValue {
                     })
             }
             PrimitiveValue::Strs(s) if !s.is_empty() => s[0]
-                .trim_end()
-                .trim_start()
+                .trim()
                 .parse()
                 .context(ParseFloatSnafu)
                 .map_err(|err| ConvertValueError {
--- a/core/src/value/primitive.rs
+++ b/core/src/value/primitive.rs
@@ -965,6 +965,7 @@ impl PrimitiveValue {
         match self {
             PrimitiveValue::Str(s) => {
                 s.trim_end()
+                    .trim_start()
                     .parse()
                     .context(ParseIntegerSnafu)
                     .map_err(|err| ConvertValueError {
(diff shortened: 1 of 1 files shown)


core/src/value/primitive.rs
```diff
@@ -965,6 +965,7 @@ impl PrimitiveValue {
         match self {
             PrimitiveValue::Str(s) => {
                 s.trim_end()
+                    .trim_start()
```
ingwinlu, line 968: is there a reason to not go with [trim](https://doc.rust-lang.org/std/string/struct.String.html#method.trim) everywhere?

lalberge: Hi @ingwinlu ,
DICOM VR definition"Decimal Strings may be padded with leading or trailing spaces. Embedded spaces are not allowed." 
So it depends if you want to raise an error with embedded spaces or not?

lalberge: Updated the code 👍

Enet4:
Looks good! We are only missing some updates to the documentation, since it still indicates that it only trims trailing whitespace. Please see lines 926, 1085, 1274, 1434, 1783 and update them accordingly. It also seems that one of the methods was missing a similar sentence, which would have been in line 1623. We can fix that too while we're at it.

I would go with something like this:

```diff
--- a/core/src/value/primitive.rs
+++ b/core/src/value/primitive.rs
@@ -923,8 +923,7 @@ impl PrimitiveValue {
     /// If the value is a string or sequence of strings,
     /// the first string is parsed to obtain an integer,
     /// potentially failing if the string does not represent a valid integer.
-    /// The string is stripped of trailing whitespace before parsing,
-    /// in order to account for the possible padding to even length.
+    /// The string is stripped of leading/trailing whitespace before parsing.
     /// If the value is a sequence of U8 bytes,
     /// the bytes are individually interpreted as independent numbers.
```

lalberge:
> Looks good! We are only missing some updates to the documentation, since it still indicates that it only trims trailing whitespace. Please see lines 926, 1085, 1274, 1434, 1783 and update them accordingly. It also seems that one of the methods was missing a similar sentence, which would have been in line 1623. We can fix that too while we're at it.
> 
> I would go with something like this:
> 
> ```diff
> --- a/core/src/value/primitive.rs
> +++ b/core/src/value/primitive.rs
> @@ -923,8 +923,7 @@ impl PrimitiveValue {
>      /// If the value is a string or sequence of strings,
>      /// the first string is parsed to obtain an integer,
>      /// potentially failing if the string does not represent a valid integer.
> -    /// The string is stripped of trailing whitespace before parsing,
> -    /// in order to account for the possible padding to even length.
> +    /// The string is stripped of leading/trailing whitespace before parsing.
>      /// If the value is a sequence of U8 bytes,
>      /// the bytes are individually interpreted as independent numbers.
> ```

Done 👍

core/src/value/primitive.rs
```diff
@@ -923,7 +923,7 @@ impl PrimitiveValue {
     /// If the value is a string or sequence of strings,
     /// the first string is parsed to obtain an integer,
     /// potentially failing if the string does not represent a valid integer.
-    /// The string is stripped of trailing whitespace before parsing,
+    /// The string is stripped of leading/trailing whitespace before parsing,
     /// in order to account for the possible padding to even length.
```
Enet4, line 927: The idea of the subsequent line is not 100% true anymore, since the value may just have many leading and trailing spaces for whatever reasons, so I suggest we remove it. Same thing for the other occurrences.

```suggestion
    /// The string is stripped of leading/trailing whitespace before parsing.
```

lalberge: fixed it. It was only for this block or the others as well?

core/src/value/primitive.rs
```diff
@@ -1083,7 +1082,7 @@ impl PrimitiveValue {
     /// If the value is a string or sequence of strings,
     /// each string is parsed to obtain an integer,
     /// potentially failing if the string does not represent a valid integer.
-    /// The string is stripped of trailing whitespace before parsing,
+    /// The string is stripped of leading/trailing whitespace before parsing,
     /// in order to account for the possible padding to even length.
```
Enet4, line 1086:
```suggestion
    /// The string is stripped of leading/trailing whitespace before parsing.
```

core/src/value/primitive.rs
```diff
@@ -1273,7 +1272,7 @@ impl PrimitiveValue {
     /// If the value is a string or sequence of strings,
     /// the first string is parsed to obtain a number,
     /// potentially failing if the string does not represent a valid number.
-    /// The string is stripped of trailing whitespace before parsing,
+    /// The string is stripped of leading/trailing whitespace before parsing,
     /// in order to account for the possible padding to even length.
```
Enet4, line 1276:
```suggestion
    /// The string is stripped of leading/trailing whitespace before parsing.
```

core/src/value/primitive.rs
```diff
@@ -1434,7 +1433,7 @@ impl PrimitiveValue {
     /// If the value is a string or sequence of strings,
     /// the strings are parsed to obtain a number,
     /// potentially failing if the string does not represent a valid number.
-    /// The string is stripped of trailing whitespace before parsing,
+    /// The string is stripped of leading/trailing whitespace before parsing,
     /// in order to account for the possible padding to even length.
```
Enet4, line 1437:
```suggestion
    /// The string is stripped of leading/trailing whitespace before parsing.
```

core/src/value/primitive.rs
```diff
@@ -1784,7 +1784,7 @@ impl PrimitiveValue {
...
```
Enet4, line 1788:
```suggestion
    /// The string is stripped of leading/trailing whitespace before parsing.
```

lalberge:
Let me know if there is anything more to do :)

Enet4 (approved):
Seems to be in order now. Thank you for your contribution! :+1:

PR #364, opened 2023-06-11
InMemDicomObject Length::UNDEFINED after modify

jmlaka:
As an `InMemDicomObject ` can have a certain defined `Length` after parsing dicom data, 
it is necessary to reset or recalculate this Length after calling any methods that modify
the object's inner entries value e.g. `.put()` `.remove_element() `...

The Length is set to `Length::UNDEFINED` as recalculating it might be expensive.

### DIFF
--- a/object/src/mem.rs
+++ b/object/src/mem.rs
@@ -612,18 +612,21 @@ where
     /// Insert a data element to the object, replacing (and returning) any
     /// previous element of the same attribute.
     pub fn put(&mut self, elt: InMemElement<D>) -> Option<InMemElement<D>> {
+        self.len = Length::UNDEFINED;
         self.put_element(elt)
     }
 
     /// Insert a data element to the object, replacing (and returning) any
     /// previous element of the same attribute.
     pub fn put_element(&mut self, elt: InMemElement<D>) -> Option<InMemElement<D>> {
+        self.len = Length::UNDEFINED;
         self.entries.insert(elt.tag(), elt)
     }
 
     /// Remove a DICOM element by its tag,
     /// reporting whether it was present.
     pub fn remove_element(&mut self, tag: Tag) -> bool {
+        self.len = Length::UNDEFINED;
         self.entries.remove(&tag).is_some()
     }
 
@@ -631,11 +634,13 @@ where
     /// reporting whether it was present.
     pub fn remove_element_by_name(&mut self, name: &str) -> Result<bool, AccessByNameError> {
         let tag = self.lookup_name(name)?;
+        self.len = Length::UNDEFINED;
         Ok(self.entries.remove(&tag).is_some())
     }
 
     /// Remove and return a particular DICOM element by its tag.
     pub fn take_element(&mut self, tag: Tag) -> Result<InMemElement<D>> {
+        self.len = Length::UNDEFINED;
         self.entries
             .remove(&tag)
             .context(NoSuchDataElementTagSnafu { tag })
@@ -645,6 +650,7 @@ where
     /// if it is present,
     /// returns `None` otherwise.
     pub fn take(&mut self, tag: Tag) -> Option<InMemElement<D>> {
+        self.len = Length::UNDEFINED;
         self.entries.remove(&tag)
     }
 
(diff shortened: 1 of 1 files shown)


object/src/mem.rs
```diff
@@ -612,30 +612,35 @@ where
...
     pub fn put_element(&mut self, elt: InMemElement<D>) -> Option<InMemElement<D>> {
+        self.len = Length::UNDEFINED;
         self.entries.insert(elt.tag(), elt)
     }
 
     /// Remove a DICOM element by its tag,
     /// reporting whether it was present.
     pub fn remove_element(&mut self, tag: Tag) -> bool {
+        self.len = Length::UNDEFINED;
         self.entries.remove(&tag).is_some()
```
Enet4, line 630:
```suggestion
        if self.entries.remove(&tag).is_some() {
            self.len = Length::UNDEFINED;
            true
        } else {
            false
        }
```

object/src/mem.rs
```diff
@@ -612,30 +612,35 @@ where
...
     }
 
     /// Remove a DICOM element by its keyword,
     /// reporting whether it was present.
     pub fn remove_element_by_name(&mut self, name: &str) -> Result<bool, AccessByNameError> {
         let tag = self.lookup_name(name)?;
+        self.len = Length::UNDEFINED;
         Ok(self.entries.remove(&tag).is_some())
```
Enet4, line 638:
```suggestion
        Ok(self.entries.remove(&tag).is_some())
            .map(|removed| {
                if removed {
                    self.len = Length::UNDEFINED;
                }
                removed
            })
```

Enet4 (commented):
Thank you for working on this.

Could you adjust some of these methods so that the length is not reset when no changes were truly made? Not that it's a huge problem, but it feels surprising for a method representing an operation as simple as removing an attribute to return an error and still make changes to the underlying object. I left some examples, it should not be hard to do the same for the other methods.

object/src/mem.rs
```diff
@@ -728,8 +745,8 @@ where
                     let vr = e.vr();
                     // replace element
                     *e = DataElement::empty(tag, vr);
+                    self.len = Length::UNDEFINED;
```
Enet4, line 748: If I see this correctly, actions `SetStr`, `Set`, and the various `Push*` might also modify data without calling `put` (see e.g. line 811), so we'd also need some length resets in `apply_push_*_impl` and `apply_change_value_impl`.

jmlaka:
I just removed some duplicate code.

The methods you mentioned did not need changes, as they internally call `self.put()`  - which is already handled.

jmlaka:
Still wrong :) Will be back

object/src/mem.rs
```diff
@@ -618,34 +618,55 @@ where
     /// Insert a data element to the object, replacing (and returning) any
     /// previous element of the same attribute.
     pub fn put_element(&mut self, elt: InMemElement<D>) -> Option<InMemElement<D>> {
-        self.entries.insert(elt.tag(), elt)
+        self.len = Length::UNDEFINED;
```
Enet4, line 621: I suppose that this is already done inside the `map`, right?
```suggestion
```

Enet4 (commented):
Great! Only one minor concern left.

object/src/mem.rs
jmlaka: right ...

object/src/mem.rs
```diff
@@ -618,34 +618,54 @@ where
     /// Insert a data element to the object, replacing (and returning) any
     /// previous element of the same attribute.
     pub fn put_element(&mut self, elt: InMemElement<D>) -> Option<InMemElement<D>> {
-        self.entries.insert(elt.tag(), elt)
+        self.entries.insert(elt.tag(), elt).map(|e| {
+            self.len = Length::UNDEFINED;
+            e
+        })
```
Enet4, line 624: Ah, so this is why I had a weird hunch about this. \*headscratch\* `BTreeMap::insert` only returns `Some` if the element already existed, but we want the length to be reset regardless of this.
Let's go with this instead.

```suggestion
        let o = self.entries.insert(elt.tag(), elt);
        self.len = Length::UNDEFINED;
        o
```

Enet4 (changes requested):
The test failed. See the suggestion inline for a possible solution.

Enet4 (approved):
Awesome! Thank you again, @jmlaka!

PR #369, opened 2023-06-15
Add dicom-dump to parent crate & extend documentation

### LINKED ISSUE #368: dicom_dump not visible
The lib part of the dicom_dump is not visible when importing dicom-rs and so is not usable from external projects.

Can this be fixed by updating the `parent` crate ?

Enet4:
### Summary

- add `dicom-dump` to parent crate
- Extend root crate documentation of `dicom` to better clarify how to use it.

Resolves #368.

### DIFF
--- a/parent/src/lib.rs
+++ b/parent/src/lib.rs
@@ -4,6 +4,7 @@
 //! the [`object`](crate::object) module.
 pub use dicom_core as core;
 pub use dicom_dictionary_std as dictionary_std;
+pub use dicom_dump as dump;
 pub use dicom_encoding as encoding;
 pub use dicom_object as object;
 pub use dicom_parser as parser;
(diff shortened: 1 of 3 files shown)


parent/src/lib.rs
```diff
@@ -1,9 +1,68 @@
...
-//! the [`object`](crate::object) module.
+//! [ts]: dicom_encoding::transfer_syntax
+//!
+//! ## Advanced
+//!
+//! - To write DICOM network application entity software,
+//!   see the [`ul`] module for PDU reading/writing
+//!   and a DICOM association API.
+//! - If you are writing or declaring your own transfer syntax,
+//!   you will need to take [`encoding`] module
```
Enet4, line 55:
```suggestion
//!   you will need to take the [`encoding`] module
```

PR #370, opened 2023-06-15
Support private creator element detection in dictionary

Enet4:
This adds `TagRange::PrivateCreator` so that private creator data elements are detected as such.

### DIFF
--- a/core/src/dictionary/mod.rs
+++ b/core/src/dictionary/mod.rs
@@ -139,7 +139,7 @@ impl FromStr for TagRange {
  * The methods herein have no generic parameters, so as to enable being
  * used as a trait object.
  */
-pub trait DataDictionary: Debug {
+ pub trait DataDictionary {
     /// The type of the dictionary entry.
     type Entry: DictionaryEntry;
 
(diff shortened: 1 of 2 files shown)


core/src/dictionary/mod.rs
```diff
@@ -17,7 +17,8 @@ use std::str::FromStr;
 /// For example,
 /// _Overlay Data_ (60xx,3000) has more than one possible tag,
 /// since it is part of a repeating group.
-/// Moreover, a unique variant is defined for group length tags.
+/// Moreover, a unique variant is defined for group length tags
+/// and for private creator tags.
```
Enet4, line 21:
```suggestion
/// and another one for private creator tags.
```

core/src/dictionary/mod.rs
```diff
@@ -139,7 +147,7 @@ impl FromStr for TagRange {
  * The methods herein have no generic parameters, so as to enable being
  * used as a trait object.
  */
-pub trait DataDictionary: Debug {
+ pub trait DataDictionary {
```
Enet4, line 150:
```suggestion
pub trait DataDictionary {
```

PR #381, opened 2023-06-26
[findscu] Add support for modality worklist information model

Enet4:
### Summary 

- SOP class _Modality Worklist Information Model (Find)_ can be enabled via `--mwl` option (or `-W`)
- replace some UID string literals with use of constants from dicom_dictionary_std


bastiao:
Tested with "Modality Worklist Information Model - FIND" and it seems to work fine!

PR #388, opened 2023-07-08
[findscu] Support combining query file with query options

Enet4:
This allows `dicom-findscu` users to pass both a query file to serve as a base or template, as well as query options to extend it.

### Summary

- change parse_queries to take a base object
- tweak build_query so that both forms of query specification may take place, but still error if neither one is provided
- update readme accordingly


bastiao:
Nice one also! Works fine.
It actually helps because most of the times we have different templates of return keys that we would like to retrieve.
As you implicit mention, for generate the dcm file as query, dump2dcm from dcmtk can help.

PR #399, opened 2023-07-25
[Fix] BOT generation for encapsulated data

dougyau:
The BOT of the pixel data should point to the beginning of each frame. The original implementation was adding 1 extra index which was wrong as the last entry pointed to the end of the file.

### DIFF
--- a/core/src/value/fragments.rs
+++ b/core/src/value/fragments.rs
@@ -107,6 +107,8 @@ impl From<Vec<Fragments>> for PixelFragmentSequence<InMemFragment> {
             fragments.append(&mut frame.fragments);
         }
 
+        offset_table.resize(offset_table.len() - 1, 0);
+
         PixelFragmentSequence {
             offset_table,
             fragments: C::from_vec(fragments),
@@ -117,6 +119,7 @@ impl From<Vec<Fragments>> for PixelFragmentSequence<InMemFragment> {
 #[cfg(test)]
 mod tests {
     use crate::value::fragments::Fragments;
+    use crate::value::{InMemFragment, PixelFragmentSequence};
 
     #[test]
     fn test_fragment_frame() {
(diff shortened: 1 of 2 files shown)


core/src/value/fragments.rs
```diff
@@ -175,4 +179,32 @@ mod tests {
...
+        let fragment_sequence: PixelFragmentSequence<InMemFragment> = data.into();
+        assert_eq!(fragment_sequence.offset_table.len(), 1);
+        assert_eq!(fragment_sequence.offset_table[0], 0);
+    }
+
+    #[test]
+    fn test_bot_multi_frame_generation() {
+        let data = vec![
+            Fragments::new(vec![0u8, 0u8, 0u8, 0u8], 0),
+            Fragments::new(vec![1u8, 1u8, 1u8, 1u8, 1u8, 1u8], 0),
```
Enet4, line 203: This might be easier to read. The other `vec!` calls may be refactored as well.

```suggestion
            Fragments::new(vec![0u8; 4], 0),
            Fragments::new(vec![1u8; 6], 0),
```

Enet4 (commented):
Looks spot on! Thank you for noticing! I will just suggest a minor tweak to the vector construction calls and we can merge afterwards.

dougyau:
The issue was kind of tricky to spot, the generated files could be open with Weasis, Aliza, etc. DCMTK did not show any issues nor GDCM. It was cornerstone dicom-parser which failed due to this issue.

Enet4:
> The issue was kind of tricky to spot, the generated files could be open with Weasis, Aliza, etc. DCMTK did not show any issues nor GDCM. It was cornerstone dicom-parser which failed due to this issue.

That is interesting. Indeed, any parser relying uniquely on the basic offset table to identify the number of frames of an image could choke on this.

In any case, thank you very much @dougyau. I will try to push a patch release soon.

Enet4 (approved):

PR #405, opened 2023-08-23
Check size fix in detect_preamble

### LINKED ISSUE #404: detect_preamble and in extension from_reader_with_all_options doesn't check for size
In detect_preamble in mem.rs

```rust
        if &buf[0..4] == b"DICM" {
            return Ok(ReadPreamble::Never);
        }
```
is done without checking if `buf` actually has at least 4 bytes, leading to panic when, for example, trying to read empty files with `from_reader_with_all_options` and friends.

Should return with io::Error.

DerOrfa:
Check for input size in detect_preamble. Fixes #404.

### DIFF
--- a/object/src/mem.rs
+++ b/object/src/mem.rs
@@ -322,8 +322,13 @@ where
         S: Read,
     {
         let buf = reader.fill_buf()?;
+        let buflen= buf.len();
 
-        if buf.len() >= 132 && &buf[128..132] == b"DICM" {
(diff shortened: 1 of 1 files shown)


Enet4 (approved):
Indeed, it appears to be in order. The code changes are sound and testing it in practice shows that it fixes the panic on empty and close-to-empty files.

Thank you very much @DerOrfa!

PR #414, opened 2023-09-02
Fix padding skipping in multiframe JPEG decoding

### LINKED ISSUE #413: First two bytes are not an SOI marker error
Hi @Enet4 , 
I am new to Rust (from python) and was just trying to do some basic image loading tasks with this library. However, I keep getting the above error, I did try your sample code from the dicom-rs documentation. But I am not sure, can you please let me know what I am missing?

```rust
use dicom::object::open_file;
use dicom_pixeldata::PixelDecoder;

fn main() {

    let path = "../0003.dcm";
    let _obj = open_file(&path).unwrap();
    let data = _obj.decode_pixel_data().unwrap();

    println!("{:?}", data)
}
```

Error: 
```rust    
Finished dev [unoptimized + debuginfo] target(s) in 0.36s
Running `target/debug/dicom_testing`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error(DecodePixelData { source: Custom { message: "JPEG decoder failure", source: Some(Format("first two bytes are not an SOI marker")) } })', src/main.rs:24:41
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Thanks!

Enet4:
Resolves #413, so that it can decode JPEG baseline with odd length fragments.

An alternative to trying to patch the `decode` implementation is to remove it entirely and let the trait's default implementation and `decode_frame` take its place.

### DIFF
--- a/encoding/src/adapters.rs
+++ b/encoding/src/adapters.rs
@@ -13,7 +13,7 @@
 //! to be able to decode and encode imaging data, respectively.
 
 use dicom_core::{ops::AttributeOp, value::C};
-use snafu::{OptionExt, Snafu};
+use snafu::Snafu;
 use std::borrow::Cow;
 
 /// The possible error conditions when decoding (reading) pixel data.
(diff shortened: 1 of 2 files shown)


Enet4:
Indeed I found issues in `JpegAdapter`, both trough the complete `decode` method and the per-frame method `decode_frame`. I believe that they have been fixed, and turning RUBO's `0003.dcm` into images should work now.

PR #422, opened 2023-10-07
Decode single frame using GDCM

dougyau:
Extending from the PR #421 when using GDCM to decode images it will load the whole image to memory. This PR implements the logic required to load only one image.

### DIFF
--- a/pixeldata/src/gdcm.rs
+++ b/pixeldata/src/gdcm.rs
@@ -1,6 +1,7 @@
 //! Decode pixel data using GDCM when the default features are enabled.
 
 use crate::*;
+use dicom_dictionary_std::tags;
 use dicom_encoding::adapters::DecodeError;
 use dicom_encoding::transfer_syntax::TransferSyntaxIndex;
 use dicom_transfer_syntax_registry::TransferSyntaxRegistry;
@@ -18,7 +19,7 @@ where
         use super::attribute::*;
 
         let pixel_data = pixel_data(self).context(GetAttributeSnafu)?;
-        
+
         let cols = cols(self).context(GetAttributeSnafu)?;
         let rows = rows(self).context(GetAttributeSnafu)?;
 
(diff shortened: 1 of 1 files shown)


pixeldata/src/gdcm.rs
```diff
@@ -144,6 +145,164 @@ where
...
+                    message: source.to_string(),
+                    source: Some(Box::new(source)),
+                };
+
+                let frame = frame as usize;
+                let data = if number_of_frames == 1 && fragments.len() > 1 {
+                    let mut buf = Vec::new();
+                    fragments
+                        .into_iter()
+                        .for_each(|fragment| buf.append(&mut fragment.clone()));
```
Enet4, line 211: I think a copy isn't needed if you use `extend` instead.

```suggestion
                        .for_each(|fragment| buf.extend(&fragment));
```

pixeldata/src/gdcm.rs
```diff
@@ -144,6 +145,164 @@ where
...
+                    buf
+                } else {
+                    fragments[frame].to_vec()
```
Enet4, line 214: What if the frame comprises more than one fragment in a multi-frame pixel data fragment sequence? I do not think that we can make this assumption.

pixeldata/src/gdcm.rs
```diff
@@ -144,6 +145,164 @@ where
...
+            })
+        } else {
+            None
+        };
+
+        Ok(DecodedPixelData {
+            data: Cow::from(decoded_pixel_data),
+            cols: cols.into(),
+            rows: rows.into(),
+            number_of_frames,
```
Enet4, line 292: As we retrieve a single frame, the decoded pixel data value will feature only 1 frame. This has likewise to be reflected here.

```suggestion
            number_of_frames: 1,
```

pixeldata/src/gdcm.rs
```diff
@@ -144,6 +145,164 @@ where
...
+            }
+            Value::Primitive(p) => {
+                // Non-encoded, just return the pixel data of the first frame
+                p.to_bytes().to_vec()
+            }
+            Value::Sequence(_) => InvalidPixelDataSnafu.fail()?,
+        };
+
+        // pixels are already interpreted,
+        // set new photometric interpretation
```
Enet4, line 270: We will still need this, because the decoded pixel data does not necessarily follow the photometric interpretation originally described.

pixeldata/src/gdcm.rs
```diff
@@ -144,6 +145,164 @@ where
...
+                        interleaved.push(frame[i + frame_size * 2]);
+                    }
+                    interleaved
+                } else {
+                    frame
+                }
+            }
+            Value::Primitive(p) => {
+                // Non-encoded, just return the pixel data of the first frame
+                p.to_bytes().to_vec()
```
Enet4, line 264: We might need to adjust this as well.

Enet4 (commented):
Great follow-up on the new frame decoding feature, much appreciated!

I assume that this is still a work in progress, but I will leave some comments for now.

pixeldata/src/gdcm.rs
dougyau: This is a nice one, still I am trying to see if I can avoid the extra allocation, this one is so the first iteration works.

pixeldata/src/gdcm.rs
dougyau: Only single frame dicoms can have more than 1 fragment. Multi frame files, must be 1 fragment per frame. At least according to the standard.

dougyau:
So far, I have tested this with Compressed (jpeg, jpeg-ls and j2k) and Uncompressed (little endian) data. With RGB, YBR_FULL_422, MONOCHROME1, MONOCRHOME2 and PALETTE_COLOR (at least decompressing)

pixeldata/src/gdcm.rs
Enet4: I guess we can move forward with this assumption and revisit if we find a conflicting sample case. :+1:

Enet4:
Hello @dougyau! Are there any other pending concerns with this PR, so that it can be reviewed?

dougyau:
This code is working, what I have to do is cover it with unit testing. I have been short of time, but once I do have some time, I will add the tests.

dougyau:
I have validated the existing test, and it uses the new code path. Of the test cases I believe we could convert `SC_rgb_2frame.dcm` to have a planar configuration = 1 to make it complete. How can I add this to the test-files repo?

Enet4:
> I have validated the existing test, and it uses the new code path. Of the test cases I believe we could convert `SC_rgb_2frame.dcm` to have a planar configuration = 1 to make it complete. How can I add this to the test-files repo?

I have been slowly working on a new release for the `dicom-test-files` crate (I am one of the owners at the moment) so that it supports more test files. You can send in a pull request that adds a custom test file so that I can review it. The steps would be 1) copy the new file to a new `data/custom/` folder, 2) run the script `generate/generate.py` to regenerate the file entries, and push the file plus the modified `entries.rs`.

Enet4 (approved):
I tested this with a few test files and could validate that the behavior has improved in accuracy and performance after these changes. I believe we are better off merging this now and reiterating on the test suite after `dicom-test-files` is updated. Thanks!

PR #434, opened 2023-11-03
[findscu] Add response status check in C-FIND-RSP response data set

Enet4:
Fixes a situation where the SCU would wait eternally for more messages if the SCP sent a command set with the status _Pending_ but then responded with status 0 in the following data set.

### DIFF
--- a/findscu/src/main.rs
+++ b/findscu/src/main.rs
@@ -322,7 +322,7 @@ fn run() -> Result<(), Error> {
                     // check DICOM status,
                     // as some implementations might report status code 0
                     // upon sending the response data
-                    let status = cmd_obj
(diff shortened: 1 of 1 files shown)


findscu/src/main.rs
```diff
@@ -318,6 +318,23 @@ fn run() -> Result<(), Error> {
                     DumpOptions::new()
                         .dump_object(&dcm)
                         .context(DumpOutputSnafu)?;
+
+                    // check DICOM status,
+                    // as some implementations might report status code 0
+                    // upon sending the response data
+                    let status = cmd_obj
```
Enet4, line 325:
```suggestion
                    let status = dcm
```

PR #437, opened 2023-11-16
ENH: Add lookups into shared/per frame functional groups

naterichman:
~~* Currently implemented and tested for rescale slope/intercept only~~

### DIFF
--- a/pixeldata/src/attribute.rs
+++ b/pixeldata/src/attribute.rs
@@ -78,12 +78,27 @@ pub enum GetAttributeError {
         value: String,
         backtrace: Backtrace,
     },
-
     #[snafu(visibility(pub))]
-    #[snafu(display("Lengths must all be the same for `{:?}`, found `{:?}`", items, values))]
-    LengthMismatch {
-        items: Vec<AttributeName>,
-        values: Vec<String>,
+    #[snafu(display("Value multiplicity of VOI LUT Function must match the number of frames. Expected `{:?}`, found `{:?}`", nr_frames, vm))]
+    LengthMismatchVoiLutFunction {
+        vm: u32,
+        nr_frames: u32,
+        backtrace: Backtrace,
+    },
+    #[snafu(visibility(pub))]
+    #[snafu(display("Value multiplicity of Rescale Slope/Intercept must match the number of frames. Expected `{:?}`, found `{:?}` (slope), `{:?}` (intercept)", nr_frames, slope_vm, intercept_vm))]
+    LengthMismatchRescale {
+        intercept_vm: u32,
+        slope_vm: u32,
+        nr_frames: u32,
+        backtrace: Backtrace,
+    },
+    #[snafu(visibility(pub))]
+    #[snafu(display("Value multiplicity of Window Center/Width must match the number of frames. Expected `{:?}`, found `{:?}` (center), `{:?}` (width)", nr_frames, wc_vm, ww_vm))]
+    LengthMismatchWindowLevel {
+        wc_vm: u32,
+        ww_vm: u32,
+        nr_frames: u32,
         backtrace: Backtrace,
     },
 }
--- a/pixeldata/src/attribute.rs
+++ b/pixeldata/src/attribute.rs
@@ -1,8 +1,9 @@
 //! Utility module for fetching key attributes from a DICOM object.
 
-use dicom_core::{DataDictionary, Tag};
+use dicom_core::{DataDictionary, Tag, ops::AttributeSelector};
 use dicom_dictionary_std::tags;
 use dicom_object::{mem::InMemElement, FileDicomObject, InMemDicomObject};
+use num_traits::FloatConst;
 use snafu::{ensure, Backtrace, OptionExt, ResultExt, Snafu};
 use std::fmt;
 
@@ -152,20 +153,90 @@ pub fn pixel_data<D: DataDictionary + Clone>(
         .context(MissingRequiredSnafu { name })
 }
 
+
+// todo: maybe generalize in the future
+fn get_f64_from_per_frame<D: DataDictionary + Clone>(obj: &FileDicomObject<InMemDicomObject<D>>, selector: [Tag; 2], default: f64) -> Option<Vec<f64>>{
+    obj.element(tags::PER_FRAME_FUNCTIONAL_GROUPS_SEQUENCE).ok()
+        .and_then(|seq| seq.items())
+        .and_then(|items| 
+            items
+            .iter()
+            .map(|item| {
+                item.element(selector[0]).ok()
+                    .and_then(|seq| seq.items())
+                    .and_then(|items| 
+                        if items.len() > 0 {
+                            items[0].element(selector[1]).ok()
+                                .and_then(|e| e.to_float64().ok())
+                        } else {
+                            Some(default)
+                        })
+            })
+            .collect::<Option<Vec<f64>>>()
+        )
+}
+
+fn get_f64_from_shared<D: DataDictionary + Clone>(obj: &FileDicomObject<InMemDicomObject<D>>, selector: [Tag; 2], default: f64) -> Option<Vec<f64>>{
+    obj.element(tags::SHARED_FUNCTIONAL_GROUPS_SEQUENCE).ok()
+        .and_then(|seq| seq.items())
+        .and_then(|items| 
+            if items.len() > 0 {
+                items[0].element(selector[0]).ok()
+                    .and_then(|seq| seq.items())
+                    .and_then(|items| if items.len() > 0 {
+                        items[0].element(selector[1]).ok()
+                            .and_then(|e|
+                                vec![e.to_float64().ok()].into_iter().collect::<Option<Vec<f64>>>()
+                            )
+                    } else {
+                        Some(vec![default])
+                    })
+            } else {
+                Some(vec![default])
+            }
+            
+        )
+}
+
+
 /// Get the RescaleIntercept from the DICOM object or returns 0
 pub fn rescale_intercept<D: DataDictionary + Clone>(
     obj: &FileDicomObject<InMemDicomObject<D>>,
-) -> f64 {
-    obj.element(tags::RESCALE_INTERCEPT)
-        .map_or(Ok(0.), |e| e.to_float64())
-        .unwrap_or(0.)
+) -> Vec<f64> {
+    obj.element(tags::RESCALE_INTERCEPT).ok()
+        .and_then(|e| vec![e.to_float64().ok()].into_iter().collect::<Option<Vec<f64>>>())
+        .or(
+            get_f64_from_per_frame(
+                obj, 
+                [tags::PIXEL_VALUE_TRANSFORMATION_SEQUENCE, tags::RESCALE_INTERCEPT],
+                0.
+        ))
+        .or(
+            get_f64_from_shared(
+                obj, 
+                [tags::PIXEL_VALUE_TRANSFORMATION_SEQUENCE, tags::RESCALE_INTERCEPT],
+                0.
+        ))
+        .unwrap_or(vec![0.])
 }
 
 /// Get the RescaleSlope from the DICOM object or returns 1.0
-pub fn rescale_slope<D: DataDictionary + Clone>(obj: &FileDicomObject<InMemDicomObject<D>>) -> f64 {
-    obj.element(tags::RESCALE_SLOPE)
-        .map_or(Ok(1.0), |e| e.to_float64())
-        .unwrap_or(1.0)
+pub fn rescale_slope<D: DataDictionary + Clone>(obj: &FileDicomObject<InMemDicomObject<D>>) -> Vec<f64> {
+    obj.element(tags::RESCALE_SLOPE).ok()
+        .and_then(|e| vec![e.to_float64().ok()].into_iter().collect::<Option<Vec<f64>>>())
+        .or(
+            get_f64_from_per_frame(
+                obj, 
+                [tags::PIXEL_VALUE_TRANSFORMATION_SEQUENCE, tags::RESCALE_SLOPE],
+                1.0
+        ))
+        .or(
+            get_f64_from_shared(
+                obj, 
+                [tags::PIXEL_VALUE_TRANSFORMATION_SEQUENCE, tags::RESCALE_SLOPE],
+                1.0
+        ))
+        .unwrap_or(vec![1.0])
 }
 
 /// Get the NumberOfFrames from the DICOM object,
@@ -497,6 +568,11 @@ pub fn photometric_interpretation<D: DataDictionary + Clone>(
 
 #[cfg(test)]
 mod tests {
+    use dicom_core::{DataElement, dicom_value, VR, ops::{AttributeOp, ApplyOp, AttributeAction, AttributeSelectorStep}, PrimitiveValue, value::DataSetSequence};
+    use dicom_dictionary_std::{tags, uids};
+    use dicom_object::{InMemDicomObject, FileDicomObject, FileMetaTable, FileMetaTableBuilder};
+    use super::rescale_intercept;
+
 
     #[test]
     fn errors_are_not_too_large() {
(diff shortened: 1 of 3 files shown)


pixeldata/src/attribute.rs
```diff
@@ -152,20 +153,90 @@ pub fn pixel_data<D: DataDictionary + Clone>(
         .context(MissingRequiredSnafu { name })
 }
 
+
+// todo: maybe generalize in the future
+fn get_f64_from_per_frame<D: DataDictionary + Clone>(obj: &FileDicomObject<InMemDicomObject<D>>, selector: [Tag; 2], default: f64) -> Option<Vec<f64>>{
```
naterichman, line 158: I'd like to generalize these methods to return a `Vec` of `DataElement`s, and then let the upstream method (`rescale_slope`/`rescale_intercept`/`voi_lut_function`) handle type conversions and errors, but I'm really struggling with figuring out how to use snafu `context`  in these chains

pixeldata/src/attribute.rs
```diff
@@ -152,20 +153,101 @@ pub fn pixel_data<D: DataDictionary + Clone>(
         .context(MissingRequiredSnafu { name })
 }
 
+fn get_from_shared<D: DataDictionary + Clone>(obj: &FileDicomObject<InMemDicomObject<D>>, selector: [Tag; 2]) -> Option<Vec<&InMemElement<D>>> {
+    obj.element(tags::SHARED_FUNCTIONAL_GROUPS_SEQUENCE).ok()
```
Enet4, line 157: `obj.get` is a more appealing method here. It was introduced some time after `.element` and takes advantage of the fact that element retrieval never fails when it exists.

```suggestion
    obj.element(tags::SHARED_FUNCTIONAL_GROUPS_SEQUENCE).ok()
```

Since this function returns an `Option<_>`, you can chain `Some` contents using the `?` operator. For example:

```rust
let e = obj.get(tags::SHARED_FUNCTIONAL_GROUPS_SEQUENCE)?;
let item = e.items()?.get(0)?;
// ...
```

Although it's worth noting that some parts of the code may be hiding input data errors (`to_float64().ok()` means that we would have no way of knowing whether the value could not be converted to `f64` or the value was missing). `Options` and `Result` of any kind can be harmonized together into a `Result` with a specific error type through Snafu.

pixeldata/src/attribute.rs
```diff
@@ -507,4 +594,71 @@ mod tests {
             size
         );
     }
+
+    #[test]
+    fn test_get_rescale_intercept(){
+        let mut dcm = FileDicomObject::new_empty_with_meta(
+            FileMetaTableBuilder::new()
+                .transfer_syntax(uids::EXPLICIT_VR_LITTLE_ENDIAN)
+                .media_storage_sop_class_uid("1")
```
Enet4, line 603: I'd be inclined to specifying a real, standard SOP class UID, so as to open up the possibility of it being looked up during the decoding process.

```suggestion
                .media_storage_sop_class_uid(uuid::BREAST_TOMOSYNTHESIS_IMAGE_STORAGE)
```

pixeldata/src/attribute.rs
```diff
@@ -507,4 +594,71 @@ mod tests {
...
+                .media_storage_sop_instance_uid("1")
```
Enet4, line 604: For the SOP instance UID, we can generate a UUID-derived one just for this test.

```suggestion
                .media_storage_sop_instance_uid("2.25.187486992583214674353309560233541086440")
```

There are some tools to generate UIDs, such as the `dicom-gen-uid` crate, or [this website](https://enet4.github.io/dicom-uid-website/).

Enet4 (commented):
Thank you for picking this up. I will just leave some comments inline for now.

naterichman:
Thanks for the review.  I will integrate the comments.  I also went through and added changes to the `DecodedPixelData` and the `ImagingProperties`, etc.  Let me integrate the changes you suggested, and fix the tests that I've broken, and then I'll ping you again for a rereview!

pixeldata/src/attribute.rs
naterichman: Leaving this open as I implemented the suggestions, but still left the `get_from_shared` and `get_from_per_frame` returning `Option`s instead of `Result`s.  For  `voi_lut_function` there are a few specific called out errors, but for the  others `rescale_slope` /`rescale_intercept`/`window_center`/`window_width` I've left them just basically just returning Options with a default value if there is any error looking up or converting.  Not sure if you want specific errors called out there

naterichman:
Tests are passing and I added changes to the `DecodedPixelData` object as well, I think this is now ready to be not a draft.

pixeldata/src/lib.rs
```diff
@@ -1724,23 +1760,53 @@ impl ImagingProperties {
         let bits_stored = bits_stored(obj)?;
         let high_bit = high_bit(obj)?;
         let pixel_representation = pixel_representation(obj)?;
+        // TODO: Update to vecs.
```
Enet4, line 1763: Is this still a todo?

pixeldata/src/lib.rs
```diff
@@ -532,17 +531,16 @@ impl DecodedPixelData<'_> {
 
     /// Retrieve object's rescale parameters.
     #[inline]
-    pub fn rescale(&self) -> Rescale {
-        Rescale {
-            intercept: self.rescale_intercept,
-            slope: self.rescale_slope,
-        }
+    pub fn rescale(&self) -> Vec<Rescale> {
```
Enet4, line 534: Maybe now we are better off storing `Rescale` values in `DecodedPixelData` itself (instead of separate `rescale_intercept` and `rescale_slope`), so that we can return a slice without making copies.

pixeldata/src/lib.rs
```diff
@@ -532,17 +531,16 @@ impl DecodedPixelData<'_> {
...
+        zip(&self.rescale_intercept, &self.rescale_slope)
+            .map(|(intercept, slope)| Rescale { intercept: *intercept, slope: *slope })
+            .collect()
     }
 
     /// Retrieve the VOI LUT function defined by the object, if any.
     #[inline]
-    pub fn voi_lut_function(&self) -> Option<VoiLutFunction> {
-        self.voi_lut_function
+    pub fn voi_lut_function(&self) -> Option<&Vec<VoiLutFunction>> {
```
Enet4, line 542: Returning `&[T]` instead of `&Vec<T>` would be more idiomatic and enable us to change the inner container type later.

```suggestion
    pub fn voi_lut_function(&self) -> Option<&[VoiLutFunction]> {
```

Enet4 (commented):
From what I understood, the decoded object's transform LUT properties were expanded into `Vec<_>` so that there would always be one for each frame, correct? There may be room for some optimizations on how we do this (if the same properties apply to all frames, we wouldn't need to allocate 1-item vectors), but they can be deferred to a later date.
Some more suggestions inline.

pixeldata/src/attribute.rs
```diff
@@ -3,6 +3,7 @@
 use dicom_core::{DataDictionary, Tag};
 use dicom_dictionary_std::tags;
 use dicom_object::{mem::InMemElement, FileDicomObject, InMemDicomObject};
+use num_traits::FloatConst;
```
Enet4, line 6: CI claims that this import is unused.

```suggestion
```

pixeldata/src/attribute.rs
```diff
@@ -76,6 +78,14 @@ pub enum GetAttributeError {
         value: String,
         backtrace: Backtrace,
     },
+
+    #[snafu(visibility(pub))]
+    #[snafu(display("Lengths must all be the same for `{:?}`, found `{:?}`", items, values))]
+    LengthMismatch {
```
Enet4, line 84: This enum variant tripped the error type size check. Large errors lead to large `Result`s, negatively affecting performance even in the happy path.

I think we can further decompose it into more specific variants with equally more specific details. It would go like this:

```rust
    /// value multiplicity of VOI LUT Function must match the number of frames ({voi_lut_function_vm} vs {number_of_frames})
    LengthMismatchVoiLutFunction {
        voi_lut_function_vm: u32,
        number_of_frames: u32,
        backtrace: Backtrace,
    },
    /// value multiplicity of Rescale Intercept and Rescale Slope must match the number of frames ({rescale_intercept_vm} & {rescale_slope_vm} vs {number_of_frames})
    LengthMismatchRescale {
        rescale_intercept_vm: u32,
        rescale_slope_vm: u32,
        number_of_frames: usize,
        backtrace: Backtrace,
    },
    /// value multiplicity of Window Center and Window Width must match the number of frames ({window_center_vm} & {window_width_vm} vs {number_of_frames})
    LengthMismatchWindowLevel {
        rescale_intercept_vm: u32,
        rescale_slope_vm: u32,
        number_of_frames: u32,
        backtrace: Backtrace,
    },
```

I believe this would constrain the size more nicely while conveying the same information.

Enet4 (changes requested):
If you need some help with the suggested changes or would prefer that I take over instead, please let me know.

pixeldata/src/attribute.rs
naterichman: No problem, I will go ahead and do that, hope you had a good holiday season!

naterichman: Okay, updated the code, I wasn't familiar with how docstrings get formatted in snafu and the rest of the enum doesn't use docstrings, so I went with the specific `snafu(display())`, lmk if you want me to change that

pixeldata/src/gdcm.rs
```diff
@@ -55,7 +56,26 @@ where
...
+                .map(|v| VoiLutFunction::try_from((*v).as_str()).ok())
+                .collect()
+            );
+        if let Some(inner) = &voi_lut_function {
+            if !(inner.len() == number_of_frames as usize || inner.len() == 1) {
+                LengthMismatchVoiLutFunctionSnafu {
+                    vm: inner.len() as u32,
+                    nr_frames: number_of_frames as u32
+                }.fail().context(GetAttributeSnafu)?;
+            }
```
naterichman, line 70: I'm not super happy with this, I wanted to use the `ensure` macro like I did in `lib.rs`, but I couldn't figure out a way to get it to work...

naterichman (commented):
Couple of questions in general if you don't mind.

* I didn't catch any of the GDCM stuff because my rust-analyzer had gdcm feature turned off by default.  I was able to work around that by adding `gdcm` to the `default` features in my pixeldata `Cargo.toml`, but that seems hacky, is there a better way to do that that you know of?
* Relating to the above, is there way to run the CI jobs locally before I push to make sure they will pass?  I'm coming over from python and have used [pre-commit](https://pre-commit.com/) quite a bit.  I've also seen some projects using [`just`](https://github.com/casey/just). Wondering if you have anything in place that I'm missing or if it would be worth adding something like `pre-commit` or `just`.  LMK!

ingwinlu:
> * I didn't catch any of the GDCM stuff because my rust-analyzer had gdcm feature turned off by default.  I was able to work around that by adding `gdcm` to the `default` features in my pixeldata `Cargo.toml`, but that seems hacky, is there a better way to do that that you know of?

Thats how I did it the first time too ;)

rust-analyzer.cargo.features is exposed in vscode if you use that. I am sure anything based on rust-analyzer would have a similar way of passing a value to rust-analyzer.

pixeldata/src/lib.rs
```diff
@@ -402,17 +402,17 @@ pub struct DecodedPixelData<'a> {
...
+    /// 
+    /// the pixel value rescale slope and intercept
+    rescale: Vec<Rescale>,
     // the VOI LUT function
-    voi_lut_function: Option<VoiLutFunction>,
+    voi_lut_function: Option<Vec<VoiLutFunction>>,
     /// the window level specified via width and center
-    window: Option<WindowLevel>,
-    // TODO(#232): VOI LUT sequence is currently not supported
+    window: Option<Vec<WindowLevel>>,
```
Enet4, line 415: Not something worth acting upon right now, but some food for thought: it is possible for a single image to have multiple window levels (plus WW&WC explanation) even if there is a single frame (https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.2.html). How should we handle these cases moving forward?

Enet4 (commented):
> * I didn't catch any of the GDCM stuff because my rust-analyzer had gdcm feature turned off by default.  I was able to work around that by adding `gdcm` to the `default` features in my pixeldata `Cargo.toml`, but that seems hacky, is there a better way to do that that you know of?

Switching back and forth can be done through cargo parameters (`--features gdcm`) or by adjusting rust_analyzer in the IDE (thanks @ingwinlu).
The two implementations are pretty much exclusive.

A step to build against GDCM is included in CI in order to prevent larger mistakes from happening because of this.

> * Relating to the above, is there way to run the CI jobs locally before I push to make sure they will pass?  I'm coming over from python and have used [pre-commit](https://pre-commit.com/) quite a bit.  I've also seen some projects using [`just`](https://github.com/casey/just). Wondering if you have anything in place that I'm missing or if it would be worth adding something like `pre-commit` or `just`.  LMK!

There is [`act`](https://github.com/nektos/act), but I often just run the tests which relate to the changes made (e.g. `cargo test -p dicom-pixeldata` if I only want to run the tests for `pixeldata`), which is much more convenient. From the experience that I had with pre-commit hooks, I feel that they can be too much a computational burden to impose onto contributors at this time. In any case, extensions to CONTRIBUTING.md with nice tips and recommendations would be good to have!

---

Just a few more recommendations inline.

pixeldata/src/gdcm.rs
naterichman: `ImagingProperties` in `pixeldata/src/lib.rs` returns a `Result<Self, attribute::GetAttributeError>`, if I moved the `LengthMismatch*` errors up to `InnerError` I would need to change change the return signature of `ImagingProperties` and then add the `fail().context(GetAttributeSnafu)` to all the fallible operations in `from_obj`, which seems to me more code changes.  

I'm happy to go with either, just lmk which interface you would prefer, or if I'm missing something!

pixeldata/src/lib.rs
naterichman: I see, would that be for highlighting different sections of tissue in an image, with different WW/WC values? I feel like we should default to using the first VOI, but expose methods on the `DecodedPixedlData` struct to apply the different VOIs

naterichman:
Thanks again for the great review comments!

Enet4 (changes requested):
I sense that we might need to give more leeway to unexpected inputs. While testing this, I found a file which features per-frame functional groups, but did not have a VOI LUT Sequence (functional groups were used for other attributes). In this scenario I would just expect the program continuing with the assumption that _VOI LUT Function_ was not explicitly provided, but instead it assumed that this attribute should have been present. There might be similar issues for the other attributes. 

```none
2024-02-02T15:13:14.513214Z ERROR dicom_toimage: failed to decode pixel data

Caused by this error:
  1: Value multiplicity of VOI LUT Function must match the number of frames. Expected `1`, found `0`
```

I'm not sure whether I'll be able to come up with a full anonymized DICOM file, but this partial dump should give you an idea of the problem. Basically, rescale attributes are available in the Pixel Value Transformation Sequence, but the whole VOI LUT is left unspecified. This should turn into a warning, as in the current behavior.

```none
(0028,0002) SamplesPerPixel              US (1,  2 bytes): 1
(0028,0004) PhotometricInterpretation    CS (1, 12 bytes): "MONOCHROME2"
(0028,0008) NumberOfFrames               IS (1,  2 bytes): "1"
(0028,0010) Rows                         US (1,  2 bytes): 672
(0028,0011) Columns                      US (1,  2 bytes): 672
(0028,0100) BitsAllocated                US (1,  2 bytes): 16
(0028,0101) BitsStored                   US (1,  2 bytes): 12
(0028,0102) HighBit                      US (1,  2 bytes): 11
(0028,0103) PixelRepresentation          US (1,  2 bytes): 0
(5200,9229) SharedFunctionalGroupsSequence SQ (1 Item)
    (FFFE,E000) na Item
      (0020,9116) PlaneOrientationSequence     SQ (1 Item)
          (FFFE,E000) na Item
            (0020,0037) ImageOrientationPatient      DS (6, 94 bytes): [".9962408542633", "-.0274781603366", "-.0821528509259", "-.0036975343246", ".93400490283966", "-.3572410345077"]
          (FFFE,E00D) ItemDelimitationItem
      (FFFE,E0DD) SequenceDelimitationItem
      (0028,9110) PixelMeasuresSequence        SQ (1 Item)
          (FFFE,E000) na Item
            (0018,0088) SpacingBetweenSlices         DS (1,  4 bytes): "4.4"
            (0028,0030) PixelSpacing                 DS (2, 30 bytes): [".3571428656578", ".3571428656578"]
          (FFFE,E00D) ItemDelimitationItem
      (FFFE,E0DD) SequenceDelimitationItem
      (0028,9145) PixelValueTransformationSequence SQ (1 Item)
          (FFFE,E000) na Item
            (0028,1052) RescaleIntercept             DS (1,  2 bytes): "0"
            (0028,1053) RescaleSlope                 DS (1, 16 bytes): "1.21514041514041"
          (FFFE,E00D) ItemDelimitationItem
      (FFFE,E0DD) SequenceDelimitationItem
    (FFFE,E00D) ItemDelimitationItem
(FFFE,E0DD) SequenceDelimitationItem
(5200,9230) PerFrameFunctionalGroupsSequence SQ (1 Item)
    (FFFE,E000) na Item
      (0020,9113) PlanePositionSequence        SQ (1 Item)
          (FFFE,E000) na Item
            (0020,0032) ImagePositionPatient         DS (3, 50 bytes): ["-117.20093870169", "-142.11923001646", "-4.5131105119807"]
          (FFFE,E00D) ItemDelimitationItem
      (FFFE,E0DD) SequenceDelimitationItem
    (FFFE,E00D) ItemDelimitationItem
(FFFE,E0DD) SequenceDelimitationItem
```


And in retrospection, I'm starting to think that different cardinality mismatches (e.g. 1 vs 3 frames) should not be something preventing the software from producing an image, although it should at the very least be a warning. This distinction would probably best be handled at the Decoded pixel data API, so that.

If you are still in good position to fix this, I would be very grateful.

naterichman:
HI @Enet4 I can definitely still work on this, might not be for a week or so though!

naterichman:
@Enet4 I added what I interpreted from your message above, but I'm not sure those checks are in the optimal place.  LMK your thoughts and if you think its good, I'll go ahead with cleaning it up and adding tests

Enet4 (changes requested):
Oh, I had my review just about ready a week ago, but I must have forgotten to submit it.

Enet4 (approved):
:+1: Thank you very much, @naterichman.

naterichman:
Thank you, glad to be officially contributing to this project and hope to continue to!

Enet4 (commented):
Oops! There are still some changes to make in the gdcm implementation. Would you be able to work on this?

Enet4:
> Thank you, glad to be officially contributing to this project and hope to continue to!

I am sorry to report that that opportunity has arrived so soon! :sweat_smile:

naterichman:
Sure thing!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

naterichman:
I got the tests working for gdcm now, but I'm getting an issue locally in the `openjpeg-sys` feature:

```rust
$ cargo test -p dicom-pixeldata --features openjpeg-sys
   Compiling jpeg2k v0.6.6
   Compiling dicom-parser v0.6.0 (/home/naterichman/Documents/rust/dicom-rs/parser)
   Compiling dicom-test-files v0.2.1
error[E0255]: the name `sys` is defined multiple times
  --> /home/naterichman/.cargo/registry/src/index.crates.io-6f17d22bba15001f/jpeg2k-0.6.6/src/lib.rs:36:1
   |
33 | pub(crate) use openjpeg_sys as sys;
   |                ------------------- previous import of the module `sys` here
...
36 | pub(crate) mod sys {
   | ^^^^^^^^^^^^^^^^^^ `sys` redefined here
   |
   = note: `sys` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
   |
33 | pub(crate) use openjpeg_sys as other_sys;
   |                ~~~~~~~~~~~~~~~~~~~~~~~~~

For more information about this error, try `rustc --explain E0255`.
error: could not compile `jpeg2k` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

I'm also getting that on master though...

Enet4:
Did you perchance add the Cargo feature "jpeg-sys" without also excluding the default feature "openjp2"?

You can also rebase the branch to remove `openjp2` from being enabled by default.

pixeldata/src/lib.rs
```diff
@@ -121,8 +121,8 @@ use num_traits::NumCast;
 use rayon::iter::{IntoParallelIterator, IntoParallelRefIterator, ParallelIterator};
 #[cfg(all(feature = "rayon", feature = "image"))]
 use rayon::slice::ParallelSliceMut;
-use snafu::OptionExt;
-use snafu::{Backtrace, ResultExt, Snafu};
+use snafu::{Backtrace, ResultExt, Snafu, OptionExt, ensure};
+use std::iter::zip;
```
Enet4, line 125:
```suggestion
use snafu::{Backtrace, ResultExt, Snafu, OptionExt};
#[cfg(not(feature = "gdcm"))]
use snafu::ensure;
#[cfg(not(feature = "gdcm"))]
use std::iter::zip;
```

Enet4 (approved):
OK, it should be in order this time. :+1: Thank you again!

PR #440, opened 2023-11-21
[dicom-dump] Add json output format

naterichman:
Add flag to output in dicom json format


dump/src/lib.rs
```diff
@@ -227,14 +224,23 @@ impl DumpOptions {
...
 
-        writeln!(to, "{:-<58}", "")?;
+                dump(&mut to, obj, width, 0, no_text_limit, no_limit)?;
 
-        dump(&mut to, obj, width, 0, no_text_limit, no_limit)?;
+                Ok(())
+            },
+            DumpFormat::Json => {
+                let json_obj = DicomJson::from(obj);
+                println!("{}", serde_json::to_string_pretty(&json_obj)?);
```
Enet4, line 239: `serde_json` has [`to_writer_pretty`](https://docs.rs/serde_json/latest/serde_json/fn.to_writer_pretty.html), which might be more efficient if we pass it [`stdout()`](https://doc.rust-lang.org/std/io/fn.stdout.html).

dump/src/main.rs
```diff
@@ -37,6 +44,14 @@ struct App {
     /// Fail if any errors are encountered
     #[clap(long = "fail-first")]
     fail_first: bool,
+    /// Output format
+    #[arg(value_enum)]
+    #[clap(short = 'o', long = "output", default_value = "text")]
```
Enet4, line 49: I'd say we rename the option to `format` (short `-F`, long `--format`), which not only aligns more with the library, but also does not give the false impression that this is for an output file path, with `-o` often being used for this purpose.

Enet4 (changes requested):
Great initiative, thank you! Some inline suggestions follow.

For transparency, I intend to throw a new patch release of the project soon, which will not include these new PRs, but after 0.6.3 we can start reviewing greater changes towards 0.7.

dump/src/main.rs
naterichman: Sounds good and makes sense, I am just used to `-o yaml` and `-o json` from `kubectl`!

Enet4:
So I took the liberty of continuing with this pull request:

- rebased it to the latest upstream revision
- updated the doc comment for `DumpFormat::Json`
- ensured that dumping as JSON also works for regular (non-file) DICOM objects
- added a test for dumping a small DICOM object as JSON

Alas, this might have hit a very weird bug in Rust, because it started raising absurd errors in another crate. I might need to look for advice with members of the Rust team on how to diagnose this.

naterichman:
Thanks for working on this, sorry I've been AWOL, my day job has been very busy!  Just checking, is there anything I can do here to help?  I still plan on working on my other PR as well

Enet4:
Thank you for the heads-up. I managed to get things sorted out in this PR. I suspect at this point that the dependency changes may have introduced trait implementations which ambiguated the `!=` comparison between message IDs in `dicom-echoscu`, leading to the unrelated error. So no more interventions should be required here, I'll just be some final testing before merging.

I would certainly be grateful if you did another iteration on #437, so we can complete that one as well.

Enet4:
Further testing shows that the tool chokes a bit on files with encapsulated pixel data, whereas it would be nicer if it just ignored the pixel data with a warning. I suspect that this would complicate the implementation though.

naterichman:
Sorry for the delay on this one, kinda forgot about it!  I added a simple serializer for `PixelDataFragment`s which splits it up into the offset table and the fragments.  LMK your thoughts.  I thought about just ignoring the `PixelData`, but I realized that would make the behavior inconsistent with the text output.  Would be pretty easy to just ignore it if you think thats better, lmk

json/src/ser/mod.rs
```diff
@@ -286,6 +284,20 @@ impl<D> Serialize for DicomJson<InMemElement<D>> {
...
 
+impl Serialize for DicomJson<&PixelFragmentSequence<Vec<u8>>> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let offset_table = self.inner().offset_table();
+        let fragments = self.inner().fragments();
+        let mut map = serializer.serialize_map(Some(2))?;
+        map.serialize_entry("Offset Table", offset_table)?;
```
Enet4, line 295: I have mixed feelings about this. :) It may be interesting for encapsulated pixel data serialization to be supported in some way, but it may also be surprising for this to exist as a non-standard extension provided specifically by this implementation.

I think we'd be able to serve both worlds if we did the following in this order:

1. At dicom-dump, we can check whether Pixel Data is encapsulated, raise a warning that it will be ignored, and remove it from the dataset before the dump.
2. Later on, we can make it possible to provide parameters to the serialization process, so that the decision of whether to serialize encapsulated pixel data or not can be explicitly stated by the API user or the `dicom-dump` user.
   - IIRC we don't quite have a way to do this yet. Intuitively I would make this a field of `DicomJson`, but since it is defined as a single item tuple, adding new fields is likely breaking change (which could probably be mitigated through some tricker, but a breaking change nevertheless).

Enet4 (commented):
Thank you for your commitment to this PR. I left some feedback inline.

Enet4 (approved):
This seems to be in good order now. Thanks! :+1:

Issue #443, opened 2023-11-24
Support for multiple character repertoires (extended character sets)


Enet4:
As described in [PS3.5 6.1.2.2](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/chapter_6.html#sect_6.1.2.2).

PR #444, opened 2023-11-25
Add ISO IR 13 and ISO IR 87 to SpecificCharacterSet

9enki:
issue #443

### DIFF
--- a/encoding/src/text.rs
+++ b/encoding/src/text.rs
@@ -18,7 +18,10 @@
 //!
 //! [`SpecificCharacterSet`]: ./enum.SpecificCharacterSet.html
 
-use encoding::all::{GB18030, ISO_8859_1, ISO_8859_2, ISO_8859_3, ISO_8859_4, ISO_8859_5, UTF_8};
+use encoding::all::{
+    GB18030, ISO_2022_JP, ISO_8859_1, ISO_8859_2, ISO_8859_3, ISO_8859_4, ISO_8859_5, UTF_8,
+    WINDOWS_31J,
+};
 use encoding::{DecoderTrap, EncoderTrap, Encoding, RawDecoder, StringWriter};
 use snafu::{Backtrace, Snafu};
 use std::borrow::Cow;
@@ -148,6 +151,10 @@ pub enum SpecificCharacterSet {
     IsoIr192,
     /// **GB18030**: The Simplified Chinese character set.
     Gb18030,
+    /// **IsoIr13**: The Simplified Japanese single byte character set.
+    IsoIr13,
+    /// **IsoIr87**: The Simplified Japanese multi byte character set.
+    IsoIr87,
     // Support for more text encodings is tracked in issue #40.
 }
 
@@ -175,6 +182,8 @@ impl SpecificCharacterSet {
             "ISO_IR_144" | "ISO_IR 144" | "ISO 2022 IR 144" => Some(IsoIr144),
             "ISO_IR_192" | "ISO_IR 192" => Some(IsoIr192),
             "GB18030" => Some(Gb18030),
+            "ISO_IR_13" | "ISO 2022 IR 13" => Some(IsoIr13),
+            "ISO_IR_87" | "ISO 2022 IR 87" => Some(IsoIr87),
             _ => None,
         }
     }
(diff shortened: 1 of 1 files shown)


ikneg:
@Enet4 Hi, thanks for creating issue#443.

For Japanese, Chinese or Korean, the Specific Character Set (0008,0005) may have multiple values. https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_H.3.html
https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_H.3.2.html

In that case, the Patient Name value needs to be decoded in the Specific Character Set for each `=` separated part. In the current code, it seems that even if there are multiple values in the Specification Character Set, they are decoded by the first element in that set, is my understanding correct?

If so, I would like to fix it in this PR or another PR, could you please advise where I should consider fixing it? I am tracing the code back from the following function
https://github.com/ikneg/dicom-rs/blob/7538e4f16d73e8bf0a36b472e0812f2541879be9/object/src/lib.rs#L189

I guessed that the place where the values are read from the binary data is the following place, so I tried to output the values to the log here with info!
https://github.com/ikneg/dicom-rs/blob/7538e4f16d73e8bf0a36b472e0812f2541879be9/object/src/mem.rs#L1514
```log
2023-11-27T02:26:26.629565Z  INFO dicom_object::mem: next tokne: Ok(PrimitiveValue(Strs(["填塹^些灼=\u{1b}$BCf;3\u{1b}(J^\u{1b}$B9'<#\u{1b}(J "])))
```
But here it is already of type Strs, so it seems that I need to understand more of the preceding code, but it is stuck here.

ikneg:
```python
>>> specification_character_set = ["shift_jis", "iso2022_jp"]
>>> "ﾔﾏﾀﾞ^ﾀﾛｳ".encode(specification_character_set[0]) + "=".encode("utf-8") + "山田^太郎".encode(specification_character_set[1])
b'\xd4\xcf\xc0\xde^\xc0\xdb\xb3=\x1b$B;3ED\x1b(B^\x1b$BB@O:\x1b(B'

```
As a sample, when written in python, the Patinant Name value seems to be generated like this, so I want to be able to decode the value generated like this as `ﾔﾏﾀﾞ^ﾀﾛｳ=山田^太郎`.

Enet4:
Thank you for working on this @ikneg! Could you please add a few sample texts as unit tests? There should be some for other text encodings at the end of the module, so you would just need to follow the pattern there with new data.

ikneg:
> @Enet4 Hi, thanks for creating issue #443
> 
> For Japanese, Chinese or Korean, the Specific Character Set (0008,0005) may have multiple values. https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_H.3.html https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_H.3.2.html
> 
> In that case, the Patient Name value needs to be decoded in the Specific Character Set for each `=` separated part. In the current code, it seems that even if there are multiple values in the Specification Character Set, they are decoded by the first element in that set, is my understanding correct?
> 
> If so, I would like to fix it in this PR or another PR, could you please advise where I should consider fixing it? I am tracing the code back from the following function https://github.com/ikneg/dicom-rs/blob/7538e4f16d73e8bf0a36b472e0812f2541879be9/object/src/lib.rs#L189
> 
> I guessed that the place where the values are read from the binary data is the following place, so I tried to output the values to the log here with info! https://github.com/ikneg/dicom-rs/blob/7538e4f16d73e8bf0a36b472e0812f2541879be9/object/src/mem.rs#L1514
> 
> ```
> 2023-11-27T02:26:26.629565Z  INFO dicom_object::mem: next tokne: Ok(PrimitiveValue(Strs(["填塹^些灼=\u{1b}$BCf;3\u{1b}(J^\u{1b}$B9'<#\u{1b}(J "])))
> ```
> 
> But here it is already of type Strs, so it seems that I need to understand more of the preceding code, but it is stuck here.

この仕様については複雑な話になるので、本PRの対象から外して別のPRを作って相談したいと思います。

Enet4:
This has been superseded by #614, but we ought to look into reviving the efforts of #445 eventually.

PR #451, opened 2024-01-15
Trim \0 characters from photometric interpretation

dougyau:
trim() does not remove \0 characters, it needs to be done using trim_matches.

### DIFF
--- a/pixeldata/src/attribute.rs
+++ b/pixeldata/src/attribute.rs
@@ -491,8 +491,7 @@ pub fn photometric_interpretation<D: DataDictionary + Clone>(
         .context(MissingRequiredSnafu { name })?
         .string()
         .context(CastValueSnafu { name })?
-        .trim()
-        .trim_matches(char::from(0))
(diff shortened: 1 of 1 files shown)


pixeldata/src/attribute.rs
```diff
@@ -492,6 +492,7 @@ pub fn photometric_interpretation<D: DataDictionary + Clone>(
         .string()
         .context(CastValueSnafu { name })?
         .trim()
+        .trim_matches(char::from(0))
```
Enet4, line 495:
```suggestion
        .trim_matches(|c: char| c.is_whitespace() || c == '\0')
```

Enet4 (commented):
The presence of null characters in this element value is not fully compliant, but we can accommodate this the same way we already did for a few other values such as UIDs.

We can use the same condition too.

dougyau:
Yes, null characters should not be there, but I have found a specific case, and Osko Xmaru Series CR which sends it that way.

Enet4 (approved):
Thank you. :+1:

PR #453, opened 2024-02-08
Datetimefix

### LINKED ISSUE #452: DICOM naive datetime retrieval
The primitive `to_datetime` methods all expect a default, fixed time-offset to be passed, which is then applied to the resulting, timezone-aware datetime struct.
The DICOM standard allows for datetime VRs to have an optional suffix for offset from Coordinated Universal Time. In this case, it shouldn't be necessary to pass a fixed time-offset to the methods that retrieve datetime data.
It would be very helpful if, for a particular datetime primitive, you could retrieve a time-zone aware datetime if it exists, or a naive datetime if not.

jmlaka:
Please see this DicomDateTime rewrite that now can handle time zone aware and naive values.
External FixedOffset values no longer required.
some methods deprecated
some methods documented to be not optimal use of API
Breaking changes
This should fix #452 

Thank you !

### DIFF
--- a/core/src/value/partial.rs
+++ b/core/src/value/partial.rs
@@ -1,7 +1,7 @@
 //! Handling of partial precision of Date, Time and DateTime values.
 
 use crate::value::range::AsRange;
-use chrono::{DateTime, Datelike, FixedOffset, NaiveDate, NaiveTime, Timelike};
+use chrono::{DateTime, Datelike, FixedOffset, NaiveDate, NaiveDateTime, NaiveTime, Timelike};
 use snafu::{Backtrace, ResultExt, Snafu};
 use std::convert::{TryFrom, TryInto};
 use std::fmt;
@@ -228,7 +228,7 @@ enum DicomTimeImpl {
 pub struct DicomDateTime {
     date: DicomDate,
     time: Option<DicomTime>,
-    offset: FixedOffset,
+    offset: Option<FixedOffset>,
 }
 
 /**
@@ -578,11 +578,41 @@ impl DicomDateTime {
     /**
      * Constructs a new `DicomDateTime` from a `DicomDate` and a given `FixedOffset`.
      */
-    pub fn from_date(date: DicomDate, offset: FixedOffset) -> DicomDateTime {
+    pub fn from_date_tz(date: DicomDate, offset: FixedOffset) -> DicomDateTime {
         DicomDateTime {
             date,
             time: None,
-            offset,
+            offset: Some(offset),
+        }
+    }
+
+    /**
+     * Constructs a new `DicomDateTime` from a `DicomDate` .
+     */
+    pub fn from_date(date: DicomDate) -> DicomDateTime {
+        DicomDateTime {
+            date,
+            time: None,
+            offset: None,
+        }
+    }
+
+    /**
+     * Constructs a new `DicomDateTime` from a `DicomDate` and `DicomTime`,
+     * providing that `DicomDate` is precise.
+     */
+    pub fn from_date_and_time(date: DicomDate, time: DicomTime) -> Result<DicomDateTime> {
+        if date.is_precise() {
+            Ok(DicomDateTime {
+                date,
+                time: Some(time),
+                offset: None,
+            })
+        } else {
+            DateTimeFromPartialsSnafu {
+                value: date.precision(),
+            }
+            .fail()
         }
     }
 
@@ -590,7 +620,7 @@ impl DicomDateTime {
      * Constructs a new `DicomDateTime` from a `DicomDate`, `DicomTime` and a given `FixedOffset`,
      * providing that `DicomDate` is precise.
      */
-    pub fn from_date_and_time(
+    pub fn from_date_and_time_tz(
         date: DicomDate,
         time: DicomTime,
         offset: FixedOffset,
@@ -599,7 +629,7 @@ impl DicomDateTime {
             Ok(DicomDateTime {
                 date,
                 time: Some(time),
-                offset,
+                offset: Some(offset),
             })
         } else {
             DateTimeFromPartialsSnafu {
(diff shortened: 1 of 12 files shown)


jmlaka:
Hello, @Enet4. Please check out this pull request. 

My initial questions:
Maybe the `[PreciseDateTimeResult]` struct name is too long ?

The `core `crate now requires the "clock" feature for `chrono` library, I don't know if this is acceptable.
It's needed for the default handling of an ambiguous DT range, where one time zone is parsed and the second one is missing.

Also, for different parsing behavior of this situation a `PrimitiveValue:to_datetime_range_custom<T>` was added.  I'm not sure if you are ok with such addition.

I'm not sure if I deprecated some methods correctly ...
 
Thanks

object/src/mem.rs
```diff
@@ -889,7 +889,7 @@ where
                 AttributeSelectorStep::Nested { tag, item } => {
                     let e = obj
                         .entries
-                        .get(tag)
+                        .get(&tag)
```
Enet4, line 892: This change doesn't seem related, and looks undesirable according to clippy.

core/src/value/primitive.rs
```diff
@@ -3172,56 +2978,191 @@ impl PrimitiveValue {
...
+    /// # Example
+    ///
+    /// ```
+    /// # use dicom_core::value::{C, PrimitiveValue};
+    /// # use std::error::Error;
+    /// use dicom_core::value::range::{AmbiguousDtRangeParser, ToLocalTimeZone, IgnoreTimeZone, FailOnAmbiguousRange, DateTimeRange};
+    /// use chrono::{NaiveDate, NaiveTime, NaiveDateTime};
+    /// # fn main() -> Result<(), Box<dyn Error>> {
+    ///
+    /// // The DICOM protocol allows for parsing text representations of date-time ranges,
```
Enet4, line 3097: Maybe some of these explanations about DICOM date-time representations would fit better in the documentation corpus, rather than inside a doctest?

Enet4 (commented):
Thank you very much for working on this, @jmlaka. I haven't had much time to review things properly, but it would be nice to bring a new version of DICOM-rs the next few months, so I'm aiming for that.

> Maybe the [PreciseDateTimeResult] struct name is too long ?

It should be OK. The only alternative I can think of is PreciseDtResult, but that would reduce the size of the most important part of the type in the name.

> The core crate now requires the "clock" feature for chrono library, I don't know if this is acceptable.It's needed for the default handling of an ambiguous DT range, where one time zone is parsed and the second one is missing.

Interesting, I did not recall that we were excluding some of the default features. As long as it still builds against major platforms and wasm32, it should be fine. If not, we could try some feature gating.

> Also, for different parsing behavior of this situation a PrimitiveValue:to_datetime_range_custom<T> was added. I'm not sure if you are ok with such addition.

At first glance the name `to_datetime_range_custom` is not very clear about what `custom` entails, but considering that it's a custom interpretation (or disambiguation) of the range, I don't see a better name for it. I will strive for a more thorough review on this part next week or so.

> I'm not sure if I deprecated some methods correctly

I left a note on that in the inline comments.

jmlaka:
I fixed the problems you mentioned.
The DicomDateTime had methods, which, on second thought, were superfluous.

core/src/value/partial.rs
```diff
@@ -173,42 +183,45 @@ enum DicomTimeImpl {
...
 ///
-/// `DicomDateTime` is always internally represented by a [DicomDate]
-/// and optionally by a [DicomTime].
+/// `DicomDateTime` is always internally represented by a [DicomDate].
+/// The [DicomTime] and a timezone [FixedOffset] values are optional.
 ///
-/// It implements [AsRange] trait and also holds a [FixedOffset] value, from which corresponding
-/// [datetime][DateTime] values can be retrieved.
+/// It implements [AsRange] trait, which serves to retrieve a [PreciseDateTimeResult] from values
+/// with missing components.
```
Enet4, line 193: cargo doc cannot resolve `PreciseDateTimeResult` in this point, so we need to give it some directions.

```suggestion
/// It implements [AsRange] trait,
/// which serves to retrieve a [PreciseDateTimeResult](crate::core::value::PreciseDateTimeResult)
/// from values with missing components.
```

(with the rename, it would be this)

```suggestion
/// It implements [AsRange] trait,
/// which serves to retrieve a [PreciseDateTime](crate::core::value::PreciseDateTime)
/// from values with missing components.
```

core/src/value/partial.rs
```diff
@@ -173,42 +183,45 @@ enum DicomTimeImpl {
...
 /// let offset = FixedOffset::east_opt(3600).unwrap();
 ///
-/// // the least precise date-time value possible is a 'YYYY'
-/// let dt = DicomDateTime::from_date(
+/// // lets create the least precise date-time value possible 'YYYY' and make it time-zone aware
+/// let dt = DicomDateTime::from_date_with_time_zone(
 ///     DicomDate::from_y(2020)?,
 ///     offset
 /// );
+/// // the earliest possible value is output as a [PreciseDateTimeResult]
```
Enet4, line 209: This is inside a doctest.

```suggestion
/// // the earliest possible value is output as a PreciseDateTimeResult
```

object/src/mem.rs
Enet4: I looked closer into this and it seems that the latest version of rust-analyzer is showing a false positive E0308 in this position. `tag` in this context is already a `&Tag`, so it would not need a reference. This can be verified by running `cargo check` after this change.

```suggestion
                        .get(tag)
```

I filed an issue at rust-analyzer in the meantime. https://github.com/rust-lang/rust-analyzer/issues/16657

Enet4 (commented):
Things look pretty good! I just have a few more changes to suggest, and then we can merge.

Enet4:
I hope you don't mind, I added in a few more changes before merging. Here is the summary:

- Adjusted the documentation around `parse_datetime_partial`, `parse_datetime`, `PreciseDateTime`, and the ambiguous DT parsers.
- Readjusted the methods in `PreciseDateTime` to better reflect its nature as a sum type of chrono date-time values.
- Made a custom `PartialOrd` implementation for `PreciseDateTime` because the derived one would be misleading. Because of how `PartialOrd` is derived, it could occur `2024-03-01 00:00:00 +00` being larger than `2024-03-01 23:59:59`, much to the surprise of its users, just because it had a time zone specified.
- Moved `PreciseDateTime` to live alongside the other DICOM date/time types.
- Added `impl FromStr for DicomDateTime`.

With this, I think we are ready to move forward. :)

jmlaka:
Thank you very much for your added insight.
Happy to help.

Enet4 (approved):
Thank you once again. :+1: We're bringing this in.

PR #459, opened 2024-03-04
[json] Allow arbitrary element order for JsonDataElement

feliwir:
Keep the value of the "value" entry in memory until all fields have been read. This allows the "vr" field to occur after the "value" field

### DIFF
--- a/json/src/de/mod.rs
+++ b/json/src/de/mod.rs
@@ -171,7 +171,7 @@ where
         if let Some(value) = value {
             // deserialize value in different ways
             // depending on VR
-            match vr.unwrap() {
+            match vr {
                 // sequence
                 VR::SQ => {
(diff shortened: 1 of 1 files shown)


Enet4 (commented):
I just added a very minor suggestion inline, but other than that looks pretty good! Thank you @feliwir for reporting this case and providing a patch!

json/src/de/mod.rs
```diff
@@ -298,6 +163,149 @@ where
...
 
+        // ensure that VR is present
+        let Some(vr) = vr else {
+            return Err(A::Error::custom("missing VR field"));
+        };
+
+        if let Some(value) = value {
+            // deserialize value in different ways
+            // depending on VR
+            match vr.unwrap() {
```
Enet4, line 174:
```suggestion
            match vr {
```

PR #463, opened 2024-03-06
[Draft] ENH: Add feature-gated async runtime for storescp

naterichman:
Half  me playing around with async and half because I couldn't get parallel dcmtk `movescu` to work easily, there is a `--fork` option on dcmtk `storescp` and I figured we should maybe have that here too.


naterichman:
Closing in favor of #542

PR #464, opened 2024-03-07
[json] Allow null for string values

feliwir:
See https://dicom.nema.org/dicom/2013/output/chtml/part18/sect_F.2.html in Section F.2.5

Based on #459

### DIFF
--- a/json/src/de/mod.rs
+++ b/json/src/de/mod.rs
@@ -194,8 +194,10 @@ where
                 | VR::TM
                 | VR::UC
                 | VR::UI => {
-                    let items: Vec<String> =
+                    let items: Vec<Option<String>> =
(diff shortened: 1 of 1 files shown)


Enet4 (approved):
Well, this indeed works! I will approve and merge while leaving these notes on potential future changes:

- There is an extra allocation from `Vec<Option<Value>>` to `Vec<Value>`, but this also happens in a few other places as well. Maybe someone could find some time to fine-tune these later.
- This takes care of the conversion from DICOM JSON to in-memory representations, but, thinking reciprocally, I suspect that empty values will still be encoded to DICOM JSON as empty strings with the current implementation. This is worth taking care of in a new pull request.

Thanks again for your contribution @feliwir. :+1:

PR #471, opened 2024-03-11
ENH: Add functionality for mutating element values

naterichman:
* Add `value_at_mut`, `entry_at`, `entry_at_mut` and `update_value_at` methods on `InMemDicomObject` struct.
* Add supporting methods to dependent objects

Let me know if this is something you would like, I'm finding I need something like this for a downstream application.  I realize I can use the `update_value` method, but its nice to be able to `update_value` with an `AttributeSelector`.  If there is a way to do that and I'm missing it, I'm happy to just update the docs to make that clearer instead of this PR.


naterichman:
If this is something you'd like, I'll go ahead and polish by adding docs, tests, and cleaning up

core/src/header.rs
```diff
@@ -230,6 +230,10 @@ impl<I, P> DataElement<I, P> {
         &self.value
     }
 
+    pub fn value_mut(&mut self) -> &mut Value<I, P> {
```
Enet4, line 233: This can make it too easy for the value to mismatch with the length described in the header. I was hoping not to introduce more gaps in which mismatches can be introduced (although the writer is safe from this and the case for text encoding conversion is known), which is why the `update_value` method was made instead. Is there something that you cannot perform with this 
method?

object/src/mem.rs
```diff
@@ -684,6 +684,10 @@ where
         self.entries.get(&tag)
     }
 
+    pub fn get_mut(&mut self, tag: Tag) -> Option<&mut InMemElement<D>> {
```
Enet4, line 687: This one is a no-go, because that makes it possible to change the tag and consequently diverge from the dictionary key.

If `update_value` is too limiting, the safe approach to modifying an entire data element is to remove it and then insert it back.

object/src/mem.rs
```diff
@@ -842,6 +846,20 @@ where
         }
     }
 
+    pub fn update_value_at(
```
Enet4, line 849: There is room for this method in the public API. Please don't forget to document it.

object/src/mem.rs
```diff
@@ -889,7 +907,7 @@ where
                 AttributeSelectorStep::Nested { tag, item } => {
                     let e = obj
                         .entries
-                        .get(tag)
+                        .get(&tag)
```
Enet4, line 910: This is just a false positive from rust-analyzer. https://github.com/rust-lang/rust-analyzer/issues/16657

```suggestion
                        .get(tag)
```

object/src/mem.rs
```diff
@@ -916,6 +934,150 @@ where
         unreachable!()
     }
 
+    pub fn entry_at(
```
Enet4, line 937: I wonder if we can reuse some of the implementation from the operations API rather than repeating most of it here.

object/src/mem.rs
```diff
@@ -916,6 +934,150 @@ where
...
+                                step_index: i as u32,
+                            })?;
+                }
+            }
+        }
+
+        unreachable!()
+    }
+
+    pub fn entry_at_mut(
```
Enet4, line 985: This one should not exist for the same reasons as `get_mut`.

Enet4 (changes requested):
Thank you for this proposal. Unfortunately, some methods of that nature were deliberately kept away from the object API until now because it opens room for inconsistencies in the object. More details inline.

core/src/header.rs
naterichman: IIRC I added it since it was needed in `value_at_mut`, but that one is being nuked anyway so this can go away as well

object/src/mem.rs
naterichman: Honestly totally forgot about that API.  As someone coming over from pydicom, I'm super used to just being able to mutate values at arbitrarily nested locations in dicoms, i.e. `dcm.PerFrameFunctionalGroups[0].ReferencedImageSequence[0].ReferencedFrameNumber = <val>`

From what I understand, I think all of the `update_value_at` functionality could be accomplished with the AttributeOp API, but (again coming from pydicom), I think having `entry_at` and `update_value_at` would be very useful shortcuts

Enet4: > I'm super used to just being able to mutate values at arbitrarily nested locations in dicoms, i.e.

As much as I am in favor of intuitive APIs (one of the aims of the project for that matter), creating one that is as intuitive as the one in pydicom would be highly unconventional in Rust. It may be feasible to write a macro which translates an expression in that format into an `AttributeOp::apply` call, although that too would require some heavy lifting.

> From what I understand, I think all of the update_value_at functionality could be accomplished with the AttributeOp API

It is possible to use one of the various actions to replicate the intended effect, although the API does not enable custom update operations, as that would severely change the nature of attribute actions. So `update_value_at` could be useful in this scenario.

naterichman:
Removed or switched to private the methods you mentioned would be dangerous.  Ended up keeping exposed

* `update_value_at`
* `entry_at`

Also called out in the docs for each function the alternatives (i.e. `apply` and `value_at`)

object/src/mem.rs
```diff
@@ -916,8 +926,180 @@ where
...
+        selector: impl Into<AttributeSelector>,
+        f: impl FnMut(&mut Value<InMemDicomObject<D>, InMemFragment>)
+    ) -> Result<(), AtAccessError> {
+        self.entry_at_mut(selector).map(|e| {
+            e.update_value(f)
+        })
+        .and_then(|_| {
+            self.len = Length::UNDEFINED;
+            Ok(())
+        })
```
Enet4, line 989: Clippy is advising the use of `map`.

```suggestion
        .map(|_| {
            self.len = Length::UNDEFINED;
        })
```

object/src/mem.rs
```diff
@@ -916,8 +926,180 @@ where
...
+                        })?;
+
+                    // get items
+                    let items = e.items_mut().with_context(|| NotASequenceSnafu {
+                        selector: selector.clone(),
+                        step_index: i as u32,
+                    })?;
+
+                    // if item.length == i and action is a constructive action, append new item
+                    obj =
```
Enet4, line 1085: This formatting looks a bit odd. Can you run rustfmt on the new code?

Enet4 (commented):
Thank you for the reiteration. Just some finishing touches and it should be ready.

Enet4 (approved):
OK, this is ready. Thank you @naterichman. :+1:

Issue #470, opened 2024-03-11
Support resolving BulkDataURIs


feliwir:
Right now `DicomJson` will fail when it sees that field.
Skipping these field is quite easy, but not really helpful. Expected behaviour:
- Parsing JSONs with `BulkDataURI` works without any issues
- An async function that allows retrieving particular fields (or all fields)
- Writing JSONs with `BulkDataURI` 

I tried adding a deserializer `DicomJson<MetaObject<InMemDicomObject>>`, but i'm not really sure where to persists the URIs.
Do you have any suggestions how to support this feature properly?

PR #472, opened 2024-03-11
[json] Skip bulk data URI when deserializing JSON

feliwir:
Relates to #470

### DIFF
--- a/json/src/de/mod.rs
+++ b/json/src/de/mod.rs
@@ -81,7 +81,7 @@ where
                 },
             ) = e;
             if bulk_data_uri.is_some() {
-                tracing::warn!("bulk data URI is not supported for InMemDicomObject");
+                tracing::warn!("bulk data URI is not supported for InMemDicomObject; skipping {}", tag);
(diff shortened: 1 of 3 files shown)


json/src/de/mod.rs
```diff
@@ -98,6 +109,10 @@ where
 struct JsonDataElement<D> {
     vr: VR,
     value: Value<InMemDicomObject<D>, InMemFragment>,
+    // TODO: we just ignore this when deserializing with
```
Enet4, line 112: This will help us track the concern. From what we have discussed on Zulip, this will very likely require us to create a new object implementation.

```suggestion
    // TODO(#470): we just ignore this when deserializing with
```

Enet4 (commented):
Looks good overall. Just adding a few recommendations inline.

Enet4 (approved):
Looks good to me. The next step would be a DICOM object implementation that can hole bulk data URIs, which can be worked on at its own pace. Thank you @feliwir!

PR #491, opened 2024-04-03
Update image lib

leonz:
Bumps the image-rs dependency to 25.1.

Two changes to features:
https://github.com/image-rs/image/commit/7c841c4a158fd6c732b6b48337659675703f6942
https://github.com/image-rs/image/commit/81b3fe66fba04b8b60ba79b3641826df22fca67e

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


leonz:
Hi @Enet4, would appreciate a review on this if possible.  This will enable us to bump the image library in our own project.

Enet4 (approved):
Thank you for the pull request. It's with good timing that we bump image to `0.25`, so that it is incorporated into the 0.7 milestone.

There is a known caveat that JPEG lossless image files will no longer work as input to `dicom-fromimage` (https://github.com/image-rs/image/pull/1877#issuecomment-1533492547), but this is a very minor situation which I believe no one is strongly depending on. JPEG baseline and JPEG lossless decoding will continue to depend on `jpeg-decoder` directly, meaning that we may have multiple JPEG decoder implementations in the final build for now. Maybe one day DICOM-rs can too move to `zune-jpeg` for decoding (https://github.com/etemesi254/zune-image/issues/144).

PR #500, opened 2024-04-18
Add promiscuous mode for ServerAssociation

### LINKED ISSUE #477: Storescp promiscuous mode
It is required to explicitly list all valid SOP class UIDs with `ServerAssociationOptions::with_abstract_syntax`. Would it be possible to support something similar to dcmtk storescp's "promiscuous" mode, so that unknown SOP classes are acceptable?

nickamzol:
This PR introduces a new `promiscuous` flag field for `ServerAssociationOptions` to allow unknown abstract syntaxes.

Fixes #477

### DIFF
--- a/ul/src/association/server.rs
+++ b/ul/src/association/server.rs
@@ -219,6 +219,8 @@ pub struct ServerAssociationOptions<'a, A> {
     max_pdu_length: u32,
     /// whether to receive PDUs in strict mode
     strict: bool,
+    /// whether to accept unknown abstract syntaxes
+    promiscuous: bool,
 }
 
 impl<'a> Default for ServerAssociationOptions<'a, AcceptAny> {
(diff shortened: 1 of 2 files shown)


nickamzol:
As requested, I've added both a positive and a negative test case for the new promiscuous mode.

In the positive test case, a server association with promiscuous mode enabled is established. The client requests `MR Image Storage`, which is not explicitly defined in the server association options as an accepted SOP class. The SCP will accept this as it is set to promiscuous mode.

In the negative test case, a server association with promiscuous mode disabled and `Ultrasound Image Storage` as an acceptable SOP class is established. The client requests `MR Image Storage` again, but this is expected to fail, because the SCP does not explicitly accept this SOP class and it's not set to promiscuous mode.

luckfamousa:
Great @nickamzol that is exactly what we need!

Enet4 (approved):
The code looks spot on, and it's well tested! :+1:

I will take care of expanding this capability to `dicom-storescp`. Thank you!

Issue #431, opened 2023-11-02
Adding private data elements (with creator)


matthiasg:
Is there any support in handling adding private elements ?

Enet4:
At the moment, you can retrieve and insert elements with private tags, but they are treated like any other elements, and the private creator needs to be added or inspected manually.
The data element dictionary can also be replaced with a custom implementation, which would enable the recognition of private tags by their alias at run-time. With some work in this direction, the ecosystem could support a more streamlined experience when working with these.

matthiasg:
Thanks for the quick feedback.

naterichman:
@Enet4 I'm interested in picking this up. Mostly for looking up private tags with a group, creator, and element mask, i.e. 

If there was:
```
(0x0009, 0x0010)    LO     "MY CREATOR"
(0x0009, 0x1001)    SH     "My value"    
```
you could look it up like `dcm.private_element(0x0009, "MY CREATOR", 0x01)`.   And similarly something for adding a private element creating/reserving a block if necessary:

```rust
dcm.put_private_element(
    0x0009,
    "MY CREATOR",
    0x01,
    VR.SH, 
    Value::Primitive(
        PrimitiveValue::from("My Value".to_string())
    )
)
```

Just looking for your thoughts on design.  LMK!

Enet4:
Thank you for the initiative, @naterichman! From what I know about private data elements, that API seems to go in the right direction, at least for the in-memory object implementation.

I've been thinking about how we would also incorporate this capability onto the Operations API (the attribute selectors only recognize tags and not private groups), and I would appreciate having more people thinking about this.

naterichman:
@matthiasg Getting back to this if you have a chance to test this out with 0.7.0 and have feedback on the interface, let me know!

PR #508, opened 2024-04-20
ENH: Add lookup and putting of private elements

naterichman:
Fairly simple lookup and adding of private tags with creator for `InMemDicomObject` as briefly discussed in #431

### DIFF
--- a/object/src/mem.rs
+++ b/object/src/mem.rs
@@ -41,7 +41,7 @@ use dicom_core::ops::{
 };
 use itertools::Itertools;
 use smallvec::SmallVec;
-use snafu::{OptionExt, ResultExt};
+use snafu::{ensure, OptionExt, ResultExt};
 use std::borrow::Cow;
 use std::fs::File;
 use std::io::{BufRead, BufReader, Read};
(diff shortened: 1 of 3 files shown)


Enet4 (commented):
Hey @naterichman, I made a quick code review, but it seems the branch also needs a rebase.

naterichman:
I’ll rebase.  I could have sworn I checkout from the master branch.  I’m still getting used to contributing to a repo that’s not my own on GitHub (use gitlab for work)

Anyway I was thinking, regarding your comment about the attribute selector API on #431 .  Do you want to keep methods like this and add supporting methods for the attribute selector API or would you be more interested in merging them into one single API for private tag access? 

I was thinking we could expand the `AttributeSelector` enum to include a `PrivateGroup` and `PrivateElement` definition and these methods would take the corresponding selectors instead of group and element separately.  LMK your thoughts

Enet4:
> I’ll rebase. I could have sworn I checkout from the master branch. I’m still getting used to contributing to a repo that’s not my own on GitHub (use gitlab for work)

It's OK, it's easy to miss the fact that one has to do both

1. fetch the latest contributions from the upstream remote: `git fetch origin`;
2. rebase against it in specific: `git rebase origin/master` instead of `git rebase master` unless the local `master` branch has also been updated.

> Anyway I was thinking, regarding your comment about the attribute selector API on #431 . Do you want to keep methods like this and add supporting methods for the attribute selector API or would you be more interested in merging them into one single API for private tag access?

We can keep them separate. Each API usually stands on its own, meaning that one can do most of the things without the operations API. In other words, we can start with what we have in this PR, which is provide these methods to the in-mem DICOM object impl.

> I was thinking we could expand the `AttributeSelector` enum to include a `PrivateGroup` and `PrivateElement` definition and these methods would take the corresponding selectors instead of group and element separately. LMK your thoughts

I'm not entirely sure how that would look like, but I suggest we defer to a separate PR. If #508 is cleaned up and made ready within the next day or two, it might just be ready in time for the upcoming major release.

Enet4 (commented):
One more review iteration, mostly to emphasize that new parts of the public API should be documented, and to take care of clippy lints warnings that may emerge.

naterichman:
Most of that is a big il’ facepalm.  I’ll get this done today!

object/src/mem.rs
```diff
@@ -758,6 +785,45 @@ where
...
+    ///     0x0009,
+    ///     "CREATOR 1",
+    ///     0x02,
+    ///     VR::DS,
+    ///     PrimitiveValue::Str("1.0".to_string()),
+    /// )
+    /// .unwrap();
+    /// assert_eq!(
+    ///     ds.private_element(0x0009, "CREATOR 1", 0x02)
+    ///         .unwrap()
```
Enet4, line 813: Examples should prefer `?` for raising errors.

naterichman: So I found [this](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#using--in-doc-tests), but I'm not sure how to get it to work without using `map_err` since I have two different error types (`PrivateElementError` and `ConvertValueError`). Any suggestions?

naterichman: Ended up getting it to work with `# Ok::<(), Box<dyn Error>>(())`.

Another question for you.  I wanted to format the code in the examples in doctests.  I found I can run `cargo fmt -- --config format_code_in_doc_comments=true object/src/mem.rs` but that this reformatted much more than I touched.  Do you have a normal way you do this?

Enet4: In those cases I just use `Box<dyn std::error::Error>` as the almost-universal common ground.

Enet4 (approved):
Just added a few minor tweaks to the documentation in the last few commits.

The method signatures are good and the tests seem to cover the intended use cases well. Future work will hopefully tackle a more generalised awareness of private elements in the rest of the ecosystem, but it's great to have this PR as the first step towards it. Much appreciated! :+1:

PR #514, opened 2024-05-01
Fix error when decoding data without VoiLut with GDCM

dougyau:
When a decoding a dicom without voilut, it will throw an error. This fix handles this situation and defaults to None.


Enet4 (approved):
Looks OK! Thank you @dougyau.

PR #518, opened 2024-05-15
[toimage] basic bulk conversion

PierreBou91:
Hi,

I worked on this a few month ago (#397) but I lacked time to finish it. Today I wanted to finish it and somehow messed my merge so I closed the other PR and reopened this clean one from main/master.

The basic functionalities of bulk conversion work fine while retaining the previous behavior.

The user can pass a single file, a folder or several files.
If a single file is passed, the user can specify an output file name with the extension.
If there are multiple files (or several files collected in the folder) the user can optionally pass an output directory where the images will be stored. The user can also optionally pass a target extension for the images.
If a folder is passed, it can be searched recursively for dicom files.

The function `convert_single_file()` is a bit clumsy because we wanted to retain the previous behavior of being able to rename the output file when passing a single file.

I could add some other functionalities but I wanted to check with you first if it was worth doing, notably:
- Handling name collisions
- Giving the possibility to convert every frames of a multiframe dicom (this would mean changing the current `frame_number` behavior)

Please tell me if this is what you had in mind in #343 and what should be modified to be able to merge.

Thanks

### DIFF
--- a/toimage/README.md
+++ b/toimage/README.md
@@ -11,19 +11,21 @@ This tool is part of the [DICOM-rs](https://github.com/Enet4/dicom-rs) project.
 ## Usage
 
 ```none
-    dicom-toimage [FLAGS] [OPTIONS] <file>
+    dicom-toimage [OPTIONS] [FILES]...
 
-FLAGS:
-        --16bit      Force output bit depth to 16 bits per sample
-        --8bit       Force output bit depth to 8 bits per sample
-    -h, --help       Prints help information
-    -V, --version    Prints version information
-    -v, --verbose    Print more information about the image and the output file
+Arguments:
+  [FILES]...  Path to the DICOM file to convert
 
-OPTIONS:
-    -F, --frame <frame-number>    Frame number (0-indexed) [default: 0]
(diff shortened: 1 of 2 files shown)


Enet4 (changes requested):
Thank you for reviving this PR. I found one issue, but it should be ready after the fix.

PierreBou91:
Hello,

Thanks for taking the time to review and for your feedback.

Nice catch indeed, however I'm confused because I have extensionless files in my test set and it doesn't panic while it should (at least I thought unwrapping a None value was panicking).

In any case, your fix works perfectly fine so I'll commit it without `.as_deref()` as clippy suggested:
```
warning: derefed type is same as origin
   --> toimage/src/main.rs:251:8
    |
251 |     if output.extension().as_deref() != Some("dcm".as_ref()) && !output_is_set {
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `output.extension()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_option_as_deref
    = note: `#[warn(clippy::needless_option_as_deref)]` on by default
    ```

Issue #473, opened 2024-03-13
Non-deterministic behavior in dicom-storescu function check_presentation_contexts


jennydaman:
https://github.com/Enet4/dicom-rs/blob/dbd41ed3a0d1536747c6b8ea2b286e4c6e8ccc8a/storescu/src/main.rs#L213-L235

I am trying to use the code of dicom-storescu, though some non-deterministic behavior is making my tests flaky.

Get some example data here: https://github.com/FNNDSC/SAG-anon

I edited storescu/src/main.rs for debugging:

```patch
diff --git a/storescu/src/main.rs b/storescu/src/main.rs
index ca29e91..daecb3b 100644
--- a/storescu/src/main.rs
+++ b/storescu/src/main.rs
@@ -224,6 +224,8 @@ fn run() -> Result<(), Error> {
                 }
                 file.pc_selected = Some(pc);
                 file.ts_selected = Some(ts);
+                dbg!(&file.ts_selected);
+                std::process::exit(0);
             }
             Err(e) => {
                 error!("{}", Report::from_error(e));
```

In one terminal window, I am running

```shell
cargo run --bin dicom-storescp
```

And in another, I am running

```shell
cargo run --bin dicom-storescu 127.0.0.1:11111 ~/fnndsc/public_data/SAG-anon/0002-1.3.12.2.1107.5.2.19.45152.2013030808110261698786039.dcm
```

Here's the output from several back-to-back repeated runs:

```
$ ./target/debug/dicom-storescu 127.0.0.1:11111 ~/fnndsc/public_data/SAG-anon/0002-1.3.12.2.1107.5.2.19.45152.2013030808110261698786039.dcm
[storescu/src/main.rs:227:17] &file.ts_selected = Some(
    "1.2.840.10008.1.2.1",
)
$ ./target/debug/dicom-storescu 127.0.0.1:11111 ~/fnndsc/public_data/SAG-anon/0002-1.3.12.2.1107.5.2.19.45152.2013030808110261698786039.dcm
[storescu/src/main.rs:227:17] &file.ts_selected = Some(
    "1.2.840.10008.1.2",
)
$ ./target/debug/dicom-storescu 127.0.0.1:11111 ~/fnndsc/public_data/SAG-anon/0002-1.3.12.2.1107.5.2.19.45152.2013030808110261698786039.dcm
[storescu/src/main.rs:227:17] &file.ts_selected = Some(
    "1.2.840.10008.1.2.1",
)
[...]

PR #520, opened 2024-06-02
[storescu] prefer file's transfer syntax

PierreBou91:
Hi,

As addressed in #473 , I have slightly adapted the [`check_presentation_context()`](https://github.com/Enet4/dicom-rs/blob/master/storescu/src/main.rs#L563) function to return early if there is an exact match between the file's TS and the TS from the presentation contexts slice that is passed to the function. Otherwise, the previous logic applies. When testing, the selected_ts is consistently the same as the file's TS.

Is it the expected behavior for storescu ?

Additionally, you commented:
> (...)
> The other criterion currently written is to transcode and accept Explicit VR Little Endian otherwise (only when `--never-transcode` is not enabled).
> (...)
_Originally posted by @Enet4 in https://github.com/Enet4/dicom-rs/issues/473#issuecomment-2117643092_

I am uncertain how to address this point and would appreciate some guidance please.

Have a nice day !

### DIFF
--- a/storescu/src/main.rs
+++ b/storescu/src/main.rs
@@ -570,8 +570,13 @@ fn check_presentation_contexts(
         .with_context(|| UnsupportedFileTransferSyntaxSnafu {
             uid: file.file_transfer_syntax.to_string(),
         })?;
-    // if destination does not support original file TS,
-    // check whether we can transcode to explicit VR LE
+
+    // Try to find an exact match for the file's transfer syntax first
+    let exact_match_pc = pcs.iter().find(|pc| pc.transfer_syntax == file_ts.uid());
+
+    if let Some(pc) = exact_match_pc {
+        return Ok((pc.clone(), pc.transfer_syntax.clone()));
+    }
 
     let pc = pcs.iter().find(|pc| {
         // Check support for this transfer syntax.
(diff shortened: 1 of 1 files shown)


Enet4:
> Is it the expected behavior for storescu ?

That seems to do the trick. :+1:

> Additionally, you commented:
> 
> > (...)
> > The other criterion currently written is to transcode and accept Explicit VR Little Endian otherwise (only when `--never-transcode` is not enabled).
> > (...)
> > _Originally posted by @Enet4 in [#473 (comment)](https://github.com/Enet4/dicom-rs/issues/473#issuecomment-2117643092)_
> 
> I am uncertain how to address this point and would appreciate some guidance please.

This should be already in place by the logic that follows.

In the face of uncertainty, we could begin the process of moving key components of `dicom-storescu` into a new library `dicom_storescu` and write tests for it.

PierreBou91:
> That seems to do the trick. 👍

Ok great, thanks for the review !

> In the face of uncertainty, we could begin the process of moving key components of `dicom-storescu` into a new library `dicom_storescu` and write tests for it.

That sounds like a good idea ! I can start working on it in this PR if you agree ?

I think I will try to mimic the structure of the [dump](https://github.com/Enet4/dicom-rs/blob/master/dump/src/lib.rs) lib. If you think of other resources that might be relevant I'll gladly take a look.

I might have some questions about tests best practices later but I need to take some time to research a bit and try things by myself first. So I'll get back with a first simple commit and probably some questions as soon as I have time !

Enet4:
Hello @PierreBou91! I don't know about the progress of the storescu lib-bin split, but given the relatively small and useful changes here that can be tested ad-hoc, it would be nice if they were made ready for review and the planned refactor deferred to a new PR.

PierreBou91:
Hello @Enet4 ! Sorry I have not started the lib split yet, work has been a bit more intense that I was expecting lately.

You're right, I'll put this PR as ready for review and open a new one at a later date to adress the split when I have a moment !

Enet4 (approved):
OK, these changes provide an acceptable means of making presentation context selection deterministic: It will prefer the same transfer syntax as the file first, then (only if transcoding is supported) Explicit VR LE second, then Implicit VR LE third.

Much appreciated! :+1:

Issue #522, opened 2024-06-14
findscu not working when dataset is included in C-FIND-RSP


naterichman:
I've been dipping my toes into learning dimse by trying to write my own custom connector to a real pacs system, mostly using the code from `findscu` as a template.  I found that my code wasn't working and I tried `dicom-findscu` itself and that also had strange behavior, whereas DCMTKs `findscu` returned immediately.

(the pacs is not my own system so I'm redacting everything just to be safe)

```bash
$ dicom-findscu --calling-ae-title <calling-ae-title> --called-ae-title <called-ae-title> -S -q AccessionNumber="<redacted>" -q SeriesInstanceUID <ip>:<port> -v
2024-06-14T23:12:43.614726Z  INFO dicom_findscu: Establishing association with '<ip>:<port>'...
2024-06-14T23:12:43.685326Z  INFO dicom_findscu: Association established
2024-06-14T23:12:43.685372Z DEBUG dicom_findscu: Transfer Syntax: Explicit VR Little Endian
2024-06-14T23:12:43.685436Z DEBUG dicom_findscu: Sending query (128 B)...
2024-06-14T23:12:43.685476Z DEBUG dicom_findscu: Awaiting response...
Match #0 Response command:
(0000,0000) CommandGroupLength           UL (1,  4 bytes): 76
(0000,0002) AffectedSOPClassUID          UI (1, 28 bytes): "1.2.840.10008.5.1.4.1.2.1.1"
(0000,0100) CommandField                 US (1,  2 bytes): 32800
(0000,0120) MessageIDBeingRespondedTo    US (1,  2 bytes): 1
(0000,0800) CommandDataSetType           US (1,  2 bytes): 258
(0000,0900) Status                       US (1,  2 bytes): 65280
2024-06-14T23:12:43.812867Z DEBUG dicom_findscu: Operation pending: ff00
------------------------ Match #0 ------------------------
(0000,0000) CommandGroupLength           UN (1, 76 bytes): [00, 00, 02, 00, 1C, 00, 00, ...
# Just hangs here indefinitely
```

I ended up finding during debugging that the dataset was included in the same `PDU` as another `PDataValue` (I think I'm using these terms correctly), so I changed the code as follows:
```diff
diff --git a/findscu/src/main.rs b/findscu/src/main.rs
index 6df7325..0601968 100644
--- a/findscu/src/main.rs
+++ b/findscu/src/main.rs
[...]

PR #528, opened 2024-06-27
Fix Harmowy worklist server

qarmin:
Fixes https://github.com/Enet4/dicom-rs/discussions/527
Fixes https://github.com/Enet4/dicom-rs/issues/522

After several attempts, it turned out that Harmony sends all the data at once, unlike orthanc and https://dicomserver.co.uk/

Now downloading patient data from Harmony works fine

Screenshot from 2024-06-27 10-41-37

### DIFF
--- a/findscu/src/main.rs
+++ b/findscu/src/main.rs
@@ -268,6 +268,17 @@ fn run() -> Result<(), Error> {
 
         match rsp_pdu {
             Pdu::PData { data } => {
+                if data.is_empty() {
+                    error!("Empty PData response");
+                    break;
+                } else if ![1, 2].contains(&data.len()) {
+                    warn!(
+                        "Unexpected number of PDataValue parts: {} (allowed 1 or 2)",
+                        data.len()
+                    );
+                    break;
+                }
+
                 let data_value = &data[0];
 
                 let cmd_obj = InMemDicomObject::read_dataset_with_ts(
@@ -300,8 +311,12 @@ fn run() -> Result<(), Error> {
                     }
 
                     // fetch DICOM data
-
-                    let dcm = {
+                    // Some worklist servers sends both command and data in the same PData
+                    // So there is no need to download another PData
+                    let dcm = if let Some(second_pdata) = data.get(1) {
+                        InMemDicomObject::read_dataset_with_ts(second_pdata.data.as_slice(), ts)
+                            .whatever_context("Could not read response data set")?
+                    } else {
                         let mut rsp = scu.receive_pdata();
                         let mut response_data = Vec::new();
                         rsp.read_to_end(&mut response_data)


Enet4 (approved):
I had the chance to test it out with some other DICOM SCPs, no issues were detected. The code is also sound.

Thank you very much for the patch! :+1:

findscu/src/main.rs
```diff
@@ -268,6 +268,17 @@ fn run() -> Result<(), Error> {
 
         match rsp_pdu {
             Pdu::PData { data } => {
+                if data.is_empty() {
+                    error!("Empty PData response");
+                    break;
+                } else if ![1, 2].contains(&data.len()) {
```
naterichman, line 274: @Enet4 thanks for mentioning my issue so I can see this!

I know this is already merged, but where is this specified in the standard,  I see this section, but that seems to indicate that we _should_ allow multiple command and data fragments within the same PDU, just not send them:

> It is strongly recommended that two consecutive PDVs in the same P-DATA Request primitive (therefore containing fragments of the same message using the same Presentation Context ID) do not contain two message Control Headers with the same type (Command or Data). These should have been combined in a single PDV by the sender. However, receivers must be able to receive and process such PDVs.

https://dicom.nema.org/medical/dicom/current/output/chtml/part08/chapter_E.html

qarmin: Do you know of any server that would send such data, to be able to test potential fix?

I know that this PR not provided fix for all possible situations in dicom standard but I was only able to test 1 command + 1 data message

naterichman: I do not know of an open source server that does this.  I've been writing a query-retrieve program against (I think) a GE PACS server deployed as the main PACS for a hospital, and it _sometimes_ does that.  I was meaning to contribute something back here once I have the time.  Basically what I came up with was like this:

```rust
pub fn read_pdata(
    data: Vec<PDataValue>,
    ts: &TransferSyntax,
) -> Result<(Option<InMemDicomObject>, Option<InMemDicomObject>), Error> {
    let mut ds_buffer = Vec::new();
    let mut cs_buffer = Vec::new();
    for pdata in data {
        match pdata.value_type {
            PDataValueType::Command => {
                cs_buffer.extend(pdata.data);
            }
            PDataValueType::Data => {
                ds_buffer.extend(pdata.data);
            }
        }
    }
    let data_set = if ds_buffer.len() > 0 {
        Some(
            InMemDicomObject::read_dataset_with_ts(&ds_buffer[..], ts)
                .whatever_context("Could not parse dicom dataset")?,
        )
    } else {
        None
    };
    let command_set = if cs_buffer.len() > 0 {
        Some(
            InMemDicomObject::read_dataset_with_ts(
                &cs_buffer[..],
                &entries::IMPLICIT_VR_LITTLE_ENDIAN.erased(),
            )
            .whatever_context("Could not parse dicom command set")?,
        )
    } else {
        None
    };
    Ok((command_set, data_set))
}
``` 

And then in my downstream findscu and movescu I use
```rust    
...
    loop {
        let rsp = scu.receive().unwrap();
        match rsp {
            Pdu::PData { data } => {
                let (maybe_cmd, _maybe_ds) =
                    read_pdata(data, ts).whatever_context("Failed to read response data")?;
                let cmd = maybe_cmd.whatever_context("No command set preset in PDU")?;
...
```

Enet4: Sorry, I ended up accepting because it covers all the reasonably probable scenarios I could think of:

- 1 PDV per PData (Command/Data interleaved)
- 2 PDVs in a single PData (1 command and 1 data)
- 0 PDVs (?)
- 1 Command PDV reporting status code 0 (no results)

I did not envision the scenario of having multiple PDVs of the same type, though it's good to know that it might happen in some exotic implementation. I was also imagining a scenario of a single PData holding more than one Cmd-Data pair, but it did not seem to be very likely (though it seems to be valid per the standard!).

Enet4: Of course, there is nothing preventing us from reiterating on `findscu` further. For example, the idea of splitting these tools into a library and binary is worth conceiving at some point in the project.

naterichman: Oh yea, no problem for sure. Sometime in the next few weeks I will probably contribute a bit of that (modularizing find-scu) and potentially this multi-data-pdv function as well.

PR #530, opened 2024-07-04
MAIN: Add read/write timeout options for ClientAssociation

naterichman:
Small MR to add read/write timeouts for the TCPStream on a ClientAssociation object.  

I am using the client extensively and find that some pacs servers (presumably when under load) are taking 1000+s to respond to a c-move.

Additionally when I get around to it I will add over my c-move implementation.

As a side-note I've used it as a learning point for async rust and have made processes as async as possible (i.e just `spawn_blocking` as little as I can) and that has been interesting, but it would definitely be great to make the Client and Server async compatible themselves.

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -28,13 +28,6 @@ use super::{
     uid::trim_uid,
 };
 
-#[derive(Debug)]
-pub enum ConnectOp {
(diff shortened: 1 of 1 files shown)


Enet4 (approved):
Thank you @naterichman!

PR #536, opened 2024-07-09
Default to None if voilut is not present

dougyau:
This is a similar fix as applied on #514

### DIFF
--- a/pixeldata/src/attribute.rs
+++ b/pixeldata/src/attribute.rs
@@ -122,7 +122,7 @@ pub fn voi_lut_function<D: DataDictionary + Clone>(
             .collect::<Result<Vec<_>>>()?;
         Ok(Some(res))
     } else {
-        Ok(None)
+        Ok(Some(vec!["LINEAR".to_string()]))
     }
 }
 
--- a/pixeldata/src/lib.rs
+++ b/pixeldata/src/lib.rs
@@ -1934,7 +1934,7 @@ impl ImagingProperties {
         let rescale_intercept = rescale_intercept(obj);
         let rescale_slope = rescale_slope(obj);
         let number_of_frames = number_of_frames(obj).context(GetAttributeSnafu)?;
-        let voi_lut_function = voi_lut_function(obj).context(GetAttributeSnafu)?;
+        let voi_lut_function = voi_lut_function(obj).unwrap_or(None);
         let voi_lut_function: Option<Vec<VoiLutFunction>> = voi_lut_function.and_then(|fns| {
             fns.iter()
                 .map(|v| VoiLutFunction::try_from((*v).as_str()).ok())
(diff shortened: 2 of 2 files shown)


Enet4 (changes requested):
Thank you for looking into this. Can you provide a sample file that tripped this line? `voi_lut_function` should not error when the attribute is missing entirely, but rather when it has an invalid value. I suspect that this was caused by VOI LUT Function being present in the data set, but with an empty value.

At this time I would rather seek to increase the resilience of `voi_lut_function` against empty values (treating them as they were not present at all) than change places where the function is called.

Here is a test function that you can add to `attribute::tests`:

```rust

    #[test]
    fn get_voi_lut_function() {
        use super::voi_lut_function;

        let mut obj = FileDicomObject::new_empty_with_meta(
            FileMetaTableBuilder::new()
                .transfer_syntax(uids::EXPLICIT_VR_LITTLE_ENDIAN)
                .media_storage_sop_class_uid(uids::ENHANCED_MR_IMAGE_STORAGE)
                .media_storage_sop_instance_uid("2.25.145929179730251416957282651365760465911")
                .build()
                .unwrap(),
        );

        // Returns None if not present
        assert_eq!(voi_lut_function(&obj).unwrap(), None);

        obj.put(DataElement::new(
            tags::VOILUT_FUNCTION,
            VR::CS,
            PrimitiveValue::Empty,
        ));
        // Returns None if empty
        assert_eq!(voi_lut_function(&obj).unwrap(), None);

        obj.put(DataElement::new(tags::VOILUT_FUNCTION, VR::CS, "LINEAR"));

        // Returns the value
        assert_eq!(
            voi_lut_function(&obj).unwrap(),
            Some(vec!["LINEAR".to_string()])
        );

        obj.remove_element(tags::VOILUT_FUNCTION);

        // per-frame functional groups
        obj.put(DataElement::new(
            tags::PER_FRAME_FUNCTIONAL_GROUPS_SEQUENCE,
            VR::SQ,
            DataSetSequence::from(vec![InMemDicomObject::from_element_iter([
                DataElement::new(tags::VOILUT_FUNCTION, VR::CS, "LINEAR_EXACT"),
            ])]),
        ));

        assert_eq!(
            voi_lut_function(&obj).unwrap(),
            Some(vec!["LINEAR_EXACT".to_string()])
        );
    }
```

dougyau:
According to the [standard](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.2.html#table_C.11-2) it says the VOI LUT function is optional, and when not present it should default to LINEAR. 

I have updated the PR to reflect this.

What do think of how I updated the PR. It should default to LINEAR in case the tag is not present.

Enet4:
> What do think of how I updated the PR. It should default to LINEAR in case the tag is not present.

Yeah, we could go in that direction for this function here. In that case we never return `Ok(None)`, so this function could return a `Result<Vec<String>>` instead of a `Result<Option<Vec<String>>>`. This is safe to do now because the function is not part of the public API.

Enet4 (approved):
I will approve this and defer optimizations to other pull requests. Thanks again! :+1:

PR #541, opened 2024-07-14
Infer Media Storage SOP Class UID and Media Storage SOP Instance UID from the SOP Class UID and SOP Instance UID when missing

### LINKED ISSUE #538: open_file support for inferring MediaStorageSOPClassUID and MediaStorageSOPInstanceUID from SOP Class UID and SOP Instance UID
Hello,

We have certain files that are missing MediaStorageSOPClassUID and MediaStorageSOPInstanceUID, however, they do have values for SOP Class UID and SOP Instance UID. Do you happen to know if there is a way to infer them from SOP Class UID and SOP Instance UID when using `open_file`?

If there isn't a way I'd be glad to send out a PR - open to suggestions on where the logic should go.

Thanks,
Chris

chrisbll971:
Resolves #538 

- Default Media Storage SOP Class UID and Media Storage SOP Instance UID to empty string when not available in the DICOM meta information table.
- In `open_file_with_all_options` attempt to infer the Media Storage SOP Class UID and Media Storage SOP Instance UID from the SOP Class UID and SOP Instance UID tag values if the Media Storage SOP Class UID and Media Storage SOP Instance UID are set to an empty string

### DIFF
--- a/object/src/lib.rs
+++ b/object/src/lib.rs
@@ -225,6 +225,11 @@ pub enum ReadError {
         #[snafu(backtrace)]
         source: crate::meta::Error,
     },
+    ParseSopAttributes {
+        #[snafu(source(from(dicom_core::value::ConvertValueError, Box::from)))]
+        source: Box<dicom_core::value::ConvertValueError>,
+        backtrace: Backtrace,
+    },
     #[snafu(display("Could not create data set parser"))]
     CreateParser {
         #[snafu(backtrace)]
--- a/object/src/lib.rs
+++ b/object/src/lib.rs
@@ -225,7 +225,8 @@ pub enum ReadError {
         #[snafu(backtrace)]
         source: crate::meta::Error,
     },
-    ParseSopAttributes {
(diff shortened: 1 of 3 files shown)


object/src/lib.rs
```diff
@@ -225,6 +225,11 @@ pub enum ReadError {
         #[snafu(backtrace)]
         source: crate::meta::Error,
     },
+    ParseSopAttributes {
```
Enet4, line 228: As each error pertains to a single SOP attribute, the singular form (`ParseSopAttribute`) might be a better name.

Don't forget to include a suitable display message!

Enet4 (commented):
Thank you for working on this! This is almost ready to enter. Please see the suggestions inline.

object/src/lib.rs
chrisbll971: That's a good point. Released an updated version to address these

Enet4 (approved):
Great! Thank you very much!

PR #542, opened 2024-07-14
ENH: Add POC async implementation, example using storescp

naterichman:
POC async implementation.

* Created a feature flag `tokio` which controls the importing of `read_pdu` and `write_pdu` from either `[read|write].rs` or `[read_nonblocking|write_nonblocking].rs` 
* Created feature gated implementations of the methods in `Client` and `Server` which depend on using `read_pdu`/`write_pdu` or any read/write to the socket themselves.

A couple other notes:
* We could consider using the [`maybe_async`](https://docs.rs/maybe-async/latest/maybe_async/) crate.
    * Since most of the code in read_pdu/write_pdu is copied for the sync/async implementation I thought this could be nice to use, but I found that I'd have to extend the macro to adjust the trait bounds for the sync/async implementation and I couldn't figure out the best way to do that
* I had to change a little bit of the internal API for `write_chunk_u16` and `write_chunk_u32`
* For now I completely ignored the `PDataReader` and `PDataWriter` since I'm not exactly clear where they should be used or not used (storescp doesn't use those at all, and in my implementation of c-find I didn't end up using either at all either)
* I just added the timeouts in another MR, but it seems like the concept of socket level timeouts doesn't exist in an async context?  So Future work will be adding the timeout to the individual methods like `send`, `receive`, `establish`, etc. via something like `tokio::time::timeout`

Also note this is meant to just be one possible implementation, and I totally expect completely rewriting this MR however you want it to look! Thanks in advance!


naterichman:
I've been doing some more reading up, and unfortunately it seems like the options are 

1. Repeat a bunch of code either manually or with a nice macro to write truly sync and async versions of `ul`, but that would (and the current PR) break users expecting `cargo install dicom-ul --all-features` to work.
2. Make the code async only, and expose a blocking client which just calls `block_on` for the blocking client

Enet4:
Much appreciated!

> I've been doing some more reading up, and unfortunately it seems like the options are
> 
> 1. Repeat a bunch of code either manually or with a nice macro to write truly sync and async versions of `ul`, but that would (and the current PR) break users expecting `cargo install dicom-ul --all-features` to work.
> 2. Make the code async only, and expose a blocking client which just calls `block_on` for the blocking client

There ought to be a bit of redundancy in this process unfortunately. Whether to go async or not requires a different function "color" which affects its inner workings altogether.

At best, there is a way to reduce the amount of redundancy, which is to write non-blocking _and_ non-polling implementations of the PDU readers. While this can be a bit trickier than the existing logic at `read_pdu`, this would couple nicely with the concept of [framing](https://tokio.rs/tokio/tutorial/framing) without requiring users to depend on tokio if they only intend to use the blocking API.

Let me know if more guidance is needed here.

naterichman:
Let me make sure I'm understanding where you're going.  For the reader implementation, it would be something like:

* Rename `read_pdu` to something like `parse_pdu` and leave the trait bound as `R: Read`, keeping the logic essentially as is but changing the return type to something like `Result<Option<Pdu>>` instead of `Result<Pdu>`
* Making an `async receive` and leaving the existing `receive` method on the `ClientAssociation`, changing the logic to something like:
```rust
pub async fn receive(&mut self) -> Result<Pdu> {
    loop {

        if let Some(pdu) = parse_pdu(&mut self.buffer)? {
            return Ok(Some(pdu));
        }

        if 0 == self.stream.read_buf(&mut self.buffer).await? {
            if self.buffer.is_empty() {
                return Ok(None);
            } else {
                return Err("connection reset by peer".into());
            }
        }
    }
}
```

And similar for `pub fn receive`? (Mostly copy/pasting code from the framing article you linked :smile: )

Enet4:
> Let me make sure I'm understanding where you're going. For the reader implementation, it would be something like:
> 
> * Rename `read_pdu` to something like `parse_pdu` and leave the trait bound as `R: Read`, keeping the logic essentially as is but changing the return type to something like `Result<Option<Pdu>>` instead of `Result<Pdu>`

That could work if we use a reader that keeps a buffer of all partial data, so that it can be read multiple times until a complete frame is available. In practice, we may be better off making this function receive a value with a better trait bound than `Read`, providing a peekable source of bytes which can be extended as more data arrives from the network, and only consumes the front bytes when individual frames comprising those bytes have been fully processed. The `bytes` crate (part of the tokio ecosystem) offers some nice data types and traits for this, so it may be worth trying it out.

naterichman:
I'm not really following, sorry!  What kind of trait bound might be better, something like `Buf` or `BufMut`?  At that point why not just make `read_pdu` take `&mut BytesMut` or `Cursor<&[u8]>`. Any chance you could provide a little code snippet of something you had in mind so I know which direction to go?

Enet4:
> What kind of trait bound might be better, something like `Buf` or `BufMut`? At that point why not just make `read_pdu` take `&mut BytesMut` or `Cursor<&[u8]>`.

Both `BytesMut` and `Cursor<&[u8]>` implement `Buf`, but they assume a contiguous portion of bytes. The way I see it, the trait would make room for memory usage optimizations by allowing us to read from chained portions of data (thinking network packet payloads).
 
> Any chance you could provide a little code snippet of something you had in mind so I know which direction to go?

I would just try and see if you can tweak the implementation so that the function signature is `fn read_pdu(x: impl Buf) -> Result<Option<Pdu>>`. `BufMut` would offer write access, which we do not need for reading.

ul/src/association/client.rs
```diff
@@ -773,11 +1014,56 @@ impl ClientAssociation {
...
+    }
+
+    #[cfg(not(feature = "tokio"))]
     /// Read a PDU message from the other intervenient.
     pub fn receive(&mut self) -> Result<Pdu> {
         read_pdu(&mut self.socket, self.requestor_max_pdu_length, self.strict).context(ReceiveSnafu)
     }
+    #[cfg(feature = "tokio")]
+    /// Read a PDU message from the other intervenient.
+    pub async fn receive(&mut self) -> Result<Pdu> {
```
naterichman, line 1041: Calling out the new `receive` method for review

ul/src/pdu/reader.rs
```diff
@@ -1,115 +1,13 @@
...
-    MissingTransferSyntax { backtrace: Backtrace },
-}
-
-pub type Result<T> = std::result::Result<T, Error>;
+pub type Result<T> = std::result::Result<T, ReadError>;
 
-pub fn read_pdu<R>(reader: &mut R, max_pdu_length: u32, strict: bool) -> Result<Pdu>
-where
-    R: Read,
+pub fn read_pdu(mut buf: impl Buf, max_pdu_length: u32, strict: bool) -> Result<Option<Pdu>>
```
naterichman, line 10: The changes here are basically just using the methods from `Buf` instead of the methods from the `ReadExt` trait from the `byteorder` crate.  Ended up removing `read_n` since the `copy_to_bytes` method basically does that

ul/src/association/client.rs
```diff
@@ -773,11 +1014,56 @@ impl ClientAssociation {
...
+            }
+            .fail();
+        }
+        self.socket
+            .write_all(&self.buffer)
+            .await
+            .context(WireSendSnafu)
+    }
+
+    #[cfg(not(feature = "tokio"))]
```
naterichman, line 1034: TODO need to figure out how the sync `receive` method would look

naterichman (commented):
Called out a few specific changes.  I have it working again as POC for the storescp command, still need to figure out how the sync client would look.  LMK your thoughts/if this is the direction you had in mind

naterichman:
Okay, I'm happier with the implementation now.  I have a POC fully working for `storescp`, but I'm having trouble getting the `AsyncPDataWriter` working for use with `storescu`.  Currently I get the association working fine, but the `send_pdata` section is not working.

<details>

Orthanc logs:
```
======================= END A-ASSOCIATE-AC ======================
T0807 10:13:53.602828          DICOM-1 CommandDispatcher.cpp:760] (dicom) Received Command:
===================== INCOMING DIMSE MESSAGE ====================
Message Type                  : C-STORE RQ
Presentation Context ID       : 1
Message ID                    : 1
Affected SOP Class UID        : MRImageStorage
Affected SOP Instance UID     : 1.3.6.1.4.1.14519.5.2.1.8421.4004.102660487069874494712993403337
Data Set                      : present
Priority                      : medium
======================= END DIMSE MESSAGE =======================
I0807 10:13:53.602850          DICOM-1 main.cpp:353] Incoming Store request from AET STORE-SCU on IP 127.0.0.1, calling AET ANY-SCP
E0807 10:13:53.603049          DICOM-1 StoreScp.cpp:273] Store SCP Failed: DIMSE Failed to receive message
```

Wireshark capture:
image

Can attach the actual capture file if needed
</details>

I'm hoping you could

1. Go over existing implementation of `read_pdu` and `write_pdu` and changes to the interface of the server/client from a high level so I know if I'm approaching it right.
2. Similar for high level review of how I've introduced the async functionality into `storescu` and `storescp`  
3. Help me with the `AsyncPDataWriter` issue  hopefully

After that, I'd like to start writing tests, I just didn't want to write them yet if the interface is going to change.

And then I'd like to do some benchmarking of the async vs sync code and also the new sync code (with the framing) vs. the old sync code.

LMK your thoughts!  Thanks again!

Enet4 (commented):
Thank you for continuing with this PR. There is a lot to go through, and it is not clear from the code what the problem with PData reading/writing could be. I can try testing this against other platforms. Until then, I left some feedback inline for things which should be taken care of.

encoding/src/text.rs
```diff
@@ -65,7 +65,7 @@ type DecodeResult<T> = Result<T, DecodeTextError>;
 
 /// A holder of encoding and decoding mechanisms for text in DICOM content,
 /// which according to the standard, depends on the specific character set.
-pub trait TextCodec {
+pub trait TextCodec: Send + Sync {
```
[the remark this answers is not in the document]
naterichman: Looks like this was leftover from how I had previously implemented async, so yes I'll go ahead and remove that

ul/src/association/client.rs
```diff
@@ -773,11 +1052,81 @@ impl ClientAssociation {
...
+                Some(pdu) => {
+                    self.read_buffer.advance(buf.position() as usize);
+                    return Ok(pdu)
+                },
+                None => {
+                    // Reset position
+                    buf.set_position(0)
+                }
+            }
+            let recv = self.socket.read_buf(&mut self.read_buffer).await.unwrap();
```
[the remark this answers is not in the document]
naterichman: fixed, forgot to add the `context`s

naterichman:
Okay, I think I've mostly finished.  I have `storescp` and `storescu` working as sync and async.  I'm having some problems with feature flags though that I could use some help with, and thats also causing the CI to fail.

When I compile each package  within the workspace separately they work, but when I try to do anything on the whole workspace I get issues, i.e. store-scu tries to pull in the `async` feature set of `ul` when I try to do `cargo build` or `cargo test` on the whole workspace, even though `async` is not a default feature for any of the members that have that as a flag.

naterichman:
Okay, I think I fixed the workspace compilation issue, and looks like all the CI is passing. I'm going to go ahead and mark this as no longer a draft!

naterichman:
As I'm thinking more about this, I'm wondering if the sync/async methods shouldn't be named differently.  Potentially keeping the existing methods like `ClientAssociation::send` and then making the async ones like `ClientAssociation::send_async`.  LMK what you think

nathan-brainkey:
@naterichman , thanks for your work on an async api. I was testing out your work on my project and ran into some errors. My issue is that I have a storescp that depends on the sync api (previously written before you wrote the async functionality) and I am working on a storescu that depends on the async features. The two don't seem compatible since functions like establish on ServerAssociationOptions are conditionally compiled based on the "async" feature flag. I can work around this but wanted to bring up the case where the same package might need to use both the async and sync apis.

nathan-brainkey:
@naterichman and @Enet4 would it make sense to separate the async and sync functionality into separate modules so that the async and sync features can be used together? You could leave all the common functionality as is but anywhere where async and sync code use the same functions, move those functions to sub-modules? Then, you could put those modules behind feature gates and people can select whether they want the async functions, sync functions, or both?

naterichman:
@nathan-brainkey I hadn't thought of using sync and async in the same program but it makes sense.  I think switching to the sync/async modules makes sense.  I was trying to avoid as much code duplication as I could but I think without conditional complication it will have to happen. 

For example in the current version of the implementation `ClientAssociation` s `socket` is a `std::net::TcpStream` when its `sync` and `tokio::net::TcpStream` when its async, so I'll definitely need to create two types, one for sync and one for async.

I'll get around to this hopefully sometime in the next few weeks!

naterichman:
@nathan-brainkey Just pushed a version with those changes, showing in `storescu` I exposed a `--blocking` flag so the comand runs in async by default but also can run in blocking mode if the flag is passed.  LMK your thoughts!

naterichman:
Hi @Enet4, I've been using this branch in a project of mine for a few weeks now and its working pretty well.  I realize this PR itself is pretty messy with a few different stages of changes. What can I do to clean it up and/or get it ready for review? Would you like me to rebase and clean up the commit history, or I could just create a new branch/PR from the current state?  

I'm also planning on going through and adding/fixing up documentation.  Let me know if there is anything else you'd like me to do!

Enet4:
Thank you for your commitment @naterichman. I haven't had much free time to work on DICOM-rs last month, but I'm looking forward to getting more contributions rolling this month.

Is CI still failing on this branch? Maybe a minor tweak is missing in one of the tests.

naterichman:
Hi @Enet4, was strange, it was only in beta and it has now gone away...

Enet4:
> Hi @Enet4, was strange, it was only in beta and it has now gone away...

Oh yes, it was likely a bad rustc lint in beta (#564), which has been reverted by now.

Enet4 (commented):
So I went through the API changes and documentation, and tried the happy path of sending a DICOM file using `dicom-storescu` into a `dicom-storescp` server. Some key concerns follow.

### Documentation

It is not clear from reading the documentation of `dicom-ul` that a non-blocking API is available (inc. how it works and how to enable it). While there are plans at the Rust project level to provide better support for documenting Cargo features, we're keeping these points explicit in the DICOM-rs project by explaining them at crate-root level.

There were also a few unrelated changes to the API in terms of module visibility that I would like to see revised (see inline comment).

### dicom-storescu to dicom-storescp

Given a C-STORE SCP running like this:

```sh
cargo run --bin dicom-storescp
```

I tried this:

```sh
cargo run --bin dicom-storescu -- localhost:11111 --never-transcode .../dicom-test-files/data/WG04/JPLY/MG1_JPLY
```

And `dicom-storescp` would fail to read the PDU data:

```none
2024-10-06T12:31:05.711079Z ERROR dicom_storescp: failed to read DICOM data object

Caused by these errors (recent errors listed first):
  1: Could not read data set token
  2: Could not read item header
  3: Could not decode element header at position 2098620
  4: Failed to read the item header
  5: failed to fill whole buffer
```

It works if I change `dicom-storescu` back to the latest stable version, so this could be a mistake in the PDU writing process somewhere.

ul/src/association/mod.rs
```diff
@@ -17,14 +17,10 @@
 //! [1]: std::net::TcpStream
 pub mod client;
 pub mod server;
+
 mod uid;
 
-pub(crate) mod pdata;
+pub mod pdata;
```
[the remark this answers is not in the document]
naterichman: I think I just made the visibility higher when I was doing some testing.  Made it back to `pub(crate)`!

Enet4:
> Where are those files you were testing on? I'd like to see if I can reproduce using that file, or maybe its because you are on windows and there are some minor differences (I'm on linux).

The file I used was `MG1_JPLY` from the WG04 test file set, though I'm not sure if the problem was specific to this file. It can also be downloaded from [dicom-test-files](https://github.com/robyoung/dicom-test-files).

I was using Linux this time, but I can try again on both machines in any case and get back to you. :)

Enet4 (commented):
I tried the same test again and this time it worked! Aside from the comments inline, all that should be left to do here is fix the tests.

.github/workflows/rust.yml
```diff
@@ -33,6 +33,7 @@ jobs:
       - run: cargo test -p dicom-pixeldata --features gdcm
       # test dicom-pixeldata without default features
       - run: cargo test -p dicom-pixeldata --no-default-features
+      - run: cargo test -p dicom-ul --features async
```
Enet4, line 36: `async` was enabled by default, so we were not testing the code with `async` disabled. I fixed the situation by feature-gating. 

I also favored making the feature opt-in, as it facilitates decoupling the crate from the tokio ecosystem when async isn't needed, and does not prevent other crates in DICOM-rs from enabling it. This is a similar approach to `dicom-pixeldata` enabling features from `dicom-transfer-syntax-registry`.

Enet4 (commented):
I took the liberty of making some corrections around feature-gating on "async" (this was explained in one of the comments inline). I only have a suggestion for one of the doctests which I would like you to validate, then we're very likely ready to merge.

Enet4:
I would have greatly wished to merge this, but right now the transfer goes wrong whenever the `concurrency` option in `dicom-storescu` is set, even when I set it to 1. For instance:

```none
cargo run --bin dicom-storescu -- -c 1 ANYSCP@localhost:1111 .../dicom-test-files/data/WG04/JPLL
```

Our store SCP starts reporting errors of this kind at random:

```none
024-10-18T13:50:03.021555Z  INFO dicom_storescp::store_async: New association from STORE-SCU
2024-10-18T13:50:03.031017Z  INFO dicom_storescp::store_async: Stored storage\1.3.6.1.4.1.5962.1.1.20.1.4.20040826185059.5457.dcm
2024-10-18T13:50:03.060996Z ERROR dicom_storescp: failed to read DICOM data object

Caused by these errors (recent errors listed first):
  1: Could not read data set token
  2: Could not read item header
  3: Could not decode element header at position 1311814
  4: Failed to read the item header
  5: failed to fill whole buffer
```

And often the files won't end up saved 

Also tested with another store SCP (Dicoogle), it also reports errors and returns status code 101h.

This could constitute a problem with PDU writing in async mode, so withholding `concurrency` might just be hiding the problem. I'm afraid this will stay blocked until we find a fix.

naterichman:
I was worried that it was a random/intermittent thing when you said it failed at first but when you ran it again it passed.  I will try to reproduce and look into it!

storescu/src/store_async.rs
```diff
@@ -0,0 +1,251 @@
...
+                    is_last: true,
+                    data: cmd_data,
+                }],
+            };
+
+            scu.send(&pdu).await.map_err(Box::from).context(ScuSnafu)?;
+
+            {
+                let mut pdata = scu.send_pdata(pc_selected.id).await;
+                pdata.write_all(&object_data).await.unwrap();
```
Enet4, line 159: Should probably turn this into a recoverable error.

storescu/src/main.rs
```diff
@@ -313,190 +351,167 @@ fn run() -> Result<(), Error> {
...
+            let _ = scu.release().await;
+            Ok::<(), Error>(())
+        });
+    }
+    while let Some(result) = tasks.join_next().await {
+        if let Err(e) = result {
+            error!("{}", Report::from_error(e));
+            if fail_first {
+                std::process::exit(-2);
             }
```
Enet4, line 505: This code is testing the error on task join, but not the application error returned by the task. This worked better on my machine.
```suggestion
        match result {
            Err(e) => {
                error!("{}", Report::from_error(e));
                if fail_first {
                    std::process::exit(-2);
                }
            }
            Ok(Err(e)) => {
                error!("{}", Report::from_error(e));
                if fail_first {
                    std::process::exit(-2);
                }
            }
            Ok(Ok(_)) => {}
        }
```

Enet4 (commented):
I was looking into this right now! Still haven't gotten to the root of the problem, but encountered a few other things in the code worth looking into.

naterichman:
Will get to those too.  Don't worry about figuring out the sendscu issue, its something within the messed up logic of `self.writing` in the implementation of `AsyncPDataWriter.poll_write`.  I've had a hell of a time figuring out how to handle the underlying stream write call returining `Poll::Pending`, so I just need to figure out the correct logic for handling that, but it will get there!

naterichman:
Okay I believe this is fixed.  It only ever came up on sufficiently large files and I believe it had to do with how the tokio `TCPStream` write method works in that a lot of data can be written immediately, but with a large enough amount to transfer, it can sometimes return `Pending` instead of `Ready`. This explains why I had no issues sending a lot of smaller files and then was able to reproduce by trying to send a large US dicom.

I'll be honest feel much better about this implementation too since the original was mostly guided by chatGPT...  This implementation I understand much better, and I left a decent amount of inline comments explaining it for future reference

Enet4 (approved):
OK, I did a few more tests and found no more issues. Terrific work!

PR #558, opened 2024-08-31
Fix typos found by codespell

DimitriPapadopoulos:


dictionary-std/src/uids.rs
```diff
@@ -1420,10 +1420,10 @@ pub const IMPLANT_TEMPLATE_GROUP_INFORMATION_MODEL_MOVE: &str = "1.2.840.10008.5
 /// SOP Class: Implant Template Group Information Model - GET
 #[rustfmt::skip]
 pub const IMPLANT_TEMPLATE_GROUP_INFORMATION_MODEL_GET: &str = "1.2.840.10008.5.1.4.45.4";
-/// Application Hosting Modle: Native DICOM Model
+/// Application Hosting Model: Native DICOM Model
```
Enet4, line 1423: Looks like the typo leaked into the generated files. It's fine to fix them here, might also think about generating updated files soon.

Enet4 (commented):
Fair, some typos slip occasionally and they add up. It's curious that typos such as the one in partial.rs:310 wasn't picked up, were they explicitly excluded because they are already fixed in #557?

DimitriPapadopoulos:
The codespell tool is based on a dictionary of common typos, `retrievies` was not part of that dictionary.

I have added this typo in https://github.com/codespell-project/codespell/pull/3530.

Enet4 (approved):
Sounds good, thank you. :+1:

PR #562, opened 2024-09-03
Fixes discrepancy between multi and single frame decode_pixel_data

thomas-tribus:
When using decode_pixel_data a new photometric interpretation is computed based on the number of channels in the input image (see below). The single frame decode_pixel_data_frame doesn't have this (seemingly incorrect) transformation.


Enet4 (approved):
Looks OK. Thanks!

PR #569, opened 2024-10-08
- add connection timeout

jmlaka:
A TCP connection od a client connection will hang in case of an invalid address or network error.
This adds the option of a connection timeout, which is different from the already implemented read/write timeout on the TcpStream

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -22,7 +22,7 @@ use crate::{
     },
     AeAddr, IMPLEMENTATION_CLASS_UID, IMPLEMENTATION_VERSION_NAME,
 };
-use snafu::{ensure, Backtrace, ResultExt, Snafu};
+use snafu::{ensure, Backtrace, OptionExt, ResultExt, Snafu};
 
 use super::{
     pdata::{PDataReader, PDataWriter},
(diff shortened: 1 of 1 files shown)


ul/src/association/client.rs
```diff
@@ -546,11 +570,22 @@ impl<'a> ClientAssociationOptions<'a> {
...
         });
 
-        let mut socket = std::net::TcpStream::connect(ae_address)
-            .context(ConnectSnafu)?;
-        socket.set_read_timeout(read_timeout)
+        let mut socket: TcpStream = if let Some(timeout) = connection_timeout {
+            let address = ae_address
+                .to_socket_addrs()
+                .context(ToAddresssSnafu)?
+                .next()
```
Enet4, line 577: The default (no connect timeout) would try any of the resolved socket addresses until one is resolved. with this approach, any other socket addresses resolved are ignored in the scenario of one string resolving to multiple socket addresses. To prevent this from being too disruptive, we can just document this behavior in `ClientAssociationOptions::connection_timeout`.

ul/src/association/client.rs
```diff
@@ -34,20 +35,31 @@ pub enum Error {
...
+    /// could not convert to socket address
+    ToAddresss {
+        source: std::io::Error,
+        backtrace: Backtrace,
+    },
+
+    /// converted SocketAddrs iterator did not yield
+    #[snafu(display("not a single tcp addreess provided"))]
+    #[non_exhaustive]
+    NoAddress {},
```
Enet4, line 47:
```suggestion
    NoAddress { backtrace: Backtrace },
```

Enet4 (commented):
OK, looks nice and simple! Please see the minor suggestions below and we should be able to merge afterwards!

ul/src/association/client.rs
jmlaka: Or maybe would you like to mirror the behavior of TcpStream::connection and only fail if all parsed addresses inside the iterator fail to establish connection ?

jmlaka:
@Enet4  would this work ?
Thanks

jmlaka:
As I had to store all the connection errors up to an Ok result, I had to use a Vec.
When calling  `.context(ConnectSnafu)`, the Backtrace is generated only in the case of Result::Err  but not in the case of Result::Ok. Is that right?

jmlaka:
So this last one fails early by borrowing the `addresses` iterator and `count()` ing it.
Then I construct one os Error in the cheapest way possible (it will never be returned).
And avoid using a Vec ...

Enet4 (approved):
Awesome. Thanks! 👍

jmlaka:
@Enet4 please retract this pull request as the code does not work as intended I am sorry.
I will rewrite it and add a connection test.
Thank you

Enet4:
Hm, I'm away from a desktop computer, but if you write a pull request within the weekend I'll likely have a moment to try it out and merge.

PR #570, opened 2024-10-12
Client Association connection timeout working

jmlaka:
- multiple SocketAddrs accepted

I removed the empty SocketAddr iterator check, as it was too complicated and did not add much information.
The previous, faulty pull request counted the iterator and consumed it in the process, thus the error.

I did not write a test for the connection timeout functionality, as it is not possible to simulate
a connection timeout with the currently used test tooling inside the library.
( `TcpListener::accept()` cannot be used for this)

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -41,11 +41,6 @@ pub enum Error {
         backtrace: Backtrace,
     },
 
-    /// converted SocketAddrs iterator did not yield
-    #[snafu(display("not a single tcp addreess provided"))]
-    #[non_exhaustive]
-    NoAddress { backtrace: Backtrace },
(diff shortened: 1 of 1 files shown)


Enet4 (approved):
Right, looks good. Thank you for the follow-up!

> I did not write a test for the connection timeout functionality, as it is not possible to simulate
a connection timeout with the currently used test tooling inside the library.
> ( TcpListener::accept() cannot be used for this)

Testing the happy path would be possible in one of the integration tests in `tests`. Ensuring that the timeout is truly in place is something which I suspect would either demand a custom TCP implementation or some system-level scaffolding to control address name resolution for the test suite. But for now we'll go with this.

Issue #129, opened 2021-05-20
Account for Planar Configuration


Enet4:
This was previously overlooked, but the Planar Configuration (0028,0006) attribute should be checked whenever the image has more than one sample per pixel. At the moment, the decoding functions assume that it is in standard order (`0`), and may have funny results if it is in channel-first order (`1`).

Relevant part of the standard: [Part 3, section C.7.6.3](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.3.html#sect_C.7.6.3.1.3)

abustany:
related PR: https://github.com/Enet4/dicom-rs/pull/575

PR #575, opened 2024-10-21
[pixeldata] Add support for PixelFirst planar configuration (#129)

abustany:
Hi,

this PR adds a very simple interleave function to handle planar pixel configurations when decoding pixel data. I only have planar RGB8 files to test, but I don't see why 16-bit files would behave very differently...

The interleave implementation is very simple, but seems to yield decent performance (30ms to interleave a 12MP image on my 10 year old laptop). I experimented with a more "imperative" version with for loops which ended up being *slower* (probably because of bound checks), and a simple SSE implementation based on mm_shuffle_epi8 wasn't significantly faster either.

### DIFF
--- a/pixeldata/src/lib.rs
+++ b/pixeldata/src/lib.rs
@@ -692,19 +692,14 @@ impl DecodedPixelData<'_> {
             3 => {
                 // Modality LUT and VOI LUT
                 // are currently ignored in this case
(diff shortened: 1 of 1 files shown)


Enet4 (approved):
The code is sound and clean. I would not worry too much about optimizing it aggressively at this time. It is great to have support for the pixel first planar configuration. Much appreciated! :+1:

PR #576, opened 2024-10-21
[pixeldata] Handle empty NumberOfFrames tags

abustany:
I'm not sure of what the semantic difference between "absent tag" and "empty tag" is, but I do have some XRays with an empty "NumberOfFrames" tag.


Enet4 (approved):
Sure, sounds fair. Providing an empty _Number Of Frames_ does not seem compliant from my interpretation of the standard, but is simple to handle with hardly any downsides. Thank you. :+1:

PR #586, opened 2024-10-30
[parser] [object] Revert use of inspect method for more Rust compiler compatibility

### LINKED ISSUE #585: Usage of `inspect`  changed rust version of dicom-rs
```
error[E0658]: use of unstable library feature 'result_option_inspect'
   --> /home/rafal/.cargo/git/checkouts/dicom-rs-165e2a1669724caa/ec178ec/parser/src/stateful/decode.rs:916:14
    |
916 |             .inspect(|_| {
    |              ^^^^^^^
    |
    = note: see issue #91345 <https://github.com/rust-lang/rust/issues/91345> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `dicom-parser` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
```

I use quite old rust 1.72.1, because this is the latest version supported by my yocto version.

Is the use of inspect(and perhaps other functions as well) currently worth increasing the minimum version of rust?

Enet4:
This reverts some of the changes done in #580. `Result::inspect` and `Option::inspect` were only introduced in 1.76.0, but there is some interest in using the DICOM-rs library with older versions.

Resolves #585.

### DIFF
--- a/object/src/mem.rs
+++ b/object/src/mem.rs
@@ -940,10 +940,11 @@ where
     /// reporting whether it was present.
     pub fn remove_element_by_name(&mut self, name: &str) -> Result<bool, AccessByNameError> {
         let tag = self.lookup_name(name)?;
-        Ok(self.entries.remove(&tag).is_some()).inspect(|&removed| {
+        Ok(self.entries.remove(&tag).is_some()).map(|removed| {
             if removed {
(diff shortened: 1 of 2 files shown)


Enet4:
Dropping this in favor of #587.

PR #587, opened 2024-10-31
Set minimal rust version to 1.72.0

qarmin:
This PR sets minimal rust version to 1.72.

Verification that libraries with default features compile with Rust 1.72 is included in the CI.

This version of rust is not enforced for binary projects and custom features, because they depend on external libraries that often increase the minimal versions of rust(an example is clap - 4.0(<1.72) and already 4.5(1.74))

Some tests failed on Rust 1.72.0, so I increased error size

Also due adding clippy.toml, clippy now should only report problems that don't need rust >=1.73

~~PR depends and contains #586~~


Enet4 (commented):
Thank you for the initiative. This should take part in 0.8.0 with a minor revision.

.github/workflows/rust.yml
```diff
@@ -24,16 +25,38 @@ jobs:
...
+          cargo test -p dicom-core
+          cargo test -p dicom-encoding
+          cargo test -p dicom-dictionary-std
+          cargo test -p dicom-parser
+          cargo test -p dicom-transfer-syntax-registry
+          cargo test -p dicom-object
+          cargo test -p dicom-dump --no-default-features --features sop-class
+          cargo test -p dicom-json
+          cargo test -p dicom-ul
+          cargo test -p dicom-pixeldata
```
Enet4, line 57: We can also check the parent crate, which some folks use.

```suggestion
          cargo test -p dicom-pixeldata
          cargo check -p dicom
```

PR #588, opened 2024-11-01
Revert "[ul] Recover API compatibility with 0.7.1"

Enet4:
Reverts Enet4/dicom-rs#582, because the pull request does not fulfill a complete compatibility with users of the previous users. To be decided whether to move forward with this.


Enet4:
#592

PR #596, opened 2024-11-07
MAIN: Add test for connect timeout

naterichman:


Enet4 (approved):
Should be OK now. :+1: Thanks!

PR #598, opened 2024-11-11
[pixeldata] Allow overriding the VoiLutFunction

feliwir:
This is required to implement windowing for rendered resources (WADO-RS):
https://dicom.nema.org/medical/dicom/2024c/output/chtml/part18/sect_8.3.5.html#sect_8.3.5.1.4


Enet4 (approved):
Looks good to me. Thank you! :+1:

Issue #232, opened 2022-04-13
[pixeldata] VOI LUT Sequence support


Enet4:

abustany:
In case it saves someone duplicated work: I'm currently preparing a PR for this.

PR #599, opened 2024-11-11
[pixeldata] Apply VOI LUT when rendering images

abustany:
Fixes: #232

Tested on couple of images locally, this seems to do the job.

There are couple of items I'd like to address before considering this one ready:
- add couple of unit tests
- get your feedback @Enet4 on whether the data structures like `VoiLut` are in the right place...
- maybe fix the naming, somehow "VOI LUT" can refer to lot of similar but different things :grimacing: This seems to be an issue with the standard already, but any feedback on how to make names clearer is welcome
- do we want to make the API public already? Else we could gate the new structs behind `pub(crate)` for now

### DIFF
--- a/pixeldata/src/attribute.rs
+++ b/pixeldata/src/attribute.rs
@@ -30,6 +30,9 @@ pub enum AttributeName {
     VoiLutFunction,
     WindowCenter,
     WindowWidth,
+    LutDescriptor,
+    LutData,
+    LutExplanation,
 }
 
 impl std::fmt::Display for AttributeName {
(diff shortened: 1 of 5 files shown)


abustany:
woops didn't realize I had broken the gdcm feature, will fix

Enet4:
This does look substantial, thank you for working on this!

> 1. add couple of unit tests
That would be greatly appreciated. I wonder if some of the DICOM test files available feature VOI LUT sequences. I also happen to have a test enhanced CT scan which I think features a VOI LUT sequence, but I need to check.

> get your feedback @Enet4 on whether the data structures like VoiLut are in the right place...
maybe fix the naming, somehow "VOI LUT" can refer to lot of similar but different things 😬 This seems to be an issue with the standard already, but any feedback on how to make names clearer is welcome

From the quick glance I found nothing egregiously wrong, but I will make a better review when I am able.

> do we want to make the API public already? Else we could gate the new structs behind pub(crate) for now

In this case there is already some value in the library picking up VOI LUT sequences when the object uses that instead of window levels, so we can start with crate-level visibility and expose it later.

Enet4:
> I think this will conflict with #598 i just created earlier today

If it helps create some coordination, we can decide an order of contribution merges so that we know which one should be rebased.

abustany:
> I think this will conflict with #598 i just created earlier today

yup, very likely :-D but I'm happy for you to merge first, I checked your PR and the conflict resolving shouldn't be too hard to do.

pixeldata/src/lib.rs
```diff
@@ -642,6 +658,36 @@ impl DecodedPixelData<'_> {
         }
     }
 
+    pub fn voi_lut_sequence(&self) -> Result<Option<&[VoiLut]>> {
```
Enet4, line 661: Consider adding a simple doc-comment to this public method.

Enet4 (commented):
Thank you again for working on this. I will only leave a few comments inline. Overall this seems to be heading in the right direction. Please let me know when the pull request is ready for a more thorough test.

abustany:
Thanks for the comments! I'll try to take a look next week. I unfortunately haven't found the time yet to finalize the tests - but that's still on my todo :grimacing:

Enet4:
Hello again! Feel free to let me know if you need assistance reviving this pull request, or if you would like someone else to take over.

abustany:
Hello! Sorry about not picking that up - it’s the typical situation where we’ve been using this in production without issues for a while, so the urgency of properly wrapping up the PR went down… And free time has somehow been scarce. Anyway, I’m totally happy with someone else taking over this PR, this is public code really :) If that helps upstreaming the code faster, then everybody wins. Else I still have it on my backlog.

pixeldata/src/attribute.rs
```diff
@@ -618,6 +621,153 @@ pub fn photometric_interpretation<D: DataDictionary + Clone>(
...
+
+    Ok(VoiLut {
+        min_pixel_value,
+        bits_stored,
+        data: lut_data.to_vec(),
+        explanation,
+    })
+}
+
+pub fn voi_lut_sequence<D: DataDictionary + Clone>(
```
Enet4, line 737:
```suggestion
/// Get the VOI LUT Sequence from the DICOM object
///
/// This function will look for the _VOI LUT Sequence_ (0028,3010) attribute
/// at the root, then the per frame functional groups,
/// and then the shared functional group.
pub fn voi_lut_sequence<D: DataDictionary + Clone>(
```

Enet4 (commented):
I took the liberty of rebasing and applying my recommendations, but then I found something curious. Would appreciate your insight.

pixeldata/src/lib.rs
```diff
@@ -644,6 +660,37 @@ impl DecodedPixelData<'_> {
...
+                                nr_frames: self.number_of_frames(),
+                            }
+                            .fail()?
+                        }
+                        tracing::warn!(
+                            "Expected `{:?}` VOI LUTs, found `{:?}`, using first value for all",
+                            self.number_of_frames,
+                            len
+                        );
+                        Ok(Some(&inner[0..0]))
```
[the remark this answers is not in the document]
abustany: Woops yes indeed, that’s a typo

abustany:
> I took the liberty of rebasing and applying my recommendations

that's perfect, thanks a lot!

pixeldata/src/transform.rs
```diff
@@ -189,6 +191,53 @@ fn window_level_sigmoid(value: f64, window_width: f64, window_center: f64, y_max
...
+        }
+
+        if bits_stored < (lut.bits_stored as u16) {
+            panic!(
+                "LUT with BitsStored {} cannot be used for an image with BitsStored {}",
+                lut.bits_stored, bits_stored
+            );
+        }
+
+        let shift = (bits_stored.next_power_of_two() - (lut.bits_stored as u16)) as u32;
```
abustany, line 221: the if should not check `bits_stored` but `bits_stored.next_power_of_two()` here: it's allowed to have an image with eg. BitsStored=12 but a LUT descriptor that has BitsStored=16 .

abustany:
Thank you so much for driving this one to the end!

PR #600, opened 2024-11-12
Correctly return the first element of the array

dougyau:
The fallback code of rescale and voi lut functions return an empty slice instead of the expected 1 element slice.


Enet4 (approved):
Thanks! Good catch, that was clearly off. :+1:

PR #601, opened 2024-11-13
Use the voi function, window and rescale of the decoded frame

dougyau:
When decoding a single frame, frame_number is set to 1 but the voi_lut_function, the rescale and the window data is kept as an array. This change will use the data corresponding to the frame it will fallback to the first element if possible.

### DIFF
--- a/pixeldata/src/lib.rs
+++ b/pixeldata/src/lib.rs
@@ -2155,12 +2155,42 @@ where
             .fail()?;
         }
 
-        let rescale = zip(&rescale_intercept, &rescale_slope)
+        let rescale_data = zip(&rescale_intercept, &rescale_slope)
             .map(|(intercept, slope)| Rescale {
                 intercept: *intercept,
                 slope: *slope,
             })
-            .collect();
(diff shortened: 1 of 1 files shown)


pixeldata/src/lib.rs
```diff
@@ -2155,12 +2155,42 @@ where
...
                 slope: *slope,
             })
-            .collect();
+            .collect::<Vec<Rescale>>();
+
+        let rescale = {
+            if rescale_data.len() > frame as usize {
+                vec![rescale_data[frame as usize]]
+            } else if !rescale_data.is_empty() {
+                vec![rescale_data[0]]
```
Enet4, line 2169: Maybe even better:
```suggestion
            } else if let Some(resc) = rescale_data.get(0) {
                vec![*resc]
```
The same trick can be done in the other cases.

dougyau: Switched the code from the if blocks to a functional style.

Enet4 (approved):
Yep, this looks OK. Thank you. :+1:

Issue #604, opened 2024-11-25
Regression when gathering worklist from OrthanC


qarmin:
I bisected that since 

31408225ec129d0f7ef4bee49afe10a7714110ee
3251368e622a385de3a01a5581f5bb730187d902
b8d82002cbadf8603320bc2166c320b9e83a330c
1f79816dcb57781c0666e07bd3a3619f4c8497bd

(I'm not sure, because all these commits fails to build)

I'm having such error, when gathering first worklist item
```
2024-11-25T13:32:07.969320Z DEBUG dicom_findscu: Operation pending: ff00
2024-11-25T13:32:07.983659Z ERROR dicom_findscu: Failed to read response data

Caused by this error:
  1: Incoming pdu was too large: length 151126016, maximum is 131072
```

qarmin:
I suspect that error is here - pdu_length seems to have decoded invalid value: 

https://github.com/Enet4/dicom-rs/blob/e81a438d05540371fe24a2fd9dc4a442ed761fd2/ul/src/pdu/reader.rs#L34-L55

qarmin:
There is difference between received buf and pdu_length

this is shown as hex
buf 0.7.1
```
0200
0400
0600
0200
0400
0400
0400
0400
0400
0400
0400
0400
0400
0400
0400
0600
```
buf 0.8.0
```
0200
0400
0400
0600
0200
0400
0000
0400
```

pdu_length 0.7.1
```
ba
58
04
ba
58
b0
58
8c
58
8c
58
b0
58
90
58
04
```
pdu_length 0.8.0
```
ba
58
58
04
ba
58
9020000
b0
```

looks that 3 pdu is duplicated in 0.8.0 and different that in 0.7.1

qarmin:
This also happens with wlmscpfs
Steps to reproduce
```
sudo apt install dcmtk
mkdir -p ~/test/worklist/a

// TODO copy worklist files to ~/test/worklist/a
// "a" name is important, because it needs to be same as called-ae-title

sudo wlmscpfs --data-files-path ~/test/worklist 11111 -d
```

```
cargo run -- 127.0.0.1:11111 --calling-ae-title "a" --called-ae-title "a" --mwl -q PatientName
```
shows error
```
2024-11-25T18:49:49.192714Z ERROR dicom_findscu: Failed to read response data

Caused by this error:
  1: Incoming pdu was too large: length 33554432, maximum is 131072
```

version 0.7.1 
```
[...]

PR #606, opened 2024-11-26
[ul] BUG: Share underlying buffer from client/server to `PDataReader`

naterichman:
By creating a new read buffer in PDataReader, we potentially miss out on
any bytes that the `receive()` method read but didn't process yet.

Additionally, we lose the opposite when returning out of `PDataReader`:
remaining data in the PDataReader buffer that it didn't process.

The solution here is to mutably share that underlying read buffer with the
`PDataReader` so that it can continue from where the `receive()` method
left off _and_ pass any data that it didn't read back to the read buffer
of the main `Client` or `Server` object.

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -142,7 +142,12 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 /// Chunks of data are read into `read_buffer`,
 /// which should be passed in subsequent calls
 /// to receive more PDUs from the same stream.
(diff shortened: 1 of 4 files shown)


naterichman:
Fixes #604

Enet4 (approved):
Thank you for looking into this. I had identified a few places where the lack of buffer sharing could cause issues, but this one was probably overlooked, even though it is important.

Somewhat related, but not caused by this PR, I found that the current fuzz target test in ul/fuzz fails, so there may still be some bugs in the new PDU reading/writing impls. Hopefully there will be more opportunities to track them down. But for now, it looks good to be merged. Thanks! :+1:

PR #607, opened 2024-11-26
Broken pixel_data fix: support parsing improperly closed encapsulated pixel data

jmlaka:
Hello @Enet4 .
So I stumbled upon some Dicom files produced on a new Siemens machine and was unable to open them with Dcmtk, Pydicom nor dicom-rs.

I found this (fix](https://github.com/fo-dicom/fo-dicom/pull/1718)) that addressed the very same issue inside fo-dicom.

By reading your code and applying this little hack, I was able to get the files working.

Would you be willing to incorporate such a fix and document it please ?

This is a sample Dicom from the machine:
[Anonymised.zip](https://github.com/user-attachments/files/17925385/Anonymised.zip)

Thank you

### DIFF
--- a/parser/src/dataset/read.rs
+++ b/parser/src/dataset/read.rs
@@ -374,6 +374,13 @@ where
                         }
                     }
                 }
+                Err(DecoderError::DecodeItemHeader {
+                    source: dicom_encoding::decode::Error::ReadItemHeader { source, .. },
+                    ..
+                }) if source.kind() == std::io::ErrorKind::UnexpectedEof => {
+                    self.hard_break = true;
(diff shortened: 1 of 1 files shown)


Enet4 (approved):
From reading the other issues at fo-dicom and inspecting that DICOM file I interpret that the device is producing a non-standard way of delimiting the end of the pixel data. The proposed solution of gracefully ending the token stream when reaching EOF while in a pixel data fragment sequence is simple and does not seem to negatively affect anything else, so it is OK to incorporate this fix.

Much appreciated. :+1:

PR #609, opened 2024-12-07
Cargo update (2024-12)

Enet4:


qarmin:
There is even more things to update
```
cargo +nightly -Z unstable-options update --breaking
cargo update
```
found also
```
   Upgrading ndarray ^0.15.1 -> ^0.16.1
   Upgrading jxl-oxide ^0.9.1 -> ^0.10.2
```


Rustrover shows several unused crates
```
devtools/dictionary-builder/Cargo.toml -> serder_json
parser/Cargo.toml -> chrono
scpproxy/Cargo.toml -> dicom-dictorionary-std
ul/Cargo.toml -> dicom-core
```

Also 2 files are not used
```
transfer-syntax-registry/src/deflate.rs
```
and
```
object/src/attribute.rs
```

Enet4:
OK, I will see what can be further updated.

> ```
>    Upgrading ndarray ^0.15.1 -> ^0.16.1
> ```

This is a breaking change because `ndarray` is part of `dicom-pixel-data`. We can keep the update ready, but it likely won't enter before 0.9.0.

> ```
>    Upgrading jxl-oxide ^0.9.1 -> ^0.10.2
> ```

This is something that we probably can update right away, but there are updates to the API which I believe `dicom-transfer-syntax-registry` can take advantage of, so it is better off handled in a separate PR.

> ```
> devtools/dictionary-builder/Cargo.toml -> serder_json
> ```

Granted, this was probably a partial realization of the intent to support producing a JSON object of tags instead of a Rust code file.

> parser/Cargo.toml -> chrono
> scpproxy/Cargo.toml -> dicom-dictorionary-std
> ul/Cargo.toml -> dicom-core
> ```

That can be done for the sake of simplifying the Cargo manifests, though the dependencies themselves are fetched transitively anyway.

> 
> Also 2 files are not used
> 
> ```
> transfer-syntax-registry/src/deflate.rs
> ```

I suggest that we revive #438 instead.

> 
> and
> 
> ```
> object/src/attribute.rs
> ```

Will take care of that in #524.

PR #613, opened 2024-12-11
Fix typos

jayvdb:


object/src/file.rs
```diff
@@ -103,7 +103,7 @@ impl<D, T> OpenFileOptions<D, T> {
     }
 
     /// Set the transfer syntax index to use when reading the file.
-    pub fn tranfer_syntax_index<Tr>(self, ts_index: Tr) -> OpenFileOptions<D, Tr>
+    pub fn transfer_syntax_index<Tr>(self, ts_index: Tr) -> OpenFileOptions<D, Tr>
```
Enet4, line 106: Changing the name of this public function would be a breaking change now. We can 1) postpone this change until the time we aim for 0.9, or 2) re-add `tranfer_syntax_index` as a deprecated function which calls `transfer_syntax_index` underneath.

jayvdb: Will re-add as deprecated.  I should have thought of that!

object/src/file.rs
```diff
@@ -116,6 +116,15 @@ impl<D, T> OpenFileOptions<D, T> {
         }
     }
 
+    /// Set the transfer syntax index to use when reading the file.
+    #[deprecated(since="0.8.0", note="please use `transfer_syntax_index` instead")]
```
jayvdb, line 120: should this be 0.9.0 ?

Enet4: We're very likely doing v0.8.1 next.

Enet4 (approved):
:+1: Thanks.

Issue #40, opened 2020-05-18
Tracker issue for text encoding


Enet4:
Keeping a table and checklist of requirements regarding text encoding in DICOM:

| encoding | implemented | validated | Issue / PR |
|--------------|-------------------|-------------|---------------|
| ISO IR-100 | :heavy_check_mark: | :x: needs more testing |  |
| ISO IR-192 (UTF-8) | :heavy_check_mark: | :x: could still use some tests for greater assurance | |

In addition:

[ ] Validate edge cases in the the default character encoding 
[ ] Support for multiple character repertoires (extended character sets)
[ ] Consider migrating the text encoding dependency to `encoding_rs`

ikneg:
下記仕様のH章にある `ISO 2022 IR 13` `ISO 2022 IR 87` を対応表に加えることはできますか？
https://dicom.nema.org/medical/dicom/current/output/pdf/part05.pdf

PR #614, opened 2024-12-13
Add character sets: Arabic, Greek, Hebrew, Japanese, Thai, Korean

rforsyth:
This change adds support for the following character sets:
- ISO-IR 13 (Japanese)
- ISO-IR 87 (Japanese multi-byte)
- ISO-IR 126 (Greek)
- ISO_IR 127 (Arabic)
- ISO_IR 138 (Hebrew)
- ISO-IR 149 (Korean)
- ISO-IR 166 (Thai)

This is related to issue: #40, and is a superset of the currently open PR #444.

The full set of DICOM supported character sets is listed in: [Table D.6.2-1. Supported Specific Character Set Defined Terms](https://dicom.nema.org/medical/dicom/2017a/output/chtml/part02/sect_D.6.2.html).  This change covers the character sets in this list which are also supported by [rust-encoding](https://github.com/lifthrasiir/rust-encoding).

It appears that `rust-encoding` does not include support for ISO-IR 148 (Latin 5) or ISO-IR 159 (Japanese supplementary characters), so these have not been included.

### DIFF
--- a/encoding/src/text.rs
+++ b/encoding/src/text.rs
@@ -6,23 +6,30 @@
 //! | Character Set                 | decoding support | encoding support |
 //! |-------------------------------|------------------|------------------|
 //! | ISO-IR 6 (default)            | ✓ | ✓ |
+//! | ISO-IR 13 (WINDOWS_31J): The JIS X 0201-1976 character set (Japanese single-byte) | ✓ | ✓ |
+//! | ISO-IR 87 (ISO_2022_JP): The JIS X 0208-1990 character set (Japanese multi-byte) | ✓ | ✓ |
 //! | ISO-IR 100 (ISO-8859-1): Right-hand part of the Latin alphabet no. 1, the Western Europe character set | ✓ | ✓ |
 //! | ISO-IR 101 (ISO-8859-2): Right-hand part of the Latin alphabet no. 2, the Central/Eastern Europe character set | ✓ | ✓ |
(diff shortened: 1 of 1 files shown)


Enet4:
Thank you for your pull request! Do you happen to know whether any of these encodings also have an implementation at [`encoding_rs`](https://crates.io/crates/encoding_rs)? We fear that we may end up stuck with rust-encoding, which is no longer maintained (#200).

> The full set of DICOM supported character sets is listed in: [Table D.6.2-1. Supported Specific Character Set Defined Terms](https://dicom.nema.org/medical/dicom/2017a/output/chtml/part02/sect_D.6.2.html).

A curious reference. The normative part of the standard would be in [PS3.5 section 6.1](https://dicom.nema.org/medical/dicom/2024d/output/chtml/part05/chapter_6.html#sect_6.1), though I agree that the latest revisions do not expose the available character sets as nicely as in that table.

rforsyth:
Based on a review of the [encoding_rs_compat](https://github.com/hsivonen/encoding_rs_compat) library, all of the new character sets are compatible with [encoding_rs](https://crates.io/crates/encoding_rs) (see: [compat.rs](https://github.com/hsivonen/encoding_rs_compat/blob/master/src/compat.rs).  So I don't think that this change will be a blocker to switching.

Yeah, I just found the `SpecificCharacterSet` "defined terms" table easier to work from.

Enet4 (approved):
OK, this seems to be in order. It's also good to hear that compatibility layers towards `encoding_rs` are a possibility. Thanks!

rforsyth:
Thanks Eduardo!  Appreciate your quick responses.

PR #615, opened 2024-12-14
Add support for non-finite float handling in DICOM JSON

chrisbll971:
We encountered a few unusual DICOM files that store NaN, inf, and -inf values in float tags. These values appear to be valid according to the [DICOM Standard](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html).

This pull request adds support for serializing non-finite float values to DICOM JSON. Previously, these values were serialized to `null` causing the values to dissapear.

**Changes:**

- Added proper handling for non-finite float values (NaN, inf, -inf) during DICOM JSON serialization.

- Added a test to verify deserialization logic correctly parses these values from their DICOM JSON representations.

**Verification:**

- Tested with several DICOM files containing NaN, inf, and -inf values.

- Ensured that all non-finite float values remain intact after serializing and deserializing them to and from DICOM JSON and DICOM.

### DIFF
--- a/json/src/de/mod.rs
+++ b/json/src/de/mod.rs
@@ -535,4 +535,41 @@ mod tests {
 
         assert!(super::from_value::<InMemDicomObject>(serialized).is_ok());
     }
+
+    #[test]
+    fn can_resolve_nan_and_inf_float() {
+        let serialized = serde_json::json!({
+            "00020011": {
+                "vr": "FL",
+                "Value": [
+                    5492.8545,
+                    5462.5205,
+                    "NaN",
+                    "-inf",
+                    "inf"
+                ]
+            }
+        });
+
+        let obj: InMemDicomObject = super::from_value(serialized).unwrap();
+        let tag = Tag(0x0002, 0x0011);
+        let element = obj.get(tag).unwrap();
+
+        // verify NAN, INFINITY, and NEG_INFINITY are correctly deserialized to f32::NAN, f32::INFINITY, and f32::NEG_INFINITY
+        let actual_values = element.float32_slice().unwrap();
+        let expected_values = &[
+            5492.8545,
+            5462.5205,
+            f32::NAN,
+            f32::NEG_INFINITY,
+            f32::INFINITY,
+        ];
+
+        // need special comparison for NAN values since assert_eq will not match
+        assert_eq!(actual_values.len(), expected_values.len());
+        assert!(actual_values
+            .iter()
+            .zip(expected_values.iter())
+            .all(|(&a, &b)| (a == b) || (a.is_nan() && b.is_nan())));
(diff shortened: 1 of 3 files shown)


json/src/de/mod.rs
```diff
@@ -535,4 +535,41 @@ mod tests {
 
         assert!(super::from_value::<InMemDicomObject>(serialized).is_ok());
     }
+
+    #[test]
+    fn can_resolve_nan_and_inf_float() {
+        let serialized = serde_json::json!({
+            "00020011": {
```
Enet4, line 542: A slightly more realistic example would probably use a tag for FL values, such as Table of Parameter Values (0018,605A).

json/src/ser/value.rs
```diff
@@ -113,6 +121,12 @@ impl Serialize for AsNumbers<'_> {
                 for number in numbers {
                     if number.is_finite() {
                         ser.serialize_element(&number)?;
+                    } else if number.is_nan() {
+                        ser.serialize_element(&f64::NAN.to_string())?;
```
Enet4, line 125: We can skip these allocations by resolving the string inline. As long as we are careful about inserting correct string literals, it should be fine.
```suggestion
                        ser.serialize_element("NaN")?;
```

json/src/de/mod.rs
```diff
@@ -535,4 +535,41 @@ mod tests {
...
+                ]
+            }
+        });
+
+        let obj: InMemDicomObject = super::from_value(serialized).unwrap();
+        let tag = Tag(0x0002, 0x0011);
+        let element = obj.get(tag).unwrap();
+
+        // verify NAN, INFINITY, and NEG_INFINITY are correctly deserialized to f32::NAN, f32::INFINITY, and f32::NEG_INFINITY
+        let actual_values = element.float32_slice().unwrap();
```
Enet4, line 559: I wonder what would happen if the consumer were to request 64-bit floats ([`.to_multi_float64`](https://docs.rs/dicom-core/latest/dicom_core/value/enum.Value.html#method.to_multi_float64)). It should hopefully retain these values, but we can probably append some checks to this test.

chrisbll971: Sounds good, added a test for this

json/src/ser/value.rs
chrisbll971: Good callout - updated to use constants to avoid the to_string()

json/src/de/mod.rs
chrisbll971: Updated

chrisbll971:
Released a new version to incorporate your feedback

Enet4 (approved):
Looks OK! Thanks!

PR #620, opened 2024-12-30
Fix the characteristic that the trim() and trim_end() functions do not remove the \0 character.

### LINKED ISSUE #618: It crashed when I using obj.element(tags::WINDOW_WIDTH)?.to_multi_float64()?
```
use dicom::{
    dictionary_std::tags,
    object::OpenFileOptions
};

pub fn test() -> Result<(), Box<dyn std::error::Error>> {
    let _p3 = r"C:\Users\paul\Downloads\DICOM\DICOM_DARASET\Dcm_Data_JPEG2000\2.dcm";

    let obj = OpenFileOptions::new().open_file(_p3)?;

    // This runs well and prints: Window Width: ["00350", "00350"]
    let window_width = obj.element(tags::WINDOW_WIDTH)?.to_multi_str()?;
    println!("Window Width: {:?}", window_width);


    // This code crashed and prints:  called `Result::unwrap()` on an `Err` value: ConvertValueError { requested: "float64", original: Strs, cause: Some(ParseFloat { backtrace: Backtrace [{ fn: "std::backtrace_rs::backtrace::dbghelp64::trace", .....
    let window_width = obj.element(tags::WINDOW_WIDTH)?.to_multi_float64()?;
    println!("Window Width: {:?}", window_width);

    Ok(())
}
```

Please tell me if you need more information or context.  [test_dicom.zip](https://github.com/user-attachments/files/18249067/test_dicom.zip)

xb284524239:
fix #618 

Hello @Enet4 ,

I have standardized the codebase by uniformly substituting the `trim()` and `trim_end()` functions with the `trim_end_matches([' ', '\u{0}'])` function, which was replicated from the `to_str()` function within the `primitive.rs` file.
I am rather uncertain as to the rationale behind the concurrent existence of the `trim()` and `trim_end()` functions in the previous implementation. Was this a deliberate design choice?
Moreover, I am pondering whether it might be more apt to utilize the `trim_matches(|c: char| c.is_whitespace() | c.is_control())` function instead. However, I remain indecisive on this matter.

### DIFF
--- a/core/src/value/primitive.rs
+++ b/core/src/value/primitive.rs
@@ -909,7 +909,7 @@ impl PrimitiveValue {
     {
         match self {
             PrimitiveValue::Str(s) => {
-                s.trim()
+                s.trim_end_matches([' ', '\u{0}'])
                     .parse()
                     .context(ParseIntegerSnafu)
                     .map_err(|err| ConvertValueError {
@@ -919,7 +919,7 @@ impl PrimitiveValue {
                     })
             }
             PrimitiveValue::Strs(s) if !s.is_empty() => s[0]
-                .trim()
+                .trim_end_matches([' ', '\u{0}'])
                 .parse()
                 .context(ParseIntegerSnafu)
                 .map_err(|err| ConvertValueError {
@@ -1078,7 +1078,7 @@ impl PrimitiveValue {
         match self {
             PrimitiveValue::Empty => Ok(Vec::new()),
             PrimitiveValue::Str(s) => {
-                let out = s.trim().parse().context(ParseIntegerSnafu).map_err(|err| {
+                let out = s.trim_end_matches([' ', '\u{0}']).parse().context(ParseIntegerSnafu).map_err(|err| {
                     ConvertValueError {
                         requested: "integer",
                         original: self.value_type(),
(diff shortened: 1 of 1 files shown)


Enet4 (changes requested):
> I have standardized the codebase by uniformly substituting the `trim()` and `trim_end()` functions with the `trim_end_matches([' ', '\u{0}'])` function, which was replicated from the `to_str()` function within the `primitive.rs` file.
> 
> I am rather uncertain as to the rationale behind the concurrent existence of the `trim()` and `trim_end()` and `trim_end_matches([' ', '\u{0}'])` functions in the previous implementation. Was this a deliberate design choice?

I believe that the occurrences of both `trim*` and `trim_end*` are indeed deliberate. This is because some values admit both leading and trailing whitespace. https://dicom.nema.org/medical/dicom/2024d/output/chtml/part05/sect_6.2.html In the cases where `trim()` was called, please retain the begin trimming.

> Moreover, I am pondering whether it might be more apt to utilize the `trim_matches(|c: char| c == '\0' || c.is_whitespace())` function instead. However, I remain indecisive on this matter.

I would stick to checking specifically for spaces, as that is what's listed by the standard. `is_whitespace` may cover more cases of whitespace than necessary and/or intended.

xb284524239:
Hello @Enet4 ,
The changes to the calls of `trim()` and `trim_end()` have been restored, and only the matching for the `\0` character has been added.

Enet4 (approved):
Should be OK. Anyone depending on the presence of nul characters in the output should be using the raw string method variants intead.

Thank you for your contribution @xb284524239. :+1:

PR #622, opened 2024-12-30
Return in ClientAssociation also User Variables

### LINKED ISSUE #621: Add option, to read implementation class uid/name

qarmin:
Fixes #621

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -756,6 +756,7 @@ impl<'a> ClientAssociationOptions<'a> {
                     read_buffer: BytesMut::with_capacity(MAXIMUM_PDU_SIZE as usize),
                     read_timeout,
                     write_timeout,
+                    user_variables,
                 })
             }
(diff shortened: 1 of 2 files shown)


ul/src/association/client.rs
```diff
@@ -946,6 +949,12 @@ where
     pub fn requestor_max_pdu_length(&self) -> u32 {
         self.requestor_max_pdu_length
     }
+
+    /// Retrieve the user variables that were taken from the server.
+    /// It usually contains the maximum PDU length and the implementation class UID/name
```
Enet4, line 954:
```suggestion
    /// Retrieve the user variables that were taken from the server.
    ///
    /// It usually contains the maximum PDU length,
    /// the implementation class UID, and the implementation version name.
```

Enet4 (commented):
Thank you for the pull request! Would you be able to extend one of the association tests with this new method?

Minor suggestion inline.

Enet4 (approved):
Seems just right. Thank you @qarmin!

PR #624, opened 2025-01-06
fix(scp): use max_pdu_length method to set the value

### LINKED ISSUE #513: max_pdu_length isn't used in Store-SCP
The parameter `max_pdu_length` is just used to initialize a vector `buffer` that itself isn't used for anything.
https://github.com/Enet4/dicom-rs/blob/master/storescp/src/main.rs#L63C5-L63C76

MikeRomaniuk:
CLoses #513

### DIFF
--- a/storescp/src/store_async.rs
+++ b/storescp/src/store_async.rs
@@ -24,7 +24,6 @@ pub async fn run_store_async(
     } = args;
     let verbose = *verbose;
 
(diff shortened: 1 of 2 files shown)


Enet4 (approved):
Looks just about right. I noticed that the tool won't validate `max_pdu_length` (should be in `4096..=131072`) until it establishes associations, but this is a condition which can be fixed separately.

Thank you for your pull request. :+1:

Issue #626, opened 2025-01-08
Panic when reading file, manually assembling it and saving to vector


qarmin:
Dicom-rs 26cb283e795dd5bba69ddf09c71d9ebcae877939

Sadly I cannot provide exact steps to reproduce problem, because this depends on our proprietary library.

When  running code similar to (logic is very similar to code from below, but we additionally decompose entire file and compose it again)
```
pub fn test_items() {
    let file_path = "/opt/BROKEN_FILES_DIR/image-00000_IDX_0_RAND_1471178784642623661843127_minimized_519.dcm";
    let Ok(file) = dicom_object::OpenFileOptions::new().open_file(file_path) else {
        println!("Error opening file");
        return;
    };
    let transfer_syntax = file.transfer_syntax_uid();

    let mut object_data = Vec::with_capacity(2048);
    let Some(ts_selected) = dicom_transfer_syntax_registry::TransferSyntaxRegistry
        .get(transfer_syntax)
         else {
             println!("Unsupported transfer syntax");
             return;
         };
    let _ = file.write_dataset_with_ts(&mut object_data, ts_selected); // Here, instead file we collect elements in hierarchy 

    let mut new_dicom_obj = InMemDicomObject::new_empty();
    for elem in file {
        new_dicom_obj.put_element(elem);
    }
    let _ = new_dicom_obj.write_dataset_with_ts(&mut object_data, ts_selected);


    println!("DONE");
}
```

I have crash
```
thread 'main' panicked at 'internal error: entered unreachable code', /home/rafal/.cargo/registry/src/index.crates.io-6f17d22bba15001f/dicom-parser-0.8.0/src/dataset/mod.rs:538:82
stack backtrace:
   0: rust_begin_unwind
             at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/core/src/panicking.rs:67:14
   2: core::panicking::panic
             at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/core/src/panicking.rs:117:5
   3: <dicom_parser::dataset::DataElementTokens<I,P> as core::iter::traits::iterator::Iterator>::next
[...]

PR #628, opened 2025-01-11
[parser] Add robustness to data element token writing

Enet4:
This tweaks the data element token writing routine so that instead of panicking from unexpected values in headers with the VR SQ, it warns about the case and gracefully tries to provide some tokens.

- A primitive value inside a data element with the VR SQ is not expected. I've added an edge case where the primitive data can be interpreted as raw DICOM bytes comprising the sequence items. This also means that it might produce malformed data inside those sequences, but this is a case of bad data in and bad data out.
- Pixel fragment sequences should only exist inside a pixel data header. I have chosen to skip the element entirely when this happens.

### DIFF
--- a/parser/src/dataset/mod.rs
+++ b/parser/src/dataset/mod.rs
@@ -532,10 +532,36 @@ where
 
                 let token = DataToken::from(header);
                 match token {
-                    DataToken::SequenceStart { .. } => {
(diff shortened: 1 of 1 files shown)


qarmin:
I checked broken files from #626, and this PR fixes all problems that I had

PR #635, opened 2025-02-19
[core] Make `PersonName` hold both borrowed and owned strings

AlexTMjugador:
At the moment, `PersonName` cannot be ergonomically used by application code as an structure to group and pass around owned person name components that are meant to conform to DICOM's PN value semantics. It is, for example, difficult to come up with code for deserializing a `PersonName` if strings cannot be borrowed from the data source, which is a common ocurrence when the source is e.g. a stream that reads data chunks to temporary buffers.

To improve on that, let's make `PersonName` have `Cow<'a, str>` fields instead, and update `PersonNameBuilder` to accept them. `Cow`s are smart pointers that can hold both borrowed and owned data, so while `&'a str` can be wrapped in a `Cow<'a, str>` at virtually no cost, an owned `String` may also be wrapped with the same type but an indefinitely long lifetime, `Cow<'static, str>`. Therefore, code that borrows name components can still use `PersonName`s as before, and new code that relies on `PersonName`s owning their components can now easily work by just using a `'static` lifetime bound.

These changes are technically semver breaking due to the removal of the `Copy` trait on `PersonName`, and minor changes in the semantics of the `PersonNameBuilder::build` method.

### DIFF
--- a/core/src/value/person_name.rs
+++ b/core/src/value/person_name.rs
@@ -1,12 +1,14 @@
 //! Handling of DICOM values with the PN (person name) value representation
 //! as per PS3.5 sect 6.2.
+use std::borrow::Cow;
 use std::fmt::{Display, Formatter};
+use std::mem;
 
 /// A DICOM _Person Name_ (PN value representation).
 ///
 /// Values of this type keep
 /// family name, given name, middle name, prefix and suffix
-/// as borrowed values.
+/// as possibly borrowed values.
 /// All name components are optional.
 ///
 /// # Example
@@ -22,13 +24,13 @@ use std::fmt::{Display, Formatter};
 /// assert_eq!(dr_seuss.prefix(), Some("Dr."));
 /// assert_eq!(dr_seuss.given(), Some("Theodor"));
 /// ```
-#[derive(Debug, Copy, Clone, Eq, Hash, PartialEq)]
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
 pub struct PersonName<'a> {
-    prefix: Option<&'a str>,
-    family: Option<&'a str>,
-    middle: Option<&'a str>,
-    given: Option<&'a str>,
-    suffix: Option<&'a str>,
+    prefix: Option<Cow<'a, str>>,
+    family: Option<Cow<'a, str>>,
+    middle: Option<Cow<'a, str>>,
+    given: Option<Cow<'a, str>>,
+    suffix: Option<Cow<'a, str>>,
 }
 
 /// A builder to construct a [`PersonName`] from its components.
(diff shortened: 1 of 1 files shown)


Enet4:
Hello @AlexTMjugador! That is an interesting idea, but I wonder whether we can introduce this without a semver breaking change. Extending the owned version of a person name into a dedicated type (`OwnedPersonName`) would be feasible and have the benefit of being as efficient as it can be in the case where no additional ownership is necessary. Admittedly, it implies replicating more code, such as having two person builders.

- New structs `OwnedPersonName` and `OwnedPersonBuilder`, would use `String` for every field;
- One could trivially obtain a `PersonName<'a>` from an `OwnedPersonName` with a method call;
- Parsing a person name would still result in a borrowed `PersonName`, but would have a `ToOwned` implementation into a `OwnedPersonName`.

Let me know your thoughts on this.

AlexTMjugador:
Hello!

I believe we could introduce the variation you proposed if our primary goal here is to avoid any semver-breaking changes. However, the current approach in this PR has several advantages that, in my view, may justify keeping it as is:

- As you stated, having a single `PersonName` struct that supports both use cases aligns better with the DRY principle, improving maintainability and making it easier to achieve higher test coverage.
- While `PersonName` is a useful struct for specific situations, it [does not appear to be widely used in open-source code I could find](https://github.com/search?q=%22dicom_core%3A%3Avalue%3A%3Aperson_name%3A%3APersonName%22&type=code). Moreover, the breaking changes introduced by this PR are relatively minor and easy to migrate to if needed, so developer churn should remain negligible.
- Though admittedly a niche use case, a single struct with `Cow`s is the only way to support person names where some component strings are owned and others are borrowed. For example, a separate `OwnedPersonName` struct would not allow the following clone-free code, which this PR enables:
```rust
let mut name_builder = PersonName::builder();
name_builder.with_given("John"); // Borrowed from constant binary data
name_builder.with_family(String::from("Doe")); // Owned, may come from a buffer
let dicom_pn = name_builder.build().to_dicom_string(); // Doe^John
```

Ultimately, I'd understand if you decide this approach is not worth the technicality of being backwards incompatible, so please let me know how you'd like to proceed. For what it’s worth, in my experience as a library user, the `PersonName` implementation in this PR feels quite ergonomic, as it's easy and efficient to use whatever string variant comes from other parts of an application, and the additional generics usually don't need to be written or even deeply understood to be effective.

Enet4:
Fair enough, this can added to the 0.9 milestone.

AlexTMjugador:
Thanks for the reminder, @Enet4! I just wanted to quietly rebase this PR onto the latest master commits to keep it mergeable and not too distant from the mainline :)

PR #644, opened 2025-04-10
Update `charls` dependency to version 0.4.1

dougyau:
cmake v3.27 introduced CMAKE_LINK_DEPENDS_USE_LINKER define (https://cmake.org/cmake/help/latest/variable/CMAKE_LINK_DEPENDS_USE_LINKER.html) which breaks the build of charls, specially when doing cross compilation. 

To fix it CMAKE_LINK_DEPENDS_USE_LINKER=0 is required.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -282,18 +282,18 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "charls"
-version = "0.3.1"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9068bd9f6619c343279045d6e92f814ffd537c8038bb7fa9d8ab5525cbc0ae80"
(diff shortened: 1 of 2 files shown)


dougyau:
Updated the dependency to 0.4.2

It includes the change to streamline the package. Additional change on that library is the `vcpkg` feature flag which is useful on windows.

Should I add a `charls_vcpkg` feature flag on this project? What do you think?

Enet4:
> It includes the change to streamline the package. Additional change on that library is the `vcpkg` feature flag which is useful on windows.
> 
> Should I add a `charls_vcpkg` feature flag on this project? What do you think?

Oops, I must've overlooked this question. I just gave that a try (add feature `charls_vcpkg` to `dicom-transfer-syntax-registry` and `dicom-pixeldata`) on a Windows machine. It seems to work rather well after installing `charls` for `x64-windows-static` triplet. Great work!

The only downside is that each new feature in `dicom-transfer-syntax-registry` is replicated to `dicom-pixeldata`, but this is acceptable for the time being.

Issue #636, opened 2025-02-20
Explicit length Sequences and Items break dataset read / write


jmlaka:
So I stumbled upon a Dicom file containing and Explicit length Item.

Sending it via storescu failed. 
Reading the file and even writing the FileDicomObject back to file worked fine.
But it produced and invalid output (Missing ItemSequenceDelimiter)

Which led me here:
https://github.com/Enet4/dicom-rs/blob/b2002bb85f9fcc60e82f62ee6ee0b9adf23bd70e/parser/src/dataset/write.rs#L192

https://github.com/Enet4/dicom-rs/blob/b2002bb85f9fcc60e82f62ee6ee0b9adf23bd70e/parser/src/dataset/write.rs#L169

As the comment inside the code states:

> explicit length sequences or items should not print
> the respective delimiter

I'd like to ask:
What is the logic behind not writing a DataToken::ItemEnd or SequenceEnd ?

Is this just some weird file ? Because it comes from the same Siemens model that was messing up the pixelsequence ending.

[anonymised.zip](https://github.com/user-attachments/files/18893893/anonymised.zip)

Enet4:
> I'd like to ask: What is the logic behind not writing a DataToken::ItemEnd or SequenceEnd ?

I just followed [Part 5 section 7.5](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_7.5.html). The encoder can choose either one of the two ways to encode a sequence or item, and the delimiter elements are specifically used for sequences or items with an undefined length. Per my interpretation, the delimiter should not be written when the length is specified.

Now, I still did not identify the root cause of the written output being wrong, but this reminds me of #563. It could be a combination of weird input data with an eventual bug in the dataset writing process. I'll get back to this issue if I find out more.

jmlaka:
The (first) offending part of the file is in `(0008,1111) ReferencedPerformedProcedureStepSequence `

I observed how Pydicom and DCMTK handle writing this file back.
(decoding and then encoding back)

When encountering Explicit length Item 
DCMTK does not emit Item NOR Sequence delimiter
Pydicom does not emit Item Delimiter
(Dicom-rs does not emit Item Delimiter)

But both DCMTK and Pydicom recalculate the original Explicit Length, dicom-rs leaves it as is, this produces a lot of
decoding bugs in the very same file. All can be corrected by recalculating the lenghts.

Enet4:
[...]

PR #645, opened 2025-04-11
DataSetWriterOptions - handling of explicit length Sequences and Items

jmlaka:
This adds `DatasetWriterOptions` to the `parser` crate and should handle #636 and similar situations.

The options now contain an `ExplicitLengthSqItemStrategy`.

`ExplicitLengthSqItemStrategy::NoChange` - is the previous behavior, where explicit length items and sequences were left with their length unchanged, which in case of modifying the sequences or items content could lead to incorrect explicit length.
As this situation would occur on a regular basis, and the resulting errors are hard to debug, this is no more the default strategy and risks resulting from it's use are documented.

`ExplicitLengthSqItemStrategy::SetUndefined ` is the new default strategy. It simply writes all Items and Sequences with
`Length::UNDEFINED` irrespective of the original `Length`  enum variant.
Upon inspection of the writer byte output, this will appear counterintuitive to the user.
This is documented.

The final strategy -  recalculating the lengths - is not implemented. 
Reasons are:
 - requires propagation of "bytes_written" value from lower level apis or using some mock writer to count bytes
 - recalculation is expensive
 - current state, even though not optimal and counterintuitive, is error free and DICOM compliant

### DIFF
--- a/parser/src/dataset/write.rs
+++ b/parser/src/dataset/write.rs
@@ -88,6 +88,10 @@ struct SeqToken {
 pub enum ExplicitLengthSqItemStrategy {
     /// All explicit length items and sequences are written with Length::UNDEFINED.
     ///
+    /// Be advised that even if you create or read a data set with explicit length
+    /// items / sequences, the resulting output of the writer will have undefined
+    /// lengths for all items and sequences.
+    ///  
     /// For reasons stated in the documentation of the `ExplicitLengthSqItemStrategy`, this
     /// is as of yet, the fastest and safest way to handle explicit length items and sequences
     /// and thus default.
--- a/parser/src/dataset/write.rs
+++ b/parser/src/dataset/write.rs
@@ -81,6 +81,52 @@ struct SeqToken {
     len: Length,
 }
 
+/// A strategy for writing Sequences and Items if the writer
+/// encounters a Sequence or Item with explicit (defined) length
+#[derive(Debug, Default, Copy, Clone, Eq, Hash, PartialEq)]
+#[non_exhaustive]
+pub enum ExplicitLengthSqItemStrategy {
+    /// All explicit length items and sequences are written with Length::UNDEFINED.
+    ///
+    /// For reasons stated in the documentation of the `ExplicitLengthSqItemStrategy`, this
+    /// is as of yet, the fastest and safest way to handle explicit length items and sequences
+    /// and thus default.
+    #[default]
+    SetUndefined,
+    /// Explicit length items and sequences are written without any change, left as they were encountered
+    /// in the data set.
+    ///
+    /// Lenghts will not be recalculated !
+    ///
+    /// As a consequence, if the content of a sequence or item with explicit length is manipulated after
+    /// it was created or read from a source ( = possibly changes it's real size), this strategy will not
+    /// update the length of the sequence or item and might produce invalid output.
+    NoChange,
+    // Explicit lenght items and sequences could as well be recalculated, as is the behavior
+    // of some DICOM libraries. Because recalculation is expensive and leaving sequences and items
+    // with length undefined is DICOM compliant, this strategy is not implemented (yet).
+    // Racalculate (todo?),
+}
+
+/// The set of options for the data set writer.
+#[derive(Debug, Default, Copy, Clone, Eq, Hash, PartialEq)]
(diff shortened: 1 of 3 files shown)


jmlaka:
Let me just throw some thoughts here.
I think this PR is a quick fix but doesn't achieve the likes of encoding standard comparable to say DCMTK.

To implement the recalculation strategy, I think there are 2 ways to approach this:
1.`DataSetWriter` will `peek()` at the following tokens. Encoder will write to a byte sink, throwing the output away but counting the bytes written. Then set the Sequence length to bytes written. Continue writing / encoding as before ...

2. Instead of a byte sink, encoder will write to memory. After setting the calculated Sequence length. Writer will `write_all() `of the memory buffer. This is more efficient.

encoder is strongly coupled with write target using the `StatefulEncoder`. 

This makes it impossible for the `DataSetWriter `to include a second (and possible third) printer / encoder with a different write target (own Vec, sink). It's because the `StatefulEncoder::new()` moves in and encoder value which, at this point would have to be Clone d and used to encode to a different target. The `EncoderTo `trait object cannot be Clone.

If you would be willing to approve the recalculation strategy, how would you go about changing the current abstractions ?

Could we just make the `StatefulEncoder` have a flag to change the internal `self.encoder` write target, e.g. switch between
`self.to` , `std::io::Sink` and `Vec<u8>` ?


Thank you

Enet4:
You are right, there is a rather tight connection between the encoder and the write destination at the moment. This is also the case for decoders, with its own set of problems alongside it. Performance-wise this has been mitigated through a `BufWriter`, but if I understood correctly, the problem is that we cannot write an accurate length field before we actually go through the whole content associated with that length. For primitive values the solution is trivial, but it stops being so when working with sequence data.

I wouldn't mind relaxing this constraint (as in changing the encoder-writer type handshake), but I suspect this would come with substantial rewriting all the way down from the `dicom-encoding` crate. But maybe there are simpler alternatives.

- `StatefulEncoder` already has a buffer, so we may be able to extend it to support accumulating even more data before it flushes to the encoder.
- Or, with something like the code below, we can create a new stateful encoder which writes the data somewhere else, allowing the data set writer to devise more sequence writing strategies.

```rust
impl<W, E, T> StatefulEncoder<W, E, T> {
    pub fn for_writer<U>(&self, to: U) -> StatefulEncoder<U, E, T>
    where
        U: Write,
        E: Clone,
        T: Clone,
    {
        StatefulEncoder {
            to,
            encoder: self.encoder.clone(),
            text: self.text.clone(),
            bytes_written: self.bytes_written,
            buffer: self.buffer.clone(),
        }
    }
}
```


Let me know if you will want to reiterate on this PR of you'd like it to be admitted as is.

parser/src/dataset/write.rs
```diff
@@ -106,41 +179,45 @@ where
...
             ts_alias: ts.name(),
         })?;
-        Ok(DataSetWriter::new_with_codec(to, encoder, charset))
+        Ok(DataSetWriter::new_with_codec(to, encoder, charset, options))
     }
 }
 
 impl<W, E> DataSetWriter<W, E> {
-    pub fn new(to: W, encoder: E) -> Self {
+    pub fn new(to: W, encoder: E, options: DataSetWriterOptions) -> Self {
```
Enet4, line 204: This is a breaking change, but it can be worked around by adding `new_with_options` instead of changing `new`.

parser/src/dataset/write.rs
```diff
@@ -106,41 +179,45 @@ where
...
             seq_tokens: Vec::new(),
             last_de: None,
+            options,
         }
     }
 }
 
 impl<W, E, T> DataSetWriter<W, E, T> {
-    pub fn new_with_codec(to: W, encoder: E, text: T) -> Self {
+    pub fn new_with_codec(to: W, encoder: E, text: T, options: DataSetWriterOptions) -> Self {
```
Enet4, line 215: This can also exist as `new_with_codec_options`.

Enet4 (commented):
I had overlooked this before, but the changes to the constructor functions in `DataSetWriter` constitute breaking changes. This isn't a problem, though it will require me to postpone merging this until I start collecting everything for v0.9, unless the breaking changes are reverted in the meantime.

PR #646, opened 2025-04-12
Add odd length reading strategy fod Dicom Dump

jmlaka:
added option to change the default "accept" odd length text strategy for the dicom-dump reader.

Comes in handy for ruling out padding / length problems in dicom files.


Enet4 (commented):
Sounds good! Just pointing some suggestion inline.

(you can also try rebasing the branch to make the Clippy errors go away)

Enet4 (approved):
Thanks :+1:

Issue #348, opened 2023-05-13
1-bit pixel data conversion is currently not supported


Enet4:
Example from pydicom tests: liver.dcm

ChayceJRoss:
Hi. First time contributor here. 
1. Am I correct that you would like to add support for when "(0028,0101) Bits Stored" is 1?
2. I have worked with raw bits in rust before and it proved challenging without the help of external libraries (e.g. [std::vector<bool>](https://en.cppreference.com/w/cpp/container/vector_bool)) is this ok?

Enet4:
Thank you for reaching out!

> 1. Am I correct that you would like to add support for when "(0028,0101) Bits Stored" is 1?

Yes, this would be for handing image files where Bits Allocated (and subsequently Bits Stored) are 1. The necessary expansion from bits to bytes would happen when turning the `DecodedPixelData` into a vector, image, or ndarray.

> 2. I have worked with raw bits in rust before and it proved challenging without the help of external libraries (e.g. support for [std::vector<bool>](https://en.cppreference.com/w/cpp/container/vector_bool)) is this ok to use?

You may be able to use [`bitvec`](https://crates.io/crates/bitvec). Once a byte-addressable view is obtained, it would only be a matter of making it go through the established pixel data transformation functions (Modality LUT, VOI LUT).

PR #654, opened 2025-05-31
Support bits_allocated == 1

Cryt1c:
Thanks a lot for this awesome project! I am using it for my learning project to create a DICOM Viewer: https://cryt1c.github.io/DICOMViewer/

Currently I want to extend the DICOMViewer with a feature where users can load DICOM-SEG files. After checking this repo I did not find this functionality or mentions on the issue queue, so I have tinkered around a bit and made it work. I would love to contribute this back upstream.

I am basically handling the case where bits_allocated is 1. In the decode_pixel_data_frame where each byte gets mapped to 8 separate bits which are handled as one byte each. Values are mapped to 0..255 so that the result should be a black and white segmentation mask.

We should probably handle the DICOM-SEG more explicitly or even create a separate crate for it? Please let me know what your preferred way would be.

I am happy to adjust the PR on your feedback.

### DIFF
--- a/pixeldata/src/lib.rs
+++ b/pixeldata/src/lib.rs
@@ -2275,7 +2275,7 @@ where
                 let frame_pixels = (rows as usize) * (cols as usize);
                 let frame_samples = frame_pixels * (samples_per_pixel as usize);
                 let frame_size = if bits_allocated == 1 {
-                    frame_samples / 8 as usize
+                    frame_samples / 8
                 } else {
                     frame_samples * ((bits_allocated + 7) / 8) as usize
                 };
--- a/pixeldata/src/lib.rs
+++ b/pixeldata/src/lib.rs
@@ -468,7 +468,7 @@ impl DecodedPixelData<'_> {
     /// Retrieve a slice of a frame's raw pixel data samples as bytes,
     /// irrespective of the expected size of each sample.
     pub fn frame_data(&self, frame: u32) -> Result<&[u8]> {
-        let bytes_per_sample = self.bits_allocated as usize + 7 / 8;
+        let bytes_per_sample = (self.bits_allocated as usize + 7) / 8;
         let frame_length = self.rows as usize
             * self.cols as usize
             * self.samples_per_pixel as usize
--- a/pixeldata/src/lib.rs
+++ b/pixeldata/src/lib.rs
@@ -169,7 +169,7 @@ pub enum InnerError {
     #[snafu(display("PixelData attribute is not a primitive value or pixel sequence"))]
     InvalidPixelData { backtrace: Backtrace },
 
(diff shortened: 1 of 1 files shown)


pixeldata/src/lib.rs
```diff
@@ -2268,17 +2272,26 @@ where
             }
             DicomValue::Primitive(p) => {
                 // Non-encoded, just return the pixel data for a single frame
-                let frame_size = ((bits_allocated + 7) / 8) as usize
-                    * samples_per_pixel as usize
-                    * rows as usize
-                    * cols as usize;
-                let frame_offset = frame_size * frame as usize;
+                let frame_bits = rows as usize * cols as usize;
```
Enet4, line 2275: `frame_bits` is only truly the number of bits in a frame if _Bits Allocated_ is 1 and _Samples Per Pixel_ is 1, it breaks in most other scenarios.

Perhaps we can redo the calculation starting from the number of pixels, then samples, and finish it off with the total number of pixels according to bits allocated.

```rust
let frame_pixels = rows as usize * cols as usize;
let frame_samples = frame_pixels * samples_per_pixel as usize;
let frame_size = if bits_allocated == 1 {
    frame_samples / 8
} else {
    frame_samples * ((bits_allocated + 7) / 8) as usize;
};
```

Enet4 (changes requested):
Thank you for working on this! It appears to bring support for images where bits allocated = 1 (#348). Please see if you can address the two concerns listed below.

- It breaks other cases such as when there is more than one sample per pixel (tests fail too: run `cargo test -p dicom-pixeldata --features=image,ndarray`). I have left some advice inline.
- We should extend this test coverage to include a file of a similar nature. `liver.dcm` from the pydicom dataset has _Bits Allocated_ = 1, so it might do the trick.

> We should probably handle the DICOM-SEG more explicitly or even create a separate crate for it? Please let me know what your preferred way would be.

That would probably fit as its own crate. Though the project at the moment does not offer many IOD abstractions on top the other generic crates, the pattern would eventually exist as additional crates in the DICOM-rs umbrella.

Cryt1c:
I have fixed the `frame_size `calculation and implemented `test_1bit_image_decoding`.
The image decoding works but the resulting image looks a bit strange:
liver-0

Cryt1c:
The bits in the bit to byte conversions were in the wrong order. Now it looks good:
liver-0

Enet4 (approved):
Looks great! It is not fully clear whether the expansion of 1-bit pixel data values to the 0-255 range will come across as surprising to some users, but it shouldn't stop them from working around the pixeldata abstraction if this is a concern to them.

Thank you and welcome to the list of DICOM-rs contributors. I also want to thank you for giving a shout out to DICOM-rs last year at EuroRust 2024! 🦀

PR #657, opened 2025-06-29
DICOM collector API

Enet4:
The DICOM collector API in `DicomObjectCollector` can be used for reading DICOM objects in contiguous chunks, thus being able to collect meta-data earlier and reducing memory footprint by not having to keep the full data set in memory.

### Summary

- [parser] add `LazyDataSetReader::peek` and `LazyDataSetReader::new_with_ts`
- [parser] change `Debug` impl for `LazyDataToken` so that it works for any data source
- [object] add collector module and `DicomCollector` abstraction
   - allows users to fetch DICOM data in chunks
   - pub use `DicomCollector` and `DicomCollectorOptions` at crate root
- Adjust documentation a bit


object/src/collector.rs
```diff
@@ -0,0 +1,1189 @@
...
+            let ts = self.guessed_ts().context(GuessTransferSyntaxSnafu)?;
+            self.source.set_parser_with_ts(ts)?
+        } else {
+            self.source.parser()
+        };
+
+        // proceed with fetching tokens,
+        // return the first fragment data found
+        while let Some(token) = parser.advance() {
+            match dbg!(token.context(ReadTokenSnafu)?) {
```
Enet4, line 574:
```suggestion
            match token.context(ReadTokenSnafu)? {
```

object/src/collector.rs
```diff
@@ -0,0 +1,1189 @@
...
+        }
+
+        Ok(None)
+    }
+
+    // --- private methods ---
+
+    fn guessed_ts(&mut self) -> Option<&'t TransferSyntax> {
+        if self.ts_hint.is_some() {
+            return self.ts_hint.clone();
```
Enet4, line 605:
```suggestion
            return self.ts_hint;
```

object/src/collector.rs
```diff
@@ -0,0 +1,1189 @@
...
+    // --- private methods ---
+
+    fn guessed_ts(&mut self) -> Option<&'t TransferSyntax> {
+        if self.ts_hint.is_some() {
+            return self.ts_hint;
+        }
+        if let Some(meta) = self.file_meta.as_ref() {
+            self.ts_hint = TransferSyntaxRegistry.get(meta.transfer_syntax());
+        }
+        self.ts_hint.clone()
```
Enet4, line 610:
```suggestion
        self.ts_hint
```

PR #658, opened 2025-07-01
Move-Scu tool

knopkem:
Based on the existing c-find and c-store tools. Rust beginner here, don't hesitate to correct things.


movescu/src/main.rs
```diff
@@ -0,0 +1,540 @@
...
+                        warn!("{} Identifier does not match SOP class in C-MOVE response", msg);
+                    } else if status == 0xc000 {
+                        warn!("{} Unable to process C-MOVE response", msg);
+                    } else if status == 0xfe00 {
+                        warn!("{} Sub-operations terminated due to cancel indication", msg);
+                    } else if status == 0xb000 {
+                        warn!("{} Sub-operations complete with one or more failures", msg);
+                    } else {
+                        warn!("{} Unknown status code", msg);
+                    }
```
Enet4, line 465: Not very important, but this would be cleaner and involve even less allocations.
```suggestion
                    let msg = match status {
                        0xa701 => "Out of resources (number of matches)",
                        0xa702 => "Out of resources (sub-operations)",
                        0x0122 => "SOP class not supported",
                        0xa801 => "Move destination unknown",
                        0xa900 => "Identifier does not match SOP class in C-MOVE response",
                        0xc000 => "Unable to process C-MOVE response",
                        0xfe00 => "Sub-operations terminated due to cancel indication",
                        0xb000 => "Sub-operations complete with one or more failures",
                        _ => "Unknown status code",
                    };
                    warn!("Operation failed (status code {status:x}) {msg}");
```

movescu/src/main.rs
```diff
@@ -0,0 +1,540 @@
...
+                        let mut response_data = Vec::new();
+                        rsp.read_to_end(&mut response_data)
+                            .whatever_context("Failed to read response data")?;
+
+                        InMemDicomObject::read_dataset_with_ts(&response_data[..], ts)
+                            .whatever_context("Could not read response data set")?
+                    };
+
+                    println!(
+                        "------------------------ Match #{i} ------------------------"
```
Enet4, line 424: I wonder what constitutes a match here.
Per [PS3.9 section 9.1.4.2](https://dicom.nema.org/medical/dicom/current/output/chtml/part07/chapter_9.html#sect_9.1.4.2), the responses from the Move SCP are not exactly "matches" but rather information about the situation with the move, so what we usually receive are "status pending" messages and a final response before closing. And some implementations only do the latter without waiting for the transfer to occur, so I've had this tool telling me that there were no matches, even though the request was processed successfully.

This match printing logic is probably a remnant of C-FIND, and it should be translated to the C-MOVE use case. My suggestion would be showing a progress indicator that ticks every time a status pending response emerges, and stopping depending on the response status.

Enet4 (commented):
Hello @knopkem! I was close to approving this pull request two weeks ago, but I encountered some oddities that called for further investigation. Can you please see the comments inline, especially the one about matches? Feel free to ask any follow-up questions.

Enet4 (commented):
Would you be able to pick this up again, or would you prefer that I take over?

movescu/src/main.rs
knopkem: Hey, sorry that there was no progress on this, but didn't find much time for it recently and also I'm not very proficient with rust (actually I wanted to base something on your lib to learn rust). So yes, would be good to get this merged but not sure when I find time for it, so maybe it would be better if you polish the rest of it. Thanks

PR #666, opened 2025-07-25
[partial] Improve fraction retrieval for DicomTime with precision checks 

pongis:
Hi,
i picked #662 and implemented a solution where leading zeros are not omitted and the value representation of the Date time remains correct. I also added 2 more tests for my change. First time adding to Open Source, hope this helps.

I am not sure if `Option<&u32>` is the right choice as return type, or if i should create a new error variant if a mismatch between frac and frac_precision occurs.

### DIFF
--- a/core/src/value/deserialize.rs
+++ b/core/src/value/deserialize.rs
@@ -3,7 +3,7 @@ use crate::value::partial::{
     check_component, DateComponent, DicomDate, DicomDateTime, DicomTime,
     Error as PartialValuesError,
 };
-use chrono::{DateTime, FixedOffset, NaiveDate, NaiveDateTime, NaiveTime, TimeZone};
+use chrono::{FixedOffset, NaiveDate, NaiveTime};
 use snafu::{Backtrace, OptionExt, ResultExt, Snafu};
 use std::convert::TryFrom;
 use std::ops::{Add, Mul, Sub};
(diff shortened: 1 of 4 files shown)


Enet4:
Thank you for the initiative! In the meantime I have already merged #665, but this can make a great addition to the 0.9 milestone!

> I am not sure if Option<&u32> is the right choice as return type, or if i should create a new error variant if a mismatch between frac and frac_precision occurs.

An enum might be better, as it forcefully narrows down the precision depth to valid cases (10th of a second, 100th of a second, millisecond, ...). But this can stay an implementation detail as long as the API is expressive and flexible enough. Right now there are a few cases where it's still hard to retrieve the fraction of a second, and relaxing it to enable fetching this fraction regardless of precision may be a good step towards that.

Moreover, any method in this module returning `Option<&u32>` should return `Option<u32>` instead. In this case there is no added value to returning the reference to the value in the `DicomTime`, and `Option<u32>` is more compact and nicer to work with.

Enet4:
I took the liberty of reworking this pull request so that hopefully it will be easier to work with fractions of a second.

- I have decided to remove the `fraction` method entirely due to its ambiguity and difficulty of use. The former implementation (pre-PR) was particularly tricky to use because it required the fraction to be precise to the microsecond, which some devices do not even produce. And the implementation originally suggested in this PR would also require the user to consult the precision for it to be useful.
- So instead, we now have `fraction_ms` and `fraction_micro`, which always return `Some(u32)` unless the time does not specify any fraction of a second.
- I also added a few extra methods and trait implementations which make complete sense but were missing for some reason: `FromStr` for `DicomDate` and `DicomTime`.

PR #673, opened 2025-08-16
[pixeldata] Correct photometric interpretation in transcoding process

Enet4:
After decoding, the photometric interpretation often needs to be updated accordingly, as the pixel data reader trait requires so in its contract. This step was missing in the transcoder module of `dicom_pixeldata`.

### DIFF
--- a/pixeldata/src/transcode.rs
+++ b/pixeldata/src/transcode.rs
@@ -419,9 +419,21 @@ mod tests {
         );
     }
 
-    #[cfg(feature = "native")]
+    #[cfg(all(feature = "jpeg", not(feature = "gdcm")))]
     #[test]
-    fn test_transcode_from_jpeg_baseline_to_native_rgb() {
+    fn transcode_from_jpeg_baseline_to_native_rgb() {
+        transcode_from_jpeg_baseline_to_native_rgb_impl();
+    }
+
+    /// !!! #674 Decoding via GDCM is buggy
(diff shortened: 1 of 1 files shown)


pixeldata/src/transcode.rs
```diff
@@ -384,6 +410,89 @@ mod tests {
...
+    #[cfg(all(feature = "jpeg", not(feature = "gdcm")))]
+    #[test]
+    fn transcode_from_jpeg_baseline_to_native_rgb() {
+        transcode_from_jpeg_baseline_to_native_rgb_impl();
+    }
+
+    /// !!! #674 Decoding via GDCM is buggy
+    #[cfg(all(feature = "jpeg", feature = "gdcm"))]
+    #[test]
+    #[skip]
```
Enet4, line 431:
```suggestion
    #[ignore]
```

pixeldata/src/transcode.rs
```diff
@@ -384,6 +410,89 @@ mod tests {
         assert_eq!(pixels.len(), rows * cols * spp);
     }
 
+    fn assert_sample_eq_approx(description: &str, found: u8, expected: u8) {
```
Enet4, line 413:
```suggestion
    #[cfg(feature = "jpeg")]
    fn assert_sample_eq_approx(description: &str, found: u8, expected: u8) {
```

PR #678, opened 2025-08-26
Add  SpecificCharacterSet Support GBK GB2312 ISO 2022 IR 58

momostarsky:
Add  GBK , GB2312, ISO 2022 IR 58 characterset Support!


Enet4:
Yes, that warning is correct, the enum variant name should be `Gbk`.

PR #679, opened 2025-08-29
Tls implementation

naterichman:
* Changes to application binaries:
  * Import of new traits
  * Change path of some errors
  * Change name of association methods (`peer_ae_title`, `max_pdu_length`)
* UL library Changes
  * Features:
    * Added rustls behind `tls` feature flag
    * Added `tokio-rustls` behind `async-tls` feature flag
  * Association
    * Added common `Association` trait for common (non-io) features
    * Added separate `SyncAssociation` and `AsyncAssociation` for
      stream-related features.
    * Added similar respective private traits for "unsafe" methods
      which are needed but should not be called by users
    * Separated both `ServerAssociation` and `ClientAssociation` into
      two different structs (async vs sync) and implemented respective
      traits
    * Added tls config options for both `ServerAssociationOptions` and
      `ClientAssociationOptions`
    * Added methods on both `AssociationOptions` structs to establish
      a TLS connection both sync and async.
    * Implemented `Drop` for both `ServerAssociation` and `AsyncServerAssociation`
      * NOTE: This is using `abort` whereas the client methods use `release`
    * Reorganized both client and server to remove separate non_blocking mods
  * Tests:
    * Added full test for mTLS including script to generate certificates and 
      instructions
  * Documentation
    * Added documentation on how to use mTLS

PR #685, opened 2025-09-16
Add ServerAssociation::requestor/acceptor_max_pdu_length()

### LINKED ISSUE #684: Make the remote PDU max length available to applications
The maximum PDU length for the remote is necessary for splitting PDUs manually. However, the field containing it is private.

pgimeno4d:
Fixes #684 by adding a reader method.

### DIFF
--- a/ul/tests/association_echo.rs
+++ b/ul/tests/association_echo.rs
@@ -1,11 +1,19 @@
 use dicom_ul::{
-    association::client::ClientAssociationOptions,
-    pdu::{Pdu, PresentationContextNegotiated, PresentationContextResultReason},
+    association::{Error, client::ClientAssociationOptions},
+    pdu::{Pdu, PDataValue, PDataValueType, PresentationContextNegotiated,
+        PresentationContextResultReason},
 };
 
 use std::net::SocketAddr;
+use rand::Rng;
 
 use dicom_ul::association::server::ServerAssociationOptions;
+
+// Check rather arbitrary maximum PDU lengths, also different for server and client
+const HI_PDU_LEN: usize = 7890;
+const LO_PDU_LEN: usize = 5678;
+const PDU_HDR_LEN: usize = 6;
+const PDV_HDR_LEN: usize = 6;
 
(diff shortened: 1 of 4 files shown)


pgimeno4d:
This will be obsoleted by #679. If the plan is to merge it, it can be closed.

Enet4:
Given that v0.9.0 is overdue and #679 does not introduce breaking changes, it would make sense to contribute with these methods separately and leave TLS support for v0.9.1. It would be great if you could rebase this pull request on top of #689, so that these can be merged in order.

ul/tests/association_echo.rs
```diff
@@ -48,6 +73,41 @@ fn spawn_scp() -> Result<(std::thread::JoinHandle<Result<()>>, SocketAddr)> {
...
+            .send(&packet)
+            .expect("failed sending packet");
+
+        // Add 1 byte to the payload to exceed maximum length
+        if let Pdu::PData { ref mut data } = packet {
+            data[0].data.push(0);
+        }
+
+        match association.send(&packet) {
+            Err(server::Error::SendTooLongPdu {..}) => (),
```
Enet4, line 107: The same error type now lives at the parent module.

Enet4 (commented):
Thanks! Can you also look into the check failures?

pgimeno4d:
I'm not sure why it's failing now. It appears to not be compiling the test in async mode or something like that.

pgimeno4d:
> As for the other failing tests, I did the quick experiment of splitting the two scenarios (max_is_server/max_is_client) for each sync/variant, and only the cases where max_is_client=true failed.

Yes, as predicted. It is testing something that is currently broken in master. Once #689 is merged, which fixes the problem, and this one rebased on top of it, it will pass. As I said, I could not rebase it on #689 because that made a mess of commits.

pgimeno4d:
This is ready to be merged. I was lucky to make the last commit after 689 was merged, so the CI already included it and the tests passed without needing a rebase.

Enet4 (approved):
Indeed, can confirm that the tests pass after the merge. Thank you for the continued support on this @pgimeno4d. :+1:

PR #686, opened 2025-09-16
Fix a pair of related bugs in Server/ClientAssociation::receive

pgimeno4d:
The problem can be quickly spotted by comparing the sync and async versions of ClientAssociation::receive() in the source:

Sync version:
https://github.com/Enet4/dicom-rs/blob/1f16e9bf785d6cfc90820f1b4605034134d24e80/ul/src/association/client.rs#L988

Async version:
https://github.com/Enet4/dicom-rs/blob/1f16e9bf785d6cfc90820f1b4605034134d24e80/ul/src/association/client.rs#L1498

One says "requestor" and the other "acceptor"? That can't be right; both sync and async should use the same criteria. So which one is right? It's checking what "we" are able to receive, and since the client is the one that makes requests, and thus the requestor, requestor is right and acceptor is wrong. That is, the sync version is wrong.

The exact same problem happens in ServerAssociation::receive:

Sync version:
https://github.com/Enet4/dicom-rs/blob/1f16e9bf785d6cfc90820f1b4605034134d24e80/ul/src/association/server.rs#L741

Async version:
https://github.com/Enet4/dicom-rs/blob/1f16e9bf785d6cfc90820f1b4605034134d24e80/ul/src/association/server.rs#L1158

The server is the acceptor, so in this case the async version is wrong.

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -985,7 +985,7 @@ where
 
         loop {
             let mut buf = Cursor::new(&self.read_buffer[..]);
-            match read_pdu(&mut buf, self.acceptor_max_pdu_length, self.strict)
+            match read_pdu(&mut buf, self.requestor_max_pdu_length, self.strict)
                 .context(ReceiveResponseSnafu)?
             {
                 Some(pdu) => {
--- a/ul/src/association/server.rs
+++ b/ul/src/association/server.rs
@@ -1155,7 +1155,7 @@ pub mod non_blocking {
             let task = async {
                 loop {
                     let mut buf = Cursor::new(&self.read_buffer[..]);
-                    match read_pdu(&mut buf, self.requestor_max_pdu_length, self.strict)
+                    match read_pdu(&mut buf, self.acceptor_max_pdu_length, self.strict)
                         .context(ReceiveRequestSnafu)?
                     {
                         Some(pdu) => {


naterichman:
I think this should be fixed in #679 [see here](https://github.com/Enet4/dicom-rs/pull/679/files#diff-1f73bda45d3f6e93ca0c5a0cd23153ca8bd53491f9bce0c5c5e6e41f2689caffR831)

I changed the Association structs to store a single `max_pdu_length` which is the minimum of the configured requestor, and the acceptor, then this is used in all of the `read_pdu`-esque methods

pgimeno4d:
I agree that #679 fixes it. I'm not sure to fully agree with the conflation of our and the peer's max_pdu_length, but it's not wrong either, so no objections.

I'm not sure whether to keep this open. I guess it depends on whether there will be a 0.8.3. If not, please feel free to close.

Enet4:
> I'm not sure whether to keep this open. I guess it depends on whether there will be a 0.8.3. If not, please feel free to close.

The next version planned is 0.9.0, as we've already started merging breaking changes. On the other hand, I have yet to to make the complete assessment of which PRs will enter. Only those on the [0.9 milestone](https://github.com/Enet4/dicom-rs/milestone/3) are practically guaranteed to be included, but I might add more in the meantime.

As for the release proper, I understand that it's been a while since the last one, so hopefully I can drop it before the end of the month. Let's see how things go. :)

Enet4:
I will consider #689 instead, which fixes this bug and improves a few other things around max PDU length negotiation. Thank you for working on this nevertheless!

Issue #683, opened 2025-09-15
Provide means for obtaining abstract syntaxes from ServerAssociation


pgimeno4d:
As it is, it's impossible to implement a compliant C-GET SCP using the dicom_ul module, based on how it should work:

- The SCU sends the abstract syntaxes and their corresponding transfer syntaxes during association request.
- The server responds with a list of presentation context IDs and transfer syntaxes during association acceptance. **At this point, the library drops any info on which  abstract syntax is associated with each presentation context.**
- The SCU sends a C-GET-RQ command and a query.
  - At this point, it's not possible to verify that the Affected SOP Class UID sent in the C-GET-RQ matches the negotiated Service Class sent in the Presentation Context. If the Affected SOP Class UID happens to be missing because the client is non-compliant, there's no way of knowing the requested Service Class. This is a minor problem.
- The SCP matches the query against the available datasets and responds with a C-GET-RSP.
- The SCP should now send a C-STORE command over the same association. However, due to the missing abstract syntax information:
  - It's not possible to match the SOP Class UID of the instance to send to a presentation context, in order to correctly set the Presentation Context ID field of the PDV for the C-STORE command. This is a major problem, as the only possible thing that can be done is setting an arbitrary PC ID and hoping that the client won't complain about the protocol violation.
  - It's not possible to select the correct transfer syntax that was negotiated for that UID for the transfer. This is a major problem.
  - It's not possible to reject the transfer if the abstract syntax was not present or was rejected. This is a major problem.

Please add some means for the ServerAssociation to allow retrieving the abstract syntax corresponding to each Presentation Context in order to solve these issues.

I have a patch; the only missing part is deciding on the format of the data to return. I've implemented it as a `HashMap<String, u8>` which maps abstract syntax UIDs to presentation context IDs, but that's probably overkill and makes it more difficult to find the abstract syntax from the PC ID, as would be needed for the first problem mentioned above.

Enet4:
Thank you for reporting. Indeed, the abstract syntax part is not included and is not saved by the association acceptor at the moment. I'd say that it's fine that we start saving that somewhere in `ServerAssociation`. Considering the anticipated spectrum of DICOM abstract syntaxes, I do not anticipate performance issues from saving this in a plain `Vec<(u8, String)>`, though anything that gives us bidirectional look-up (PC ID → AS and AS → PC ID) would be OK.

pgimeno4d:
[...]

PR #687, opened 2025-09-17
Return the abstract syntax in the presentation_contexts() method

pgimeno4d:
This is finally a fix for #683 that I'm comfortable with. The only problem is that it's a breaking change.

The return type of presentation_contexts() is now `&[PresentationContextNegotiated]`, and includes the abstract syntax that was negotiated on each PC. The PresentationContextResult struct remains for protocol handling. This avoids kludges such as using None for the field during internal handling, or adding another separate vector to relate abstract syntax and PC ID.

A broken (or rogue) server may return PC IDs that are not in the list of PCs proposed by the client. These are now filtered out, as it would be impossible to assign an abstract syntax to them.

This adds a description in the docs to PresentationContextResult, while on it.

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -16,7 +16,7 @@ use std::{
 use crate::{
     pdu::{
         read_pdu, write_pdu, AbortRQSource, AssociationAC, AssociationRJ, AssociationRQ, Pdu,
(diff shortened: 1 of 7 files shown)


Enet4:
This looks rather sound and not too complicated, so I would say it's fine to enter in 0.9.0.

PR #720, opened 2026-01-02
MAIN: Don't use all rustls default features

### LINKED ISSUE #719: Latest version of dicom-rs, forces users, to use `awc-lc-rs` which is very hard to cross-compile
https://github.com/Enet4/dicom-rs/blob/22507e3d2740d0e6438eb924379775075028ba3f/ul/Cargo.toml#L20

By default, the awc-lc-rs dependency is used, which is very hard to cross-compile (I’m trying to build a Windows binary on Linux).

In all places, I use rustls without default features, which allows me to use ring. I think this repository should also allow choosing which crypto library is used.

Image

naterichman:
* This disables ring and aws-lc-rs  in the `ul` library which allows users of the library to choose their own cryptoprovider
* The binaries (storescu/storescp/etc) still use aws-lc-rs as default since the `app-common` package specifies rustls with default features.

Fixes #719

### DIFF
--- a/ul/Cargo.toml
+++ b/ul/Cargo.toml
@@ -18,7 +18,6 @@ cfg-if = "1.0.3"
 dicom-encoding = { path = "../encoding/", version = "0.9" }
 dicom-transfer-syntax-registry = { path = "../transfer-syntax-registry/", version = "0.9", default-features = false }
 snafu = "0.8"
-tokio-rustls = {version = "0.26.2", optional = true }
 tracing = "0.1.34"
 
 [dependencies.tokio]


ul/Cargo.toml
```diff
@@ -32,6 +31,16 @@ features = [
...
     "time"
 ]
+[dependencies.rustls]
+version = "0.23.31"
+optional = true
+default-features = false
+features = [
+  "logging",
+  "tls12",
+  "prefer-post-quantum",
```
Enet4: Only aws-lc-rs supports this feature at the moment: https://github.com/rustls/rustls/issues/2801

```suggestion
```

ul/Cargo.toml
```diff
@@ -32,6 +31,16 @@ features = [
...
+  "std"
+]
```
Enet4: `tokio-rustls` also includes aws_lc_rs, so we need to disable it there as well.

```suggestion
]

[dependencies.tokio-rustls]
version = "0.26.2"
optional = true
default-features = false
features = [
  "logging",
  "tls12",
]
```

Enet4 (approved):
Thank you!

Issue #714, opened 2025-11-19
Update state machine; send/receive P-DATA PDUs only


pgimeno4d:
The implementation of the Drop trait for `[Async](Server/Client)Association` has resulted in problems when the association is released.

An easy way to reproduce is to use a wire dump tool such as wireshark, then open a dicom-storescp server and use the dicom-echoscu on it. Example output with wireshark, using `dicom` as display filter follows. The client is 10.0.0.11 and the server is 10.0.128.11.

```
No.     Time           Source                Destination           Protocol Length Info
      4 0.000197660    10.0.0.11             10.0.128.11           DICOM    318    A-ASSOCIATE request ECHOSCU --> ANY-SCP
      6 0.000531384    10.0.128.11           10.0.0.11             DICOM    276    A-ASSOCIATE accept  ECHOSCU <-- ANY-SCP
      8 0.000929172    10.0.0.11             10.0.128.11           DICOM    148    P-DATA, C-ECHO-RQ ID=1
      9 0.001465826    10.0.128.11           10.0.0.11             DICOM    132    P-DATA, C-ECHO-RSP ID=1 (Success)
     10 0.008328674    10.0.0.11             10.0.128.11           DICOM    78     A-RELEASE request
     11 0.008474849    10.0.128.11           10.0.0.11             DICOM    78     A-RELEASE response
     13 0.008605302    10.0.128.11           10.0.0.11             DICOM    78     ABORT ECHOSCU <-- ANY-SCP (Not specified)
```
The problem is the last packet, numbered 13. The association is already released; it makes no sense to send an abort.

This appears to be a consequence of the `Drop` trait implementation for `[Async]ServerAssociation`, which unconditionally performs an `abort()`.

Similarly, the `Drop` trait for [Async]ClientAssociation unconditionally performs a `release()`.

Not sure how this should be handled. I think that removing the Drop trait and leaving the release or abort up to the user is the natural thing to do. Another, more complex option is to keep the association's state machine properly updated and take it into account at drop time, to see if the association is still open before deciding whether to abort or release it.

Pinging @naterichman as the author of the commit that introduces the `Drop` trait for the association objects.

pgimeno4d:
[...]

PR #728, opened 2026-01-29
Remove drop traits

pgimeno4d:
Implements point 1 of https://github.com/Enet4/dicom-rs/issues/714#issuecomment-3664462480. There was a unit test that seemed to rely on packets being sent upon dropping the association object, so I had to fix those too. However I'm not very clear on what's going on exactly, as the functions they use seem to return `Err(ConnectionClosed)` in some instances, which doesn't seem to affect the test.

If all is OK, I will remove the FIXME's.

### DIFF
--- a/ul/src/association/tests.rs
+++ b/ul/src/association/tests.rs
@@ -301,7 +301,7 @@ mod successive_pdus_during_client_association {
                 .accept_any()
                 .with_abstract_syntax(VERIFICATION)
                 .ae_title("THIS-SCP");
-                
+
             let association = server_options.establish_with_extra_pdus(stream, vec![echo_pdu]).unwrap();
             association
         });
@@ -319,7 +319,7 @@ mod successive_pdus_during_client_association {
         let mut association = scu_options.broken_establish(server_addr.into()).unwrap();
 
         // Initiate abort server side
-        server_handle.join().unwrap();
+        server_handle.join().unwrap().abort().unwrap();
         
         // Client should not have anything to receive, only receives abort Rq
         let received_pdu = association.receive().unwrap();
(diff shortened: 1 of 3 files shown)


ul/src/association/tests.rs
```diff
@@ -379,7 +379,7 @@ mod successive_pdus_during_client_association {
             server_addr.into()
         ).await.unwrap();
         // Initiate abort server-side
-        server_handle.await.unwrap();
+        let _ = server_handle.await.unwrap().abort().await; // FIXME: result is Err(ConnectionClosed)
```
Enet4: Not sure if I did something unintended, but unwrapping this result did not raise the error reported in the comment.

It _did_ happen in the other places where the result is explicitly ignored.

Enet4 (commented):
>  However I'm not very clear on what's going on exactly, as the functions they use seem to return `Err(ConnectionClosed)` in some instances, which doesn't seem to affect the test.

These tests were made to cover the unlikely scenario of an SCU sending more data than it should during the A-ASSOCIATE handshake. This does not seem to be a critical change in behavior, though it would be nice to understand why those results had to be ignored.

ul/src/association/tests.rs
pgimeno4d: I seem to recall this being prone to failures, failing under some circumstances and working under others. But as it is now now, I can't get it to fail again at all. The other one however does fail. I'll `.unwrap()` this one.

Enet4 (approved):
This is now considered acceptable. Despite the significant change in behavior, it is safe to say that this is well preferred over the current state. Thank you again!

pgimeno4d:
If I understand the comments correctly, the [A]SyncAssociationSealed traits in the `private` module in association/mod.rs ([link](https://github.com/Enet4/dicom-rs/blob/467c07a6e6564b4357c851b41cf5c50222c06fc3/ul/src/association/mod.rs#L278-L393)) are there to work around issues with abort() and release() in combination with the Drop trait, and thus they should no longer be necessary after this patch. These methods don't properly deal with the release or abortion of the association according to the state machine, but to simplify the changes later, it'd be better to first reorganize the code to get rid of that `private` mod, if that's OK.

Enet4:
Sure, we're in good position to refactor the code again for the next release. (perhaps even as part of #734?)

Issue #731, opened 2026-02-02
Allow role selection for association requests


stefan-hubert-bl:
**Context:** `dicom_ul::ClientAssociation`

**Why:** In order to implement C-Get (client) in a way conforming to DICOM Standard, there needs to be proper configuration of role selection in the A-ASSOCIATE-RQ. This seems to be specifically required/important for C-Get (as opposed to C-Store or C-Find), because, as far as I understand, for C-Get the AE that requests the association (the client) wants to use the association as SCU for the C-Get service class, but at the same time as SCP for the C-Store sub operations that it wishes to come back. Our verification with `dcmqrscp` (command line tool of https://dicom.offis.de/dcmtk.php.de) showed, that it does not properly serve C-Get operations without the association request declaring the roles accordingly, but did successfully served them when role selection was configured (we were using a locally patched version of dicom-ul with the requested feature added).

**What:** I wish I could use ClientAssociation[Options] to configure role selection, so that A-ASSOCIATE-RQ is sent with role selection configured.

**Background/References**
* [CP/SCU Role Selection Negotiation](https://dicom.nema.org/medical/dicom/2022b/output/chtml/part07/sect_D.3.3.4.html)
* [C-GET Operation](https://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_C.4.3.html)

stefan-hubert-bl:
@pgimeno4d: Great, sorry I didn't previously understand that you already working on that. In case you're still interested in what I attempted so far, find it there: https://github.com/stefan-hubert-bl/dicom-rs/pull/1 .  Looking forward to see and use yours.

pgimeno4d:
@stefan-hubert-bl Indeed. Your patch and mine essentially do the same thing, except that instead of the server blindly accepting whatever the client proposes (by returning the same as the client sent) during the negotiation, said negotiation is performed through a callback that the server can set, giving it a chance to reject some, which is the whole point of the existence of these user variables. The only problem is that it depends on a previous PR (#734) to be accepted, which implements the trait that allows the callbacks to be defined. The purpose of having that negotiation and the default values is not very clear to me, but whatever; it's just one of these rarities of the standard.

My patch is a bit more "anarchic-style" than yours in that it uses bools instead of enums and tuples instead of structs, but that's something I'm willing to change if Enet4 deems that unacceptable.

[...]

PR #734, opened 2026-02-03
[ul] Implement generic Extended Negotiation

pgimeno4d:
Add a `Negotiation` trait which contains, for the moment, the `extended_negotiation` method. In future it can also help implement Role Selection negotiation (#731) and possibly other currently unimplemented features like Common Extended Negotiation. Note I'm also interested in #731 and will probably submit a patch if this one is merged, as it will depend on said trait.

This is a case where the client and the server differ substantially. The `Negotiation` trait is only for the server; the client just adds pairs SOP Class / Byte array. After the association succeeds, they both have the same set of pairs, which they can look up to determine which extended options are active, through the new method `extended_negotiation_for(sop_class_uid: &str)` or by walking the `user_variables()`.

Everything is quite low-level, to cover for all current and hopefully future versions of the standard. It's easy to send invalid bytes, for instance.

I'm not quite sure if this is a breaking change. It did not break anything in our application, which is a server. The addition of a new generic concerned me a bit.

Edit: Forgot to mention, this obviously depends on #727, otherwise the bytestream is not correct.

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -16,8 +16,8 @@ use std::{
 use crate::association::AsyncAssociation;
 use crate::{
     association::{
-        encode_pdu, private::SyncAssociationSealed, read_pdu_from_wire, Association,
-        NegotiatedOptions, SocketOptions, SyncAssociation,
+        encode_pdu, extended_negotiation_filter_impl, private::SyncAssociationSealed,
+        read_pdu_from_wire, Association, NegotiatedOptions, SocketOptions, SyncAssociation,
     },
     pdu::{
         write_pdu, AbortRQSource, AssociationAC, AssociationRQ, Pdu, PresentationContextNegotiated,
@@ -263,6 +263,8 @@ pub struct ClientAssociationOptions<'a> {
     saml_assertion: Option<Cow<'a, str>>,
     /// User identity JWT
     jwt: Option<Cow<'a, str>>,
+    /// Extended Negotiation info
+    extended_negotiation: Vec<(Cow<'a, str>, Vec<u8>)>,
     /// Socket options for TCP connections
     socket_options: SocketOptions,
     /// TLS configuration to use for the connection
@@ -292,6 +294,7 @@ impl Default for ClientAssociationOptions<'_> {
             kerberos_service_ticket: None,
             saml_assertion: None,
             jwt: None,
+            extended_negotiation: Vec::new(),
             socket_options: SocketOptions {
                 read_timeout: None,
                 write_timeout: None,
@@ -496,6 +499,22 @@ impl<'a> ClientAssociationOptions<'a> {
         self
     }
 
+    /// Add an Extended Negotiation SOP Class UID and raw data.
+    /// The data must be a byte slice with the information
+    /// required by the standard for the given SOP Class. This is
+    /// kind of a low-level function, intended to support all
+    /// possible current and future use cases for this feature,
+    /// so it's the caller's responsibility to fill in this field
+    /// correctly.
+    pub fn with_extended_negotiation<T>(mut self, sop_class: T, scai: &[u8]) -> Self
+    where
+        T: Into<Cow<'a, str>>,
+    {
+        self.extended_negotiation
+            .push((sop_class.into(), scai.to_vec()));
+        self
+    }
+
     /// Set the TLS configuration to use for the connection
     #[cfg(feature = "sync-tls")]
     pub fn tls_config(mut self, config: impl Into<std::sync::Arc<rustls::ClientConfig>>) -> Self {
@@ -686,6 +705,7 @@ impl<'a> ClientAssociationOptions<'a> {
             kerberos_service_ticket,
             saml_assertion,
             jwt,
+            extended_negotiation,
             ..
         } = self;
         // fail if no presentation contexts were provided: they represent intent,
@@ -729,6 +749,12 @@ impl<'a> ClientAssociationOptions<'a> {
             UserVariableItem::ImplementationClassUID(IMPLEMENTATION_CLASS_UID.to_string()),
             UserVariableItem::ImplementationVersionName(IMPLEMENTATION_VERSION_NAME.to_string()),
         ];
+        for sub_item in extended_negotiation {
+            user_variables.push(UserVariableItem::SopClassExtendedNegotiationSubItem(
+                sub_item.0.to_string(),
+                sub_item.1.clone(),
+            ));
+        }
 
         if let Some(user_identity) = Self::determine_user_identity(
             username.as_deref(),
@@ -1031,6 +1057,14 @@ where
 
     fn user_variables(&self) -> &[UserVariableItem] {
         &self.user_variables
+    }
+
+    /// Given a SOP Class UID, obtain the bytes matching
+    /// that UID with the result from the negotiation, or
+    /// None if that SOP Class UID was rejected or not
+    /// proposed.
+    fn extended_negotiation_for(&self, sop_class_uid: &str) -> Option<&[u8]> {
+        extended_negotiation_filter_impl(&self.user_variables, sop_class_uid)
     }
 }
 
@@ -1550,6 +1584,14 @@ impl<S> Association for AsyncClientAssociation<S> {
 
     fn user_variables(&self) -> &[UserVariableItem] {
         &self.user_variables
+    }
+
+    /// Given a SOP Class UID, obtain the bytes matching
+    /// that UID with the result from the negotiation, or
+    /// None if that SOP Class UID was rejected or not
+    /// proposed.
+    fn extended_negotiation_for(&self, sop_class_uid: &str) -> Option<&[u8]> {
+        extended_negotiation_filter_impl(&self.user_variables, sop_class_uid)
     }
 }
 
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -499,19 +499,20 @@ impl<'a> ClientAssociationOptions<'a> {
         self
     }
 
-    /// Add an Extended Negotiation SOP Class UID and raw data.
-    /// The data must be a byte slice with the information
+    /// Add an Extended Negotiation SOP Class UID and raw data. SCAI
+    /// stands for Service-class-application-information as defined in
+    /// PS3.7. The data must be a byte slice or vector with the information
     /// required by the standard for the given SOP Class. This is
     /// kind of a low-level function, intended to support all
     /// possible current and future use cases for this feature,
     /// so it's the caller's responsibility to fill in this field
     /// correctly.
-    pub fn with_extended_negotiation<T>(mut self, sop_class: T, scai: &[u8]) -> Self
+    pub fn with_extended_negotiation<T>(mut self, sop_class: T, scai: impl Into<Vec<u8>>) -> Self
     where
         T: Into<Cow<'a, str>>,
     {
         self.extended_negotiation
-            .push((sop_class.into(), scai.to_vec()));
+            .push((sop_class.into(), scai.into()));
         self
     }
 
--- a/ul/src/association/mod.rs
+++ b/ul/src/association/mod.rs
@@ -234,6 +234,19 @@ impl CloseSocket for rustls::StreamOwned<rustls::ServerConnection, std::net::Tcp
     }
 }
 
+// Local function to not repeat the same code in every implementation
+pub(crate) fn extended_negotiation_filter_impl<'a>(
+    user_vars: &'a [UserVariableItem],
+    sop_class_uid: &str,
+) -> Option<&'a [u8]> {
+    user_vars.iter().find_map(|uv| match uv {
+        UserVariableItem::SopClassExtendedNegotiationSubItem(uid, data) if uid == sop_class_uid => {
+            Some(data.as_slice())
+        }
+        _ => None,
+    })
+}
+
 /// Trait that represents common properties of an association
 pub trait Association {
     /// Obtain the remote DICOM node's application entity title.
@@ -273,6 +286,12 @@ pub trait Association {
     /// It usually contains the maximum PDU length,
     /// the implementation class UID, and the implementation version name.
     fn user_variables(&self) -> &[UserVariableItem];
+
+    /// Retrieve the bytes associated to a specific SOP Class
+    /// that was negotiated with extended negotiation.
+    ///
+    /// Returns None if that SOP class was rejected or not requested.
+    fn extended_negotiation_for(&self, sop_class_uid: &str) -> Option<&[u8]>;
 }
 
 mod private {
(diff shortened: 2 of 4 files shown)


Enet4:
Thank you for your message @stefan-hubert-bl.

> Revisiting this while monitoring status for the other issue #731, which depends on this. Do I understand correctly, that you're waiting to proceed here because this being a breaking change? ... and you don't want to merge this until you're preparing the next version 0.10 which allows breaking changes? What merge policy are you following: do you have a dedicated separate branch for breaking changes? Or do you wait with merging those until you think time has come for breaking changes on master? Considering this, when do you think is a good time to merge this? (I observed, you did one or two releases (which allowed breaking changes) per year in the past... Thank you!

The current merging policy would be more of the latter: when there are enough non-breaking contributions to justify a patch release, I merge those and release. There is no fixed timeline, since it has mostly depended on my own availability to work the pull requests.

In any case, what I wish to do next is wrap up 0.9.1 by merging #752 and initiating the release process, creating and updating all crates affected. Then the next version will definitely be 0.10.0 and should likely include this PR, so I will do further testing and reviewing at that time to make it happen.

pgimeno4d:
Rebased, squashed, and some very minor wording changes applied.

Enet4:
Oops, server.rs was formatted in #755, so this PR will need a rebase again. Please let me know whether you able to do this.

pgimeno4d:
Sure, no problem. I hope there's no hurry because it might take a few weeks.

Enet4:
I was thinking about getting back on this topic this weekend, but I can also try to resolve the conflicts myself.

pgimeno4d:
Rebased. `cargo clippy` passes, but is giving me a warning about something duplicate:

```
warning: the MSRV in `clippy.toml` and `Cargo.toml` differ; using `1.72.0` from `clippy.toml`
```

pgimeno4d:
I've been wondering. As it is now, the trait's `&self` is not mutable, so the user can't use a mutable reference to modify their own object with the result of the negotiation. Doing it just mutable creates a borrow violation. However, I believe it would be possible to do it by having an Option field with the user's object, where the association object takes ownership during association establishment, and the user can `.take()` it after finishing.

Say, something like this:

```
// The user struct
struct OptionsNegotiation {
    application fields
}

// The trait implemented by this PR, but with mutable self
impl Negotiation for OptionsNegotiation {
    fn extended_negotiation(&mut self, ...) { ... }
    // this is from the yet-to-come role negotiation PR (#731):
    fn negotiate_roles(&mut self, ...) { ... }
}

// Create an instance
let negotiated_options = OptionsNegotiation { ... };

let assoc = ServerAssociationOptions::new()
    ....
    // Give ownership to the association
    .with_negotiation(negotiated_options)
    .establish()?;

// Grab the instance back
let negotiated_options = assoc
    .take_negotiation()  // internally that would return a .take() of the Option field
    .unwrap();  // at this point we know it's Some(our object) and it's safe to unwrap
```

In the current implementation, the user's object is read-only and lost after establishment, so it makes little sense to use anything but an empty object, just to implement the trait for it. This is the same as in the `AccessControl` trait, which I basically copied. As a consequence, the only way of checking the negotiation result is to look into the user variables and re-parse the byte vectors. It would be possible to perform a single pass during negotiation if the `self` reference received by the trait methods were mutable, as described above.

Which one is preferable?

pgimeno4d:
Thinking more about it, the problems with the borrow checker stem from the fact that the establish_xxx methods don't take ownership of the XxxAssociationOptions object. I imagine this was done to allow reusability of the XxxAssociationOptions objects. In fact, the idea explained above has the problem that the OptionsNegotiation object needs to live in both XxxAssociationOptions and in XxxAssociation, unless it's passed as a parameter to the establish_* methods. It's becoming a bit of a mess.

So, I can only think of two solutions:

1. Leave it as is (read-only), forcing the user to pass an empty object and re-parse the negotiated byte arrays.
2. Make establish_xxx take ownership of the XxxAssociationOptions object. To reuse it, using .clone() would be necessary.

Any more ideas? Which one would be best?

ul/src/association/server.rs
```diff
@@ -101,6 +101,101 @@ impl AccessControl for AcceptCalledAeTitle {
...
+    /// # }
+    /// ```
+    fn extended_negotiation(&self, _sop_class_uid: &str, _input: &[u8]) -> Option<Vec<u8>> {
+        None
+    }
+}
+
+/// An Extended Negotiation rule that ignores all SOP Classes
+#[derive(Debug, Default, Copy, Clone, Eq, Hash, PartialEq)]
+pub struct DefaultNegotiation;
```
Enet4, line 209: Can we have a name which describes this rule better? Perhaps `IgnoreExtendedNegotiation`?

Enet4:
> Thinking more about it, the problems with the borrow checker stem from the fact that the establish_xxx methods don't take ownership of the XxxAssociationOptions object. I imagine this was done to allow reusability of the XxxAssociationOptions objects. In fact, the idea explained above has the problem that the OptionsNegotiation object needs to live in both XxxAssociationOptions and in XxxAssociation, unless it's passed as a parameter to the establish_* methods. It's becoming a bit of a mess.
> 
> So, I can only think of two solutions:
> 
> 1. Leave it as is (read-only), forcing the user to pass an empty object and re-parse the negotiated byte arrays.
> 2. Make establish_xxx take ownership of the XxxAssociationOptions object. To reuse it, using .clone() would be necessary.
> 
> Any more ideas? Which one would be best?

The association options types were conceived with the intention to describe the full behavior of the association bootstrap, and to make it possible to replicate the same association request at a later time. In this mindset, negotiation objects in the options would be purely descriptive of the intended behavior, and would not have any state nor require being partially moved to a fresh association.

This too could be replicated here if we treat the result of the extended negotiation as a separate, independent type. I suppose most of this concern is about having a way for negotiation implementers to collect custom information about the negotiation's outcome? That too would be possible through trait associated types, though I might need to take a look at what the plan would be to support the other kinds of negotiation.

```rust
trait NegotiationPolicy {
    type Negotiation;

    fn extended_negotiation(&self, sop_class_uid: &str, input: &[u8]) -> Option<Self::Negotiation> { None }
}
```

ul/src/association/server.rs
```diff
@@ -101,6 +101,101 @@ impl AccessControl for AcceptCalledAeTitle {
...
+    /// (Study Root Query/Retrieve Information Model - FIND),
+    /// the first byte in the output should be 1 if the first
+    /// input byte is 1 (client requests relational queries)
+    /// and the server supports relational queries, and 0
+    /// otherwise,and so on. For reference, see Table C.5-2 in
+    /// PS3.4 of the Standard:
+    /// https://dicom.nema.org/medical/dicom/2025d/output/chtml/part04/sect_C.5.html#table_C.5-2
+    ///
+    /// The default action is to not perform extended negotiation
+    /// (returns None for all SOP Classes, therefore no Extended
```
Enet4:
```suggestion
    /// (returns `None` for all SOP Classes, therefore no Extended
```

ul/src/association/mod.rs
```diff
@@ -273,6 +286,12 @@ pub trait Association {
     /// It usually contains the maximum PDU length,
     /// the implementation class UID, and the implementation version name.
     fn user_variables(&self) -> &[UserVariableItem];
+
+    /// Retrieve the bytes associated to a specific SOP Class
+    /// that was negotiated with extended negotiation.
+    ///
+    /// Returns None if that SOP class was rejected or not requested.
```
Enet4:
```suggestion
    /// Returns `None` if that SOP class was rejected or not requested.
```

ul/src/association/server.rs
```diff
@@ -101,6 +101,101 @@ impl AccessControl for AcceptCalledAeTitle {
     }
 }
 
+/// Interface for negotiation of certain aspects of the association.
+pub trait Negotiation {
+    /// User-provided extended negotiation. The result of
+    /// the negotiation must be returned as either None, if
+    /// extended negotiation is not supported by the
+    /// application for the given SOP Class, or Some(Vec<u8>)
```
Enet4: The advice given from rustdoc is particularly important here, as the characters `<` and `>` may be interpreted in surprising ways if we do not declare them as code.
```suggestion
    /// application for the given SOP Class, or `Some(Vec<u8>)`
```

ul/src/association/server.rs
```diff
@@ -101,6 +101,101 @@ impl AccessControl for AcceptCalledAeTitle {
...
+    /// supported and negotiation is successful. The meaning
+    /// is SOP Class-specific and typically index-specific. For
+    /// example, for the SOP Class 1.2.840.10008.5.1.4.1.2.2.1
+    /// (Study Root Query/Retrieve Information Model - FIND),
+    /// the first byte in the output should be 1 if the first
+    /// input byte is 1 (client requests relational queries)
+    /// and the server supports relational queries, and 0
+    /// otherwise,and so on. For reference, see Table C.5-2 in
+    /// PS3.4 of the Standard:
+    /// https://dicom.nema.org/medical/dicom/2025d/output/chtml/part04/sect_C.5.html#table_C.5-2
```
Enet4: rustdoc suggests wrapping URLs, otherwise they will not be turned into clickable links.

```rust
    /// <https://dicom.nema.org/medical/dicom/2025d/output/chtml/part04/sect_C.5.html#table_C.5-2>
```

But better yet:

```suggestion
    /// otherwise, and so on.
    /// For reference, see
    /// [Table C.5-2 in PS3.4 of the Standard][1].
    ///
    /// [1]: https://dicom.nema.org/medical/dicom/2025d/output/chtml/part04/sect_C.5.html#table_C.5-2
```

We can also refer to revision 2026a of the standard now, though there have been no changes to this section since.

Enet4 (commented):
Just leaving a small code review while the details on the negotiation API are being worked on. Would the suggestion in my comment above be suitable for what you were hoping for, or is there something more to consider?

ul/src/association/server.rs
pgimeno4d: Hm, the comment needs revision, but I believe the name is descriptive enough. The trait is intended to implement more negotiations in future than just Extended Negotiation (in particular Role Negotiation, #731, and possibly more in future); the idea is that this object implements the default behaviour in absence of negotiation for all negotiations supported by the trait.

- By ignoring all SOP classes during extended negotiation, they are set to their default values:
  - [PS3.4 B.3.1](https://dicom.nema.org/medical/dicom/current/output/html/part04.html#sect_B.3.1): *Extended Negotiation is optional. In the event that either the SCU or the SCP does not support Extended Negotiation, the defaults shall apply.*
  - [PS3.4 C.5.1.1](https://dicom.nema.org/medical/dicom/current/output/html/part04.html#sect_C.5.1.1): *This negotiation is optional. If absent, the default conditions shall be: \<list of default values\>*
  - [PS3.4 K.5.1](https://dicom.nema.org/medical/dicom/current/output/html/part04.html#sect_K.5.1): *If the SOP Class Extended Negotiation Sub-Item is not returned by the Association-acceptor then fuzzy semantic matching of person names is not supported and timezone query adjustment is unspecified over the Association (default condition).*
  - ...
- By ignoring all SOP classes during role negotiation, roles are set to their default values:
  - [PS3.7 D.3.3.4](https://dicom.nema.org/medical/dicom/current/output/html/part07.html#sect_D.3.3.4): *If the SCP/SCU Role Selection item is not returned by the Association-acceptor then the role of the Association-requestor shall be SCU and the role of the Association-acceptor shall be SCP.*

Maybe this?
```
/// A Negotiation rule that causes the server to omit the negotiation, thus forcing
/// default values for all kinds of negotiations supported, for all SOP classes.
```

pgimeno4d:
> I suppose most of this concern is about having a way for negotiation implementers to collect custom information about the negotiation's outcome? That too would be possible through trait associated types, though I might need to take a look at what the plan would be to support the other kinds of negotiation.

Yes, that's basically the concern. As is, it needs the application to perform a second pass over the extended negotiation user variables after the association establishment, because the user's callbacks are not able to modify the object they're passed, so that object will basically always be an empty one, and the callbacks can't fill in data in an own structure as they parse and return results.

It's not that bad, really. I'm OK with the current approach; I was just lamenting the need for multiple passes over the same data due to limitations in the design. I think my vote is to leave it as-is, as any changes to that would need to be implemented by extending the establish_xxx functions, at least given the way I'm conceiving them.

pgimeno4d:
I've pushed a commit that improves and extends some comments, including the one mentioned in the review.

ul/src/association/client.rs
```diff
@@ -496,6 +499,22 @@ impl<'a> ClientAssociationOptions<'a> {
...
+    /// kind of a low-level function, intended to support all
+    /// possible current and future use cases for this feature,
+    /// so it's the caller's responsibility to fill in this field
+    /// correctly.
+    pub fn with_extended_negotiation<T>(mut self, sop_class: T, scai: &[u8]) -> Self
+    where
+        T: Into<Cow<'a, str>>,
+    {
+        self.extended_negotiation
+            .push((sop_class.into(), scai.to_vec()));
```
Enet4: This is so we can add a `Vec<u8>` without making a redundant copy.

```suggestion
    pub fn with_extended_negotiation<T>(mut self, sop_class: T, scai: impl Into<Vec<u8>>) -> Self
    where
        T: Into<Cow<'a, str>>,
    {
        self.extended_negotiation
            .push((sop_class.into(), scai.into()));
```

ul/src/association/server.rs
```diff
@@ -101,6 +101,114 @@ impl AccessControl for AcceptCalledAeTitle {
...
+    ///     // 2nd byte position (index 1) is the combined date/time matching option.
+    ///     // If it is zero or absent, consider it disabled.
+    ///     combined_date_time = values.get(2 - 1).copied().unwrap_or(0) != 0;
+    ///     // 6th byte position (index 5) is the empty value matching option
+    ///     empty_value_matching = values.get(6 - 1).copied().unwrap_or(0) != 0;
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    fn extended_negotiation(&self, _sop_class_uid: &str, _input: &[u8]) -> Option<Vec<u8>> {
```
Enet4: These underscore prefixes will appear in documentation. Given the simplicity of the impl block, we can go for a single allow attribute to clear the warnings.

```suggestion
    #[allow(unused_variables)]
    fn extended_negotiation(&self, sop_class_uid: &str, input: &[u8]) -> Option<Vec<u8>> {
```

Enet4 (approved):
This serves as a good low-level construct for extended negotiation, and we will have the opportunity to finetune the new methods in #765. I will just place some inline suggestions which can be either added here before I merge or in #765.

Once again, thank you for your contributions!

pgimeno4d:
I've applied the suggested changes and added a comment about the meaning of SCAI.

I couldn't apply the suggestions using the interface, because the diff that changed the type of the parameter was incomplete (it only removed one line), so I've instead mentioned authorship in the commit message.

Enet4 (approved):

PR #738, opened 2026-02-09
[encoding] Add function to access frame pixeldata

feliwir:
Makes the behavior of `toimage` accessible as a trait fn in PixelDataObject

### DIFF
--- a/encoding/src/adapters.rs
+++ b/encoding/src/adapters.rs
@@ -182,6 +182,70 @@ pub trait PixelDataObject {
     ///
     /// Returns `None` if no pixel data is found.
     fn raw_pixel_data(&self) -> Option<RawPixelData>;
+
+    /// Return the pixel data of a specific frame as a byte slice/vector,
+    ///
+    /// Returns `None` if there is no such frame or there is no pixeldata at all.
+    fn frame_pixel_data(&self, frame: u32) -> Option<Cow<'_, [u8]>> {
+        match self.number_of_fragments() {
+            Some(number_of_fragments) => {
+                if number_of_fragments == self.number_of_frames()? {
+                    self.fragment(frame as usize)
+                } else {
+                    // In this case we look up the basic offset table
+                    // and gather all of the frame's fragments in a single vector.
+                    // Note: not the most efficient way to do this,
+                    // consider optimizing later with byte chunk readers
+                    let offset_table = self.offset_table()?;
+                    let base_offset = offset_table.get(frame as usize).copied();
+                    let base_offset = if frame == 0 {
+                        base_offset.unwrap_or(0) as usize
+                    } else {
+                        base_offset? as usize
+                    };
+                    let next_offset = offset_table.get(frame as usize + 1);
+
+                    let mut offset = 0;
+                    let mut frame_data = Vec::new();
+                    for idx in 0..number_of_fragments as usize {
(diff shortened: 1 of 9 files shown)


encoding/src/adapters.rs
```diff
@@ -182,6 +182,70 @@ pub trait PixelDataObject {
...
+                    Some(Cow::Owned(frame_data))
+                }
+            }
+            // In the case of native pixel data, the whole data is considered as a single fragment.
+            None => {
+                let rows = self.rows()?;
+                let columns = self.cols()?;
+                let samples_per_pixel = self.samples_per_pixel()?;
+                let bits_allocated = self.bits_allocated()?;
+                let frame_size = rows * columns * samples_per_pixel * ((bits_allocated + 7) / 8);
```
Enet4: This reminds me that the native frame size calculation that we've been using might not work for `YBR_FULL_422` (in which case the 3 RGB samples are saved in 2 YBR sample values per pixel instead). I suppose it is a rare case, and adjusting the calculation here without a test would not be very interesting.

encoding/src/adapters.rs
```diff
@@ -182,6 +182,70 @@ pub trait PixelDataObject {
...
     fn raw_pixel_data(&self) -> Option<RawPixelData>;
+
+    /// Return the pixel data of a specific frame as a byte slice/vector,
+    ///
+    /// Returns `None` if there is no such frame or there is no pixeldata at all.
+    fn frame_pixel_data(&self, frame: u32) -> Option<Cow<'_, [u8]>> {
+        match self.number_of_fragments() {
+            Some(number_of_fragments) => {
+                if number_of_fragments == self.number_of_frames()? {
+                    self.fragment(frame as usize)
```
Enet4: An if guard could be used here to cut on the extra indentation level:

```rust
match self.number_of_fragments() {
    Some(number_of_fragments) if number_of_fragments == self.number_of_frames()? => {
        self.fragment(frame as usize)
    }
    Some(number_of_fragments) => {
        // ...
```

Enet4 (changes requested):
Thank you for working on this extension to the PixelDataObject trait, it will really make it easier to work with pixel data.

I have identified a few concerns with the implementation, please let me know if you need assistance.

encoding/src/adapters.rs
```diff
@@ -182,6 +188,93 @@ pub trait PixelDataObject {
...
+    ///
+    /// Returns `None` if there is no such frame or there is no pixel data at all.
+    ///
+    /// _Note:_ If pixel data is uncompressed and Bits Allocated is 1,
+    /// the slice may include leading or trailing bits
+    /// belonging to other frames,
+    /// depending on the size of each frame.
+    fn frame_pixel_data(&self, frame: u32) -> Option<Cow<'_, [u8]>> {
+        match self.number_of_fragments() {
+            Some(number_of_fragments) if number_of_fragments == self.number_of_frames()? => {
```
Enet4: I found the culprit. `number_of_frames` returns the value explicitly provided by the object, which can be missing. Instead of returning `None` with the `?` operator, we should instead assume the object not to be multi-frame, and therefore have 1 frame for practical purposes.

Enet4 (approved):
I think it's in good condition now. Thank you for the initiative and continued assistance! :+1:

PR #753, opened 2026-03-24
[ul] Run `cargo fmt` on some files

pgimeno4d:
This commit makes no other changes whatsoever, and can be reproduced by running `rustfmt --edition 2018` on the files present.

Since I had to rebase #734 anyway, I thought it would be convenient to start with a cleanly formatted base. Some of the files that are formatted will be used in the upcoming Role Selection Negotiation PR.

This may affect some other PRs, notably #707.

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -12,25 +12,23 @@ use std::{
     time::Duration,
 };
 
-use crate::{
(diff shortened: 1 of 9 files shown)


Enet4 (approved):
Sounds good! I was only holding it back a bit because I knew that it would conflict with existing pull requests. Thanks!

PR #755, opened 2026-04-05
ENH: Expose called_ae_title method on server association structs

naterichman:
Sometimes it is useful to know which AE title a calling entity used, such as if different behavior or access to different images are gated by AE title.

Called AE title is now stored on the server association structs, and methods are implemented to retrieve it

Context: I am writing a storescp server implementation which will route dicoms to various places based on which AE title has been called.  I use `accept_any` and then want to check which AE was called in order to route

### DIFF
--- a/ul/src/association/server.rs
+++ b/ul/src/association/server.rs
@@ -11,27 +11,27 @@ use std::{io::Write, net::TcpStream};
 
 use crate::association::private::SyncAssociationSealed;
 use crate::association::{
-    encode_pdu, read_pdu_from_wire, AbortedSnafu, Association, CloseSocket,
-    MissingAbstractSyntaxSnafu, RejectedSnafu, SendPduSnafu, SocketOptions, SyncAssociation,
-    UnexpectedPduSnafu, UnknownPduSnafu, WireSendSnafu,
+    AbortedSnafu, Association, CloseSocket, MissingAbstractSyntaxSnafu, RejectedSnafu,
+    SendPduSnafu, SocketOptions, SyncAssociation, UnexpectedPduSnafu, UnknownPduSnafu,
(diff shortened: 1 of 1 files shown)


pgimeno4d:
Looks good to me, functionality-wise, but could you please run `rustfmt --edition 2018` on the file, or at least add commas at the end of the places that add the field `called_ae_title`, so that if another field is added in future, it will be a one line change as well? (note: I am not a maintainer, just giving suggestions)

ul/src/association/server.rs
```diff
@@ -830,6 +835,8 @@ pub struct ServerAssociation<S> {
     socket: S,
     /// The application entity title of the other DICOM node
     client_ae_title: String,
+    /// The AE title the calling client used.
```
Enet4, line 838: This may be clearer and more aligned with the doc comment to `client_ae_title`, in which "application entity title" is also expanded. Please consider updating in other equivalent places.

```suggestion
    /// The application entity title used by the client
    /// to call this server
```

ul/src/association/server.rs
```diff
@@ -868,6 +875,11 @@ impl<S> ServerAssociation<S> {
     pub fn client_ae_title(&self) -> &str {
         &self.client_ae_title
     }
+
+    /// Obtain the AE title the client called.
```
Enet4, line 879:
```suggestion
    /// Obtain the application entity title called by the client.
```

Enet4 (commented):
Yup, this is useful indeed. We can clean it up a bit and proceed with merging.

Enet4 (approved):
I went ahead and adjusted the doc-comments, and also formatted the entire server.rs file. We can proceed with this! :+1:

PR #756, opened 2026-04-09
Add adaptive VR decoder for explicit/implicit mismatch

### LINKED ISSUE #521: Failure when Explicit VR contains Implicit VR
So this is another one with an invalid dicom, which is a case I think should be handled.

This is one where a dicom comes in with ExplicitVRLittleEndian TS and some elements don't actually have VRs.  The way the explicit VR decoder is doing that is by setting the VR to `UN` [here](https://github.com/Enet4/dicom-rs/blob/master/encoding/src/decode/explicit_le.rs#L45) which then causes it to read the two extra reserved bytes which may or may not be there depending on what the actual VR is supposed to be.

I've reproduced this by taking a basic pydicom test file, adding a private tag and manually deleting the VR bytes, so the section of the file looks like this:
```
00000316: 11 00 10 00 4c 4f 04 00                (0011, 0010) LO Length: 4
0000031e: 54 65 73 74                            b'Test'
00000322: 11 00 01 10 02 00 00 00                (0011, 1001) None Length: 2
0000032a: 31 32                                  b'12'
```

 so this section of the file looks like this, which gives an error when I `dicom-dump`

```
➜  ~ $ dicom-dump ~/Downloads/mr_broken.dcm
/home/naterichman/Downloads/mr_broken.dcm:
Could not read data set token

Caused by these errors (recent errors listed first):
  1: Could not read 1585713 value bytes for element tagged (0011,1001)
  2: Could not read value from source at position 480
  3: failed to fill whole buffer
```

Part of the reason I think this should be supported is because pydicom supports it :wink:  The way its done in pydicom is [here](https://github.com/pydicom/pydicom/blob/0beb18676c4df71a308ef41b599812ce71df3cc7/src/pydicom/filereader.py#L150) where they first try to read explicit and then if the resulting VR is not alphabetical, they fall back to implicit.  

I'm not sure how that would work with the design here and I fully understand that this library cannot be as dynamic as python, but I think at least the error should be clearer, i.e. if the VR is not present when reading when reading an explicit DS make that the warning instead of trying to parse the length from the next bytes.

Link to documents, I uploaded both the original and the one I edited to be broken: [onedrive](https://flywheelio-my.sharepoint.com/:f:/g/personal/naterichman_flywheel_io/EpgD64aVK75DpZhePnKU8GUBhdgv7cG50RHOgyavfA27YQ?e=HSTthU)

nathannncurtis:
Closes #521

  Adds `AdaptiveVRLittleEndianDecoder`, which handles files that declare Explicit VR LE in the transfer syntax but actually encode the dataset as Implicit VR.

  ## How it works

  On the first non-delimiter dataset element, the decoder reads the two bytes after the tag and runs `VR::from_binary`. If they're a recognized VR code, it locks to explicit; otherwise it treats them as the first two bytes of a 4-byte
   implicit length and locks to implicit.

  To prevent false positives (where implicit length bytes happen to match a VR code), the probed VR is cross-checked against the data dictionary. If the dictionary knows the tag and disagrees, the decoder falls back to implicit. The
  check handles `VirtualVr` variants (`Xs`, `Ox`, `Px`, `Lt`) that allow more than one concrete VR.

  ## What changed

  - `encoding/src/decode/adaptive_le.rs` — the decoder, implements `Decode`/`DecodeFrom`, with tests
  - `encoding/src/decode/mod.rs` — module registration
  - `parser/src/dataset/read.rs` — `flexible_decoding` field on `DataSetReaderOptions` (default `false`). When true and the transfer syntax is LE, the reader uses the adaptive decoder instead.

  Default behavior is unchanged. All existing tests pass.

### DIFF
--- a/encoding/src/decode/adaptive_le.rs
+++ b/encoding/src/decode/adaptive_le.rs
@@ -1,0 +1,552 @@
+//! Adaptive VR Little Endian syntax transfer implementation.
+//!
+//! This decoder handles non-conformant DICOM files that declare
+//! Explicit VR Little Endian in their transfer syntax but actually
+//! encode the dataset in Implicit VR. On the first non-meta,
+//! non-delimiter element, it probes the bytes after the tag to
+//! determine whether they form a valid VR code. If they do,
+//! it locks to explicit VR for the rest of the file; if not,
+//! it switches to implicit VR.
+
+use crate::decode::basic::LittleEndianBasicDecoder;
+use crate::decode::{
+    BadSequenceHeaderSnafu, BasicDecode, Decode, DecodeFrom, ReadHeaderTagSnafu,
+    ReadItemHeaderSnafu, ReadItemLengthSnafu, ReadLengthSnafu, ReadReservedSnafu, ReadTagSnafu,
+    ReadVrSnafu, Result,
+};
+use byteordered::byteorder::{ByteOrder, LittleEndian};
+use dicom_core::dictionary::{DataDictionary, DataDictionaryEntry, VirtualVr};
+use dicom_core::header::{DataElementHeader, Length, SequenceItemHeader};
+use dicom_core::{Tag, VR};
+use dicom_dictionary_std::StandardDataDictionary;
+use snafu::ResultExt;
+use std::cell::Cell;
+use std::fmt;
+use std::io::Read;
+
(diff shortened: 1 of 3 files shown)


Enet4 (approved):
Looks OK. The API conventions from `ImplicitVRLittleEndianDecoder` were followed and it has reasonable test coverage. Thanks! :+1:

PR #759, opened 2026-04-28
Resolve Collector API Missing Offset Table

beatrice-b-m:
This PR resolves issue #745 where DICOMs with encapsulated pixel data have an empty BOT when loaded using the Collector API. This was occurring because the BOT was being decoded, but never registered to the `offset_table` var where it was expected (causing an empty table to be initialized instead).

https://github.com/Enet4/dicom-rs/blob/d5d44c6b6fadf435c3ffc67010eee2e1073fb5de/object/src/collector.rs#L1201-L1221

Also added the test included in #745 to replicate the error and fix. All relevant new/existing tests pass after the fix.


object/src/collector.rs
```diff
@@ -1786,9 +1787,77 @@ mod tests {
...
+        // inspect the basic offset table
+        let pixel_data = dset
+            .get(tags::PIXEL_DATA)
+            .expect("Should have pixel data")
+            .value();
+        let DicomValue::PixelSequence(seq) = pixel_data else {
+            panic!("Expected encapsulated pixel data");
+        };
+        let bot = seq.offset_table();
+        assert_eq!(bot.len(), 2); // <-- FAILS HERE
```
Enet4: It _would_ have! Not anymore! :)

```suggestion
        assert_eq!(bot.len(), 2);
```

Enet4 (commented):
Great!

Enet4 (approved):
Thanks again for patching the collector API!

PR #763, opened 2026-05-06
Add `read_to` parameter, that allows to read certain dicom file, without excluding chosen tag

qarmin:
`read_until` is a parameter that reads a file until it finds a specific tag and excludes it from the results. The problem is that, to speed up reading data from disk, I often want to read only part of file. 
In that case, I currently need to manually determine a greater tag to search for in order to include the desired one, e.g. to read MANUFACTURER_MODEL_NAME, I need to read until I find SERIES_INSTANCE_UID(or other greater tag), which looks quite strange
```
    let obj = OpenFileOptions::new()
        .read_until(tags::SERIES_INSTANCE_UID) 
        .open_file(path)
        .ok()?
        .into_inner();
```

`read_to` reads data up to and including the specified tag, without discarding it.

### DIFF
--- a/object/src/collector.rs
+++ b/object/src/collector.rs
@@ -794,7 +794,7 @@ where
             self.source.parser()
         };
 
-        Self::collect_to_object(&mut self.state, parser, false, None, to, &self.dictionary)
+        Self::collect_to_object(&mut self.state, parser, false, None, None, to, &self.dictionary)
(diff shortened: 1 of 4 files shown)


Enet4 (approved):
This is useful and appears to be well written. Thanks! :+1:

PR #765, opened 2026-05-11
[ul] Implement role negotiation

### LINKED ISSUE #731: Allow role selection for association requests

pgimeno4d:
Implements role negotiation for association objects, allowing applications to implement a C-GET SCU that follows the standard (currently the library does not allow it). Fixes #731.

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -788,7 +788,10 @@ impl<'a> ClientAssociationOptions<'a> {
         for sub_item in scu_scp_role_selection {
             user_variables.push(UserVariableItem::ScuScpRoleSelectionSubItem(
                 sub_item.0.to_string(),
-                RequestorRoles { scu: sub_item.1, scp: sub_item.2 },
+                RequestorRoles {
+                    scu: sub_item.1,
+                    scp: sub_item.2,
+                },
             ));
         }
 
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -23,8 +23,8 @@ use crate::{
     pdu::{
         AbortRQSource, AssociationAC, AssociationRQ, DEFAULT_MAX_PDU, LARGE_PDU_SIZE,
         PDU_HEADER_SIZE, Pdu, PresentationContextNegotiated, PresentationContextProposed,
-        PresentationContextResultReason, UserIdentity, UserIdentityType, UserVariableItem,
-        write_pdu,
+        PresentationContextResultReason, RequestorRoles, UserIdentity, UserIdentityType,
+        UserVariableItem, write_pdu,
     },
 };
 use snafu::{ResultExt, ensure};
@@ -788,8 +788,7 @@ impl<'a> ClientAssociationOptions<'a> {
         for sub_item in scu_scp_role_selection {
             user_variables.push(UserVariableItem::ScuScpRoleSelectionSubItem(
                 sub_item.0.to_string(),
-                sub_item.1,
-                sub_item.2,
+                RequestorRoles { scu: sub_item.1, scp: sub_item.2 },
             ));
         }
 
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -265,6 +265,8 @@ pub struct ClientAssociationOptions<'a> {
     jwt: Option<Cow<'a, str>>,
     /// Extended Negotiation info
     extended_negotiation: Vec<(Cow<'a, str>, Vec<u8>)>,
+    /// SCU/SCP Role Selection info
+    scu_scp_role_selection: Vec<(Cow<'a, str>, bool, bool)>,
     /// Socket options for TCP connections
     socket_options: SocketOptions,
     /// TLS configuration to use for the connection
@@ -295,6 +297,7 @@ impl Default for ClientAssociationOptions<'_> {
             saml_assertion: None,
             jwt: None,
             extended_negotiation: Vec::new(),
+            scu_scp_role_selection: Vec::new(),
             socket_options: SocketOptions {
                 read_timeout: None,
                 write_timeout: None,
@@ -516,6 +519,15 @@ impl<'a> ClientAssociationOptions<'a> {
         self
     }
 
+    pub fn with_role_selection<T>(mut self, sop_class: T, scu_role: bool, scp_role: bool) -> Self
+    where
+        T: Into<Cow<'a, str>>,
+    {
+        self.scu_scp_role_selection
+            .push((sop_class.into(), scu_role, scp_role));
+        self
+    }
+
     /// Set the TLS configuration to use for the connection
     #[cfg(feature = "sync-tls")]
     pub fn tls_config(mut self, config: impl Into<std::sync::Arc<rustls::ClientConfig>>) -> Self {
@@ -707,6 +719,7 @@ impl<'a> ClientAssociationOptions<'a> {
             saml_assertion,
             jwt,
             extended_negotiation,
+            scu_scp_role_selection,
             ..
         } = self;
         // fail if no presentation contexts were provided: they represent intent,
@@ -754,6 +767,13 @@ impl<'a> ClientAssociationOptions<'a> {
             user_variables.push(UserVariableItem::SopClassExtendedNegotiationSubItem(
                 sub_item.0.to_string(),
                 sub_item.1.clone(),
+            ));
+        }
+        for sub_item in scu_scp_role_selection {
+            user_variables.push(UserVariableItem::ScuScpRoleSelectionSubItem(
+                sub_item.0.to_string(),
+                sub_item.1,
+                sub_item.2,
             ));
         }
 
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -15,20 +15,21 @@ use std::{
 #[cfg(feature = "async")]
 use crate::association::AsyncAssociation;
 use crate::{
+    AeAddr, IMPLEMENTATION_CLASS_UID, IMPLEMENTATION_VERSION_NAME,
     association::{
-        encode_pdu, extended_negotiation_filter_impl, private::SyncAssociationSealed,
-        read_pdu_from_wire, Association, NegotiatedOptions, SocketOptions, SyncAssociation,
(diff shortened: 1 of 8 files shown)


ul/src/association/mod.rs
```diff
@@ -292,6 +292,28 @@ pub trait Association {
     ///
     /// Returns `None` if that SOP class was rejected or not requested.
     fn extended_negotiation_for(&self, sop_class_uid: &str) -> Option<&[u8]>;
+
+    /// Roles that the Association-requestor may assume. Returns a pair
+    /// of booleans; the first is whether it can assume an SCU role and
+    /// the second whether it can assume an SCP role.
+    fn requestor_roles_for(&self, sop_class_uid: &str) -> (bool, bool) {
```
Enet4: Guideline [C-CUSTOM-TYPE](https://rust-lang.github.io/api-guidelines/type-safety.html#c-custom-type) would tell us here to make a custom type for the output of this function, so that there is less room for ambiguity on what the two booleans mean.

There would be more than one way to do this, but it seems to me that we can turn the pair into a single type that can hold the equivalent information. This has two benefits: we can forbid the case of assuming neither role SCU nor SCP; and the same type can also be used in the association options, in method `with_role_selection`.

```rust
/// The roles that an application entity can serve,
/// provided as part of role negotiation.
#[derive(Debug, Copy, Clone, Eq, Hash, PartialEq)]
enum ScuScpRole {
    /// Can assume the role of SCU
    Scu = 1,
    /// Can assume the role of SCP
    Scp = 2,
    /// Can assume the role of both SCU and SCP
    ScuAndScp = 3,
}
```

ul/src/association/server.rs
```diff
@@ -101,6 +101,171 @@ impl AccessControl for AcceptCalledAeTitle {
...
+    /// # }
+    /// ```
+    fn negotiate_roles(
+        &self,
+        _sop_class_uid: &str,
+        _scu_role: bool,
+        _scp_role: bool,
+    ) -> Option<(bool, bool)> {
+        None
+    }
```
Enet4, line 274:
```suggestion
    #[allow(unused_variables)]
    fn negotiate_roles(
        &self,
        sop_class_uid: &str,
        scu_role: bool,
        scp_role: bool,
    ) -> Option<(bool, bool)> {
        None
    }
```

sjswerdloff:
This is a lot more than just the role negotiation, it's extended negotiation, which is a good thing (the example of "...date/time matching and empty value matching ..." in server.rs makes it clear that this is extended negotiation).
But it certainly addresses #731

pgimeno4d:
@sjswerdloff That's because it includes the commits in #734 since they are a prerequisite. Only the last commit counts here, and it deals with role negotiation; the rest are part of #734. Should #734 be rejected, this would need to be redesigned.

ul/src/association/client.rs
```diff
@@ -515,6 +518,15 @@ impl<'a> ClientAssociationOptions<'a> {
         self
     }
 
+    pub fn with_role_selection<T>(mut self, sop_class: T, scu_role: bool, scp_role: bool) -> Self
```
Enet4, line 538:
```suggestion
    /// Add an item for role negotiation.
    pub fn with_role_selection<T>(mut self, sop_class: T, scu_role: bool, scp_role: bool) -> Self
```

pgimeno4d:
Now that #734 is merged, I've rebased it on top of current master and it's ready for review.

Enet4 (commented):
A few notes on the API.

ul/src/association/mod.rs
pgimeno4d: I figured this would be an issue. I was discontent with it as well.

The combination SCU=false, SCP=false is not invalid; if the client offers, say, SCU=true and SCP=false for a particular SOP class, the server must be able to reject the SCU role, and it should never set any to true. [PS3.7 D.3.3.4](https://dicom.nema.org/medical/dicom/2025d/output/html/part07.html#sect_D.3.3.4):
> The Association-acceptor, for each SCP/SCU Role Selection item offered, either accepts the Association-requestor proposal by returning the same value (1) or turns down the proposal by returning the value (0). The Association-acceptor shall not return the value (1) if the Association-requestor has not proposed the role, i.e., it has sent a value (0).

In that example, returning false/false is the only valid option.

In any case, I'm not very happy with the enum even if it included the Neither case, because it makes a bit harder to separate whether the SCU or the SCP role are active than just using two bools; for example, something like this would be necessary:
```
let can_be_scu = scu_scp_role == Scu || scu_scp_role == ScuAndScp;
let can_be_scp = scu_scp_role == Scp || scu_scp_role == ScuAndScp;
```
I don't think making it copyable would yield any tangible benefits here. So, I'd rather prefer a struct. Here are my proposals:

```
struct RequestorRoles {
    can_be_scu: bool,
    can_be_scp: bool,
}
```
That conveys the exact meaning of the options. Another alternative that shortens the field names but makes the type name a bit more cumbersome:
```
struct RolesRequestorCanTake {
  scu: bool,
  scp: bool,
}
```
Or just use the docs and keep the names simple (not final docs, that's just for emphasizing that they mention how the fields behave):
```
/// Roles that the association requestor can adopt for a particular SOP class.
struct RequestorRoles {
    scu: bool,
    scp: bool,
}
```
I like the latter the most, for its brevity. I'm open to other proposals as well, like one using `bitflags`.

ul/src/association/client.rs
pgimeno4d: Totally forgot to document this one, thanks!

pgimeno4d: My understanding is that `#[allow(unused_variables)]` is not necessary here because they *are* used. Did I miss something?

Enet4: Oh yes, right! Updated the suggestion accordingly.

ul/src/association/mod.rs
Enet4: > Or just use the docs and keep the names simple (not final docs, that's just for emphasizing that they mention how the fields behave):
> 
> ```
> /// Roles that the association requestor can adopt for a particular SOP class.
> struct RequestorRoles {
>     scu: bool,
>     scp: bool,
> }
> ```

All right, that will do nicely. We can also derive some traits (`Debug`, `Copy`, `Clone`, `Eq`, `Hash`, `PartialEq`) and offer associated constants.

> I'm open to other proposals as well, like one using `bitflags`.

I would not be bothered to use bitflags for a struct with only two bit flags. A manual bitwise-or implementation could also be done with ease.

pgimeno4d:
Thanks for the review. Hopefully addressed all comments. `rustfmt` insisted in reformatting the headers, so this time I've decided to not fight it, which results in a bit of header shuffling noise. Sorry about that. I'm not sure if `Copy` is really suitable here but I've added it per your request.

Enet4:
> I'm not sure if `Copy` is really suitable here but I've added it per your request.

I could be missing something, but it is not anticipated for this struct to ever have non-Copy fields, right? If the struct ever changes, it would be through the addition of more booleans or other integral types, which are also `Copy`. This would be the key criterion, as documented in the [trait's documentation](https://doc.rust-lang.org/stable/std/marker/trait.Copy.html#when-should-my-type-be-copy).

ul/src/pdu/mod.rs
```diff
@@ -489,6 +489,7 @@ pub enum UserVariableItem {
     ImplementationClassUID(String),
     ImplementationVersionName(String),
     SopClassExtendedNegotiationSubItem(String, Vec<u8>),
+    ScuScpRoleSelectionSubItem(String, bool, bool),
```
Enet4: Could we use `RequestorRoles` here instead of the two bools (changing the other parts accordingly)?

```suggestion
    ScuScpRoleSelectionSubItem(String, RequestorRoles),
```

Enet4 (commented):
Just one last concern involving the public API and I'd say we're good to merge.

pgimeno4d:
> I could be missing something, but it is not anticipated for this struct to ever have non-Copy fields, right?

It's just that I don't like to encourage inadvertent copying of multi-field structures that may become large, much preferring explicit .clone() instead, so one realizes that it's not a trivial zero-cost operation like copying a number, or a tuple with a number, or a simple enum.

But sure, no problem in adhering to that directive.

Enet4:
> > I could be missing something, but it is not anticipated for this struct to ever have non-Copy fields, right?
> 
> It's just that I don't like to encourage inadvertent copying of multi-field structures that may become large, much preferring explicit .clone() instead, so one realizes that it's not a trivial zero-cost operation like copying a number, or a tuple with a number, or a simple enum.

Fair, but since all fields are public at the moment, adding new fields will be a breaking change anyway, so if it happens that the copy is longer zero-cost in the future, then we can also adjust trait derivations accordingly.

The more forward-compatible alternative is to make these fields private and offer different helpers to construct and inspect the requested roles, though I'm not sure if it's worth it. :)

Enet4 (approved):
Thank you once again for this important contribution!

stefan-hubert-bl:
Thank you so much @Enet4 and @pgimeno4d for making this possible. Appreciate your work and attention to this.

PR #766, opened 2026-05-13
test(ul): tolerate peer-first socket close on association release

sjswerdloff:
## Summary

The `successive_pdus_during_server_association` tests unwrap the result of `association.release()`. On macOS, the peer commonly closes its half of the socket before our release request lands, and the kernel surfaces this as ENOTCONN (errno 57). The release semantics have already completed at that point, so the test panic is a false positive caused by platform-specific socket teardown ordering.

Discards the release result in those two tests (sync and async), matching the standard pattern for graceful peer-close handling. Linux CI is unaffected because the kernel does not surface the race the same way — which is why this bug has been invisible to upstream until now.

## Reproduction

On macOS, on clean master:

```
cargo test -p dicom-ul --lib --all-features successive_pdus
```

Result:
```
test ...test_server_association_receives_extra_pdu_589_impl ... FAILED
test ...test_server_association_receives_extra_pdu_589_impl_async ... ok
test result: FAILED. 11 passed; 1 failed
```

Panic:
```
called `Result::unwrap()` on an `Err` value: Close {
  source: Os { code: 57, kind: NotConnected,
              message: "Socket is not connected" }, ... }
```

After this PR, all 12 tests pass on macOS.

## Why fix both the sync and async sites?

Only the sync site is currently failing on macOS — the async one happens to win the race more reliably. Both have the same latent bug and the same tolerance pattern; treating them symmetrically prevents future macOS flakiness on the async variant when timing shifts.

## Scope

- Only test code is touched.
- Only the two `release()` call sites in `successive_pdus_during_server_association` mod.
- No production code is changed.

## Test plan

- [x] cargo test -p dicom-ul --lib --all-features successive_pdus (12 passed; previously 1 failed on macOS)
- [x] Verified the diff is confined to two release sites in `ul/src/association/tests.rs`

### DIFF
--- a/ul/src/association/tests.rs
+++ b/ul/src/association/tests.rs
@@ -599,8 +599,10 @@ mod successive_pdus_during_server_association {
             .establish_with_extra_pdus(AeAddr::new_socket_addr(server_addr), vec![client_pdu])
             .unwrap();
 
-        // Clean shutdown
-        association.release().unwrap();
+        // Clean shutdown. The peer may close its half of the socket before
+        // our release request lands, which surfaces as ENOTCONN on macOS;
+        // the release semantics have completed in either case.
+        let _ = association.release();
         server_handle.join().unwrap();
     }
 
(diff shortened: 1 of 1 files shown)


Enet4 (approved):
Sounds about right. We did think about the need to ignore the result entirely in [past reiterations](https://github.com/Enet4/dicom-rs/pull/728#discussion_r2830287301), but we could not get to the reason why this had to be done in the first place. Let's keep those tests like that.

Thank you!

PR #767, opened 2026-05-13
fix(ul): echo SCP/SCU Role Selection items in Association Accept

sjswerdloff:
## Summary

Strict DICOM SCUs (notably pynetdicom >= 2.0) require the Association Acceptor to confirm requested SCP/SCU roles by echoing Role Selection sub-items (item type 0x54) in the Association Accept (AC) PDU. Without this echo, C-GET operations fail when the server is acting as SCP for negotiated storage classes.

This change filters `UserVariableItem::Unknown(0x54, _)` entries from the requestor's user variables in `process_a_association_rq` and appends them to the AC's user variables. Both synchronous and asynchronous establish paths flow through this shared function, so a single change covers sync, async, plain TCP, and TLS acceptors.

Refs upstream issue: https://github.com/Enet4/dicom-rs/issues/731

## Changes

- `ul/src/association/server.rs`: collect role-selection items from the request and extend them onto the AC response.
- `ul/src/association/server.rs` (tests): add `test_ac_echoes_role_selection_items` covering the echo behaviour and asserting unrelated unknown sub-items are NOT propagated.

## Compatibility

Strictly additive, non-API-breaking. SCUs that do not send role-selection items observe the previous AC structure unchanged.

## Credit

Original patch authored by Christof Schadt for the rust-dicom-broker project. Ported unchanged in substance; rebased onto current upstream master where the sync/async AC sites have converged into a single shared method.

## Test plan

- [x] cargo build -p dicom-ul
- [x] cargo test -p dicom-ul --lib --all-features (new test passes)
- [x] cargo clippy -p dicom-ul --no-deps --all-features -- -D warnings
- [x] cargo fmt --check clean on modified file

### DIFF
--- a/ul/src/association/server.rs
+++ b/ul/src/association/server.rs
@@ -600,6 +600,25 @@ where
                     })
                     .collect();
 
+                // Collect SCP/SCU Role Selection sub-items (type 0x54) from the
+                // requestor's user variables so we can echo them back in the AC,
+                // confirming the requested roles (required by strict SCUs such as
+                // pynetdicom >= 2.0 for C-GET where we act as SCP for storage classes).
+                let role_selection_items: Vec<UserVariableItem> = user_variables
+                    .iter()
+                    .filter(|item| matches!(item, UserVariableItem::Unknown(0x54, _)))
+                    .cloned()
+                    .collect();
+
+                let mut ac_user_variables = vec![
+                    UserVariableItem::MaxLength(self.max_pdu_length),
+                    UserVariableItem::ImplementationClassUID(IMPLEMENTATION_CLASS_UID.to_string()),
+                    UserVariableItem::ImplementationVersionName(
+                        IMPLEMENTATION_VERSION_NAME.to_string(),
+                    ),
+                ];
+                ac_user_variables.extend(role_selection_items);
+
                 let pdu = Pdu::AssociationAC(AssociationAC {
                     protocol_version: self.protocol_version,
                     application_context_name,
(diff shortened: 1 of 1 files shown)


sjswerdloff:
I see that there has been discussion of the issue, but no PR for it yet.
If someone else's work is more complete and correct, by all means, let's close this one.
This is blocking for some work being done to help move forward on proxies (DIMSE <==> DICOMWeb) with the intent of making use of the easier/better understood security approach in DICOMWeb (https), and being able to drop proxies in to legacy environments.  See RT Systems Engineering Council (sponsored by AAPM).

Enet4:
Thank you for contributing! Yes, I was going to comment on the fact that #765 seems to provide the equivalent functionality, though there wasn't a reference to #731 on there yet. I was in the middle of reviewing the other one, so this could be an opportunity to clarify what has been done and perhaps set some points for coordinated efforts.

Enet4:
#765 has just been merged, which should serve this use case well. If there is still an open concern, feel free to raise a new issue.

pgimeno4d:
@sjswerdloff For avoidance of doubt, to implement the equivalent functionality to this patch (accept all proposed roles), you would do something like this:

```rust
struct AcceptAllRoles;

impl Negotiation for AcceptAllRoles {
    fn negotiate_roles(
        &self,
        sop_class_uid: &str,
        scu: bool,
        scp: bool,
    ) -> Option<RequestorRoles> {
        // Return what the client asked no matter the SOP class
        Some(RequestorRoles { scu, scp })
    }
}

let accept_all_roles = AcceptAllRoles;
let options = ServerAssociationOptions::new()
    .with_negotiation(accept_all_roles)
    ...
```

PR #769, opened 2026-05-24
Update implementation class UID and version name for 0.10.0

Enet4:


pgimeno4d:
This is a very minor issue, but since the implementation UID claims to be a UUID-based UID, I think it should at least try to be based on a valid UUID. As such, it lacks a version and variant, per 
https://en.wikipedia.org/wiki/Universally_unique_identifier#Variants

If it is a random number, it should be version 4 or 8. The version is the first digit of the third group of hex numbers. The variant should be 10 (binary), which is in the first two bits of the fourth group of hex numbers, hence the first digit of that group should be 8, 9, a or b. Example of valid UUID:

3c06ae65-e531-**4**315-**9**6ea-74a70cf33f2d

The 4 is constant, and the 9 could be an 8, a 9, an A or a B. Currently it is

2.25.239162405814983375278132849753143205923

which translates to this UUID:

b3ed00ff-22fd-**b**c80-**9**34a-141092916023

Version 11 (0xB) does not exist. The variant value is acceptable, I suppose that's by chance.

This Python line could be used to generate a valid UUID-based UID:
```python
import uuid; '2.25.%d' % int.from_bytes(uuid.uuid4().bytes, byteorder='big', signed=False)
```
(function uuid4() generates a version 4 UUID as the one above). Alternatively:
```python
import random; '2.25.%d' % (random.randint(0, 2**128-1) & ~(0xF000C<<60) | 0x40008<<60)
```

Enet4:
Thank you for bringing this up. The UID was generated using [dicom-gen-uid](https://github.com/Enet4/dicom-gen-uid), so I may need to fix it there. That discrepancy is likely due to interpreting the bytes of the UUID in little endian byte order rather than big endian.

pgimeno4d:
Yes, I think you're right. The last three versions would be:

b3ed00ff-22fd-bc**8**0-93**4**a-141092916023
cacaebcc-f89b-ff**a**1-be**4**c-ef725da8ad32
a13b22f4-68c4-36**8**7-3c**4**f-34d53dac4041

The digits in bold are where the version and variant (in reverse order) would be when interpreted as little endian. Clearly they are there, just not in the correct place.

Enet4:
I have sent in a patch to dicom-gen-uid (https://github.com/Enet4/dicom-gen-uid/pull/1), this pull request will then need to be updated.

PR #770, opened 2026-05-26
Move extended_negotiation_filter_impl to the Association trait

pgimeno4d:
This is a simple code quality improvement. It's an oversight in #734.


Enet4 (approved):
It is a small, but welcomed simplification. :+1: Thanks!

Issue #768, opened 2026-05-19
Claude Code Review Report


qarmin:
Recently, I ran Claude Code across several projects(like image-rs, lofty of symphonia) to look for potential issues.

In this projects, a significant portion of reported findings turned out to be incorrect or minor false positives, but a non-trivial subset was valid and included real issues ranging from incorrect comments to actual logic bugs. As a result, the findings generally require manual validation to separate noise from actionable problems.

Full Reports (`findings-interactive.html` - interactive report view for manual inspection, `findings-short.md` - compact list ready to copy into GitHub, `findings-table.html` - compact tabular report version):

[findings-interactive.html](https://github.com/user-attachments/files/28030234/findings-interactive.html)
[findings-short.md](https://github.com/user-attachments/files/28030235/findings-short.md)
[findings-table.html](https://github.com/user-attachments/files/28030236/findings-table.html)

Example findings (this is only a subset of the findings, specifically those most likely to be actual bugs; for the full list, see the reports above):

=================================

### API_1 `MEDIUM`

**Description:** `storescu::run_async` always reads `App::parse()` again (line 426) even though `main` already called `App::parse()` and dispatched to `run_async()`. While clap parsing is idempotent, re-parsing means args are read twice, environment is read twice, and any side effects (like exit on `--help`) are duplicated. The sync `run` correctly takes `app: App` by argument.

**Fix:** Make `run_async` accept the `App` by value (like `run` does) and pass `app` from main.

### CPY_1 `MEDIUM`

**Description:** `storescp/src/store_sync.rs` and `store_async.rs` are essentially identical files with extensive duplicated logic for PDU loop, C-ECHO handling, C-STORE handling, file metadata building, response sending. Any bug fix must be applied to both files (and indeed bugs already differ — see the duplicate log line in store_sync.rs). This dramatically increases maintenance cost and bug risk.

**Fix:** Extract the shared PDU processing logic into a generic helper parameterised over a sync/async send trait, or wrap the sync version in `block_on` to share one implementation.

### CPY_2 `LOW`

**Description:** Duplicate `info!("New association from {}", ...)` is logged twice in the TLS branch of storescp::store_sync.rs (lines 58-59). Clearly a copy-paste bug — the async variant only logs it once.

**Locations:**

```
storescp/src/store_sync.rs:58-59
   58 |         info!("New association from {}", association.peer_ae_title());
   59 |         info!("New association from {}", association.peer_ae_title());
```

[...]

PR #771, opened 2026-05-26
Fix more Max PDU confusion

pgimeno4d:
Fix more instances of broken maximum lengths.  This is a follow-up to #733 which fixes a few oversights. Standarize to use max_pdu_size for everything and with a uniform meaning (total PDU size minus the PDU header, which is the same definition the standard uses for the PDU Length field and the Maximum PDU Length items).

Includes a patch that fixes the case where if the user configures a max_pdu_size between MAXIMUM_PDU_LENGTH+1 and u32::MAX, there's a potentially catastrophic addition wraparound (panic or impossibly small buffer).

Prompted by a bug reported by @qarmin in #768.

Fixes LOGIC_12, OVF_3 and OVF_4 from #768.

### DIFF
--- a/ul/src/association/client.rs
+++ b/ul/src/association/client.rs
@@ -381,8 +381,10 @@ impl<'a> ClientAssociationOptions<'a> {
 
     /// Override the maximum PDU length
     /// that this application entity will admit.
+    /// Values larger than MAXIMUM_PDU_SIZE will
+    /// be silently truncated to MAXIMUM_PDU_SIZE.
     pub fn max_pdu_length(mut self, value: u32) -> Self {
-        self.max_pdu_length = value;
+        self.max_pdu_length = value.min(MAXIMUM_PDU_SIZE);
         self
     }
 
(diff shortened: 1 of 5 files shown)


Enet4 (approved):
I understand that `MAXIMUM_PDU_SIZE` represents the maximum value of the _PDU-Length_ field that any implementation could ever support, even with a very relaxed interpretation of "within reason". :) I hope that this is the last mix-up involving maximum PDU sizes!

I also ran the fuzzer for a while and caught nothing.

pgimeno4d:
> I hope that this is the last mix-up involving maximum PDU sizes!

Alas, that's not the case. I've realized that there are many calls to `with_capacity()` in client.rs and server.rs with a size that is short by a PDU header, so if a full-size PDU comes, due to the fact that the header is unaccounted for, a reallocation will be necessary. I noticed when looking at the lines reported in OVF_4 but forgot about them until today, sorry.

PR #773, opened 2026-05-26
Avoid double copying of bytes in pdu/reader.rs

pgimeno4d:
This avoids one copy, but doesn't avoid the zero initialization.

Addresses PERF_4 (also reported in DEAD_2) in #768.

### DIFF
--- a/ul/src/pdu/reader.rs
+++ b/ul/src/pdu/reader.rs
@@ -324,11 +324,13 @@ pub fn read_pdu(mut buf: impl Buf, max_pdu_length: u32, strict: bool) -> Result<
                     bytes.remaining() >= (item_length - 2) as usize,
                     InvalidPduFieldLengthSnafu {},
                 );
+                let mut data = vec![0; (item_length - 2) as usize];
+                bytes.copy_to_slice(&mut data);
                 values.push(PDataValue {
                     presentation_context_id,
                     value_type,
                     is_last,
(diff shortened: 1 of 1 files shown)


Enet4 (commented):
This one, I am not so sure. Although the report claims that `copy_to_bytes` already allocates, the documentation of [Buf::copy_to_bytes](https://docs.rs/bytes/latest/bytes/buf/trait.Buf.html#method.copy_to_bytes) indicates that implementations can be optimized so that they do not allocate, and in particular, that the implementation for `Bytes` (as is the case here) will return a shallow ref-counted copy. An inspection of the source code confirms this, as it only calls method `to_split`, which in turn is O(1).

I ran the tests on top of DHAT a few times and that did not give me significant differences either, though I could be misinterpreting the results.

Until we can confirm that the current reader functions have an unnecessary allocation here, I would refrain from moving forward with this PR.

pgimeno4d:
I'm happy with it being closed. `copy_to_bytes` seems like a misleading name that made me believe what the AI said. Apologies for wasting your time.

In fact, I think having it open will only be disturbing, so I'm closing it for the time being. It can always be reopened at some point.

Enet4:
> I'm happy with it being closed. `copy_to_bytes` seems like a misleading name that made me believe what the AI said. Apologies for wasting your time.

I suppose a case could be made here on the misleading method name, regardless of whether a human or AI inspected it, but greater caution is advised, as these models can be unpredictably erratic.

I still thank you for taking your time going through the report!

PR #774, opened 2026-05-26
Fix recursion issue with possibly malicious file

pgimeno4d:
To ease review, the commit has been separated in three parts:

- The first commit applies the functional changes, but does not change indentation.
- The other two commits are purely cosmetic: one indents (just whitespace changes) and the other is a run of `rustfmt --edition 2024`.

Fixes LOOP_1 from https://github.com/Enet4/dicom-rs/issues/768.

### DIFF
--- a/parser/src/dataset/read.rs
+++ b/parser/src/dataset/read.rs
@@ -369,7 +369,7 @@ where
                                                 len: len.0,
                                             }
                                             .fail(),
-                                        )
+                                        );
                                     }
                                 };
                                 // entered a new item
@@ -383,7 +383,7 @@ where
                                                 bytes_read: self.parser.position(),
                                             }
                                             .fail(),
-                                        )
+                                        );
                                     }
                                 };
                                 self.push_sequence_token(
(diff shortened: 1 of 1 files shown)


pgimeno4d:
Admittedly, this patch impairs the maintainability of the affected function.

I've just learned that there's an experimental [`become`](https://doc.rust-lang.org/stable/std/keyword.become.html) feature that implements manual tail calls. When/if this becomes stable, the loop trick can be replaced by a tail call, making the code clean as before again. I tried implementing this through the `tailcall` crate but it caused borrow issues that I decided not to dive into.

Enet4:
It mostly comes down to whether this big `loop` block is acceptable or not. One way I can think of to mitigate the effect of that level of indentation is by moving the current implementation to a private method. It would return `Option<Option<T>>`, or a custom type of our own:

```rust
enum NextItem<T> {
   Return(Option<T>),
   Continue,
}
```

Then at `next` we would know whether to try again or return immediately:

```rust
fn next(&mut self) -> Option<Self::Item> {
    loop {
        match self.next_impl() {
            NextItem::Return(o) => o,
            NextItem::Continue => continue,
        }
    }
}
```

So the pattern would be similar to async polling.

pgimeno4d:
I thought of that but rejected it because it implied changing every return point, and there are a number. It has better maintainability compared to the other solution, but at the cost of complication and of making the patch big.

Also, if/when `become` is ready, that will be harder to undo than the loop approach.

Still, if you want me to try that approach I'll be happy to oblige.

Enet4 (approved):
I will allow this. Thanks! :+1:

PR #777, opened 2026-05-27
[encoding] Check for overflow of u16 before writing the header

pgimeno4d:
Fixes OVF_6 of #768.


Enet4 (approved):
I guess that is right. Should someone try to create something in memory which is too big for the associated VR, the least we can do is prevent it from producing and disseminating inconsistent DICOM data. Thank you! :+1:

PR #778, opened 2026-05-28
Make buffer allocations more precise

pgimeno4d:
Ensure all buffers have room for the maximum they specify plus the header size. Use self.max_pdu_length limited to LARGE_PDU_SIZE for receiving, and DEFAULT_MAX_PDU for sending.

Also, the -1 in the formula for MAXIMUM_PDU_SIZE was actually wrong. The resulting value is the same without it though, but the Clippy warning is different.

Worth noting that DEFAULT_MAX_PDU may be too short from the start in many common circumstances. I've observed an A-ASSOCIATE-RQ PDU sent by a DCMTK's `getscu` client that is 17503 bytes long, for instance: just a default query without even any extended negotiation items. The bulk of the data in that case comes from the presentation contexts and the role negotiation (one of each for every SOP Class supported, with three transfer syntaxes per SOP Class in every presentation context). You can blame the complexity of the DICOM standard for this.

### DIFF
--- a/ul/src/pdu/mod.rs
+++ b/ul/src/pdu/mod.rs
@@ -37,8 +37,8 @@ pub const MINIMUM_PDU_SIZE: u32 = 1_024 - PDU_HEADER_SIZE;
 /// and when adding the PDU header size, it must not overflow.
 // `clippy` complains that !1 and u32::MAX-1 are equal, but this
 // formulation expresses intent, so disable the warning
-#[allow(clippy::eq_op)]
-pub const MAXIMUM_PDU_SIZE: u32 = ((u32::MAX - 1) & !1) - PDU_HEADER_SIZE;
+#[allow(clippy::identity_op)]
+pub const MAXIMUM_PDU_SIZE: u32 = (u32::MAX & !1) - PDU_HEADER_SIZE;
 
 /// A generous PDU size for internal use.
 /// Prefer to initialize buffers no larger than this size
--- a/ul/src/pdu/mod.rs
+++ b/ul/src/pdu/mod.rs
@@ -26,7 +26,7 @@ pub const PDU_HEADER_SIZE: u32 = 6;
 pub const PDV_HEADER_SIZE: u32 = 6;
 
 /// The default maximum PDU size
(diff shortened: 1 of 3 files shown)


ul/src/pdu/mod.rs
```diff
@@ -37,8 +37,8 @@ pub const MINIMUM_PDU_SIZE: u32 = 1_024 - PDU_HEADER_SIZE;
 /// and when adding the PDU header size, it must not overflow.
 // `clippy` complains that !1 and u32::MAX-1 are equal, but this
```
Enet4: This comment might be slightly outdated, now that we don't have `u32::MAX-1`.

Enet4 (commented):
Looks good! This makes memory reservations more predictable, and indeed will be more efficient for PDU sizes around the maximum.

> Worth noting that DEFAULT_MAX_PDU may be too short from the start in many common circumstances. I've observed an A-ASSOCIATE-RQ PDU sent by a DCMTK's getscu client that is 17503 bytes long, for instance: just a default query without even any extended negotiation items. The bulk of the data in that case comes from the presentation contexts and the role negotiation (one of each for every SOP Class supported, with three transfer syntaxes per SOP Class in every presentation context). You can blame the complexity of the DICOM standard for this.

Unless there are undesired side effects, I would say that we can expand the default MAXU PDU size for 0.10, perhaps to double the size?

---

Also, minor comment inline.

pgimeno4d:
> Unless there are undesired side effects, I would say that we can expand the default MAXU PDU size for 0.10, perhaps to double the size?

DEFAULT_MAX_PDU is pretty new anyway. That's only the maximum size before reallocation, so it shouldn't have any negative performance or functionality effects. The only potential trouble I can imagine is excessive allocation, but you need something in the order of 100,000 simultaneous associations for that to use something in the order of 10 GB. By that time, I presume you'd have worse problems than the output buffer allocation anyway.

An argument can perhaps be made that many little things can end up making a big thing, though, and if all the buffers are increased because they are individually insignificant, the total can become significant, but other than that I think it's pretty safe.

I've fixed the comment and doubled DEFAULT_MAX_PDU in a single commit. Thanks for the review!

Enet4 (approved):
No more concerns to report. Thank you!

Issue #722, opened 2026-01-05
Add better error message, when trying to connect non-tls client to tls server and vice-versa


qarmin:
Currently sending dcm from tls client to non-tls server, results in such error messages
SCU Error
```
2026-01-05T14:08:07.685321Z  INFO dicom_app_common: Using client certificate authentication
2026-01-05T14:08:07.712494Z ERROR dicom_storescu: Could not initialize SCU

Caused by these errors (recent errors listed first):
  1: failed to send PDU message on wire
  2: unexpected end of file

```
SCP Error
```
026-01-05T14:08:07.712372Z ERROR dicom_storescp: could not establish association

Caused by these errors (recent errors listed first):
  1: failed to receive pdu *
  2: Incoming pdu was too large: length 17149697, maximum is 16378

NOTE: Some redundant information has been removed from the lines marked with *. Set SNAFU_RAW_ERROR_MESSAGES=1 to disable this behavior.
```

Such errors are not really helpful to diagnose real problem.

I hit such problem, when my custom client could not read for some reason its certificate(still don't know why this happen, so maybe I will create another issue later) and  probably used non-tls connection(not 100% sure if this happened, but error is same as with tls <-> non-tls communication)
```
15:25:23.384 [ERROR] c_store: Failed to build TLS client config: PemParse { source: NoItemsFound, path: Some("/home/rafal/certs//client-cert.pem") }
15:25:23.392 [ERROR] c_store: Error sending file /home/rafal/ap600_opv.dcm(Ophthalmic Visual Field Static Perimetry Measurements Storage)(15.78 KiB): SCU initialization error: failed to receive pdu: Incoming pdu was too large: length 50332162, maximum is 16384 in 134.7361ms
```

SCU command
```
RUST_LOG=trace dicom-storescu  127.0.0.1:11111   /home/rafal/SPRAWDZONE/dcms_po_kazdym_po_jednym/ap600_opv.dcm    --tls   --key certs/client-key.pem   --cert certs/client-cert.pem   --add-certs certs/ca-cert.pem  --called-ae-title OCULINKDB
```
SCP command
```
RUST_LOG=trace dicom-storescp --promiscuous -s -v --key certs/key.pem --cert certs/cert.pem --add-certs certs/ca-cert.pem --allow-unauthenticated --calling-ae-title OCULINKDB
```

Script to generate certs
```
#!/usr/bin/env bash
set -euo pipefail

BASE_DIR="$(pwd)/certs"
[...]

PR #779, opened 2026-06-01
ENH: Add clearer error messages for TLS connection issues

naterichman:
* For both client and server, drive handshake to completion at association establishment, reporting on TLS connection errors.
* For both client and server, provide best effort checking for mismatch where local doesn't support TLS but remote has TLS enabled

This should solve the issues reported in #722 @qarmin


storescu/src/main.rs
```diff
@@ -523,10 +523,19 @@ async fn run_async() -> Result<(), Error> {
         });
     }
     while let Some(result) = tasks.join_next().await {
-        if let Err(e) = result {
-            error!("{}", Report::from_error(e));
-            if fail_first {
-                std::process::exit(-2);
+        match result {
```
naterichman, line 526: This was a sort of unrelated bug fix, errors were being silently ignored since we only checked the join result, not the inner

naterichman:
This may need to be [updated when rustls 0.24 comes out](https://github.com/rustls/rustls/issues/3071#issuecomment-4593115085)

ul/Cargo.toml
```diff
@@ -17,6 +17,7 @@ bytes = "1.11.1"
 cfg-if = "1.0.3"
 dicom-encoding = { path = "../encoding/", version = "0.9.1" }
 dicom-transfer-syntax-registry = { path = "../transfer-syntax-registry/", version = "0.9.1", default-features = false }
+dicom-app-common = { version = "0.9", path = "../app-common", optional = true }
```
Enet4: Can we avoid having to add `dicom-app-common` here? The crate was designed for use by end-user tools, so it adds a lot of other dependencies which we do not want to include in a library.

ul/tests/association.rs
```diff
@@ -1,295 +1,12 @@
...
 use dicom_ul::{ClientAssociationOptions, Pdu, ServerAssociationOptions};
 use rstest::rstest;
-#[cfg(feature = "sync-tls")]
-use std::sync::Arc;
 use std::time::Instant;
 
 type Result<T, E = Box<dyn std::error::Error + Send + Sync + 'static>> = std::result::Result<T, E>;
 
-#[cfg(feature = "sync-tls")]
-fn ensure_test_certs() -> Result<()> {
```
Enet4, line 13: We're dropping a bunch of tests, is this intentional?

Enet4 (changes requested):
Thanks! I took the liberty of resolving merge conflicts from #778. More concerns in the comments below.

ul/Cargo.toml
naterichman: Let me look into this, I pulled it in to avoid the boilerplate of building `ClientConfig` in rustls, but I can probably remove that

ul/tests/association.rs
naterichman: I moved these all into a separate `ul/tests/association_tls.rs`, they are all still there, and I added ~8 more tests.

ul/src/association/client.rs
```diff
@@ -921,7 +925,82 @@ impl<'a> ClientAssociationOptions<'a> {
         let mut buf = BytesMut::with_capacity(
             (self.max_pdu_length.min(LARGE_PDU_SIZE) + PDU_HEADER_SIZE) as usize,
         );
-        let resp = read_pdu_from_wire(&mut socket, &mut buf, self.max_pdu_length, self.strict)?;
+        let resp = read_pdu_from_wire(&mut socket, &mut buf, self.max_pdu_length, self.strict)
+            .map_err(|e| {
+                // // If we're in non-TLS mode and `read_pdu_from_wire` fails, it
```
Enet4: Is this comment block supposed to have been cleaned up? It's a lot of code behind comments.

ul/src/association/client.rs
```diff
@@ -921,7 +925,82 @@ impl<'a> ClientAssociationOptions<'a> {
...
+                    let res = acceptor.read_tls(&mut read);
+                    if res.is_err() {
+                        return e;
+                    }
+                    match acceptor.accept() {
+                        Ok(_) => {
+                            // This means the server responded with a valid TLS
+                            // state, but since we didn't originally send a TLS
+                            // ClientHello, that shouldn't be possible
+                            error!("Recieved TLS response to non-TLS request!");
```
Enet4:
```suggestion
                            error!("Received TLS response to non-TLS request!");
```

ul/src/association/client.rs
```diff
@@ -921,7 +925,82 @@ impl<'a> ClientAssociationOptions<'a> {
...
+                            return super::TlsNotSupportedSnafu.build()
+                        }
+                        Err((err, _alert)) => {
+                            // Recieved a valid TLS message, means the server expects TLS
+                            if let rustls::Error::InappropriateMessage{..} = err {
+                                error!("Recieved TLS response to non-TLS request!");
```
Enet4:
```suggestion
                                error!("Received TLS response to non-TLS request!");
```

ul/src/association/client.rs
```diff
@@ -1401,8 +1492,43 @@ impl<'a> ClientAssociationOptions<'a> {
...
+                    match LazyConfigAcceptor::new(Acceptor::default(),cursor).await{
+                        Ok(_res) => {
+                            // This means the server responded with a valid TLS
+                            // state, but since we didn't originally send a TLS
+                            // ClientHello, that shouldn't be possible
+                            return Err(e)
+                        },
+                        Err(err) => {
+                            if let Ok(rustls::Error::InappropriateMessage{..}) = err.downcast::<rustls::Error>() {
+                                    error!("Recieved TLS response to non-TLS request!");
```
Enet4:
```suggestion
                                    error!("Received TLS response to non-TLS request!");
```

Enet4 (commented):
I tried it out and it seems to mostly work. This made me realize that it is not yet possible to set some timeouts in neither `storescu` nor `storescp`, which would have been useful for this, as the async store SCU was getting stuck forever when trying to connect to a store SCP without TLS.

But this would be something worth improving separately. For now, we can attend to the concerns in the comments inline with the code and proceed.

Enet4 (approved):

PR #783, opened 2026-06-05
BUG: Fix hanging of AsyncPdataWriter

### LINKED ISSUE #780: AsyncPDataWriter hangs indefinitely when writing large DICOM data (> Max PDU Length)
## Bug Description

`AsyncPDataWriter::write_all` (and `write`) hangs indefinitely when the input data size exceeds the negotiated Max PDU Length. The future never resolves and no error is returned. This appears to be an internal deadlock/stall in the writer's backpressure or PDU fragmentation logic, rather than a network or peer issue.

## Reproduction Steps

1. Establish an async DICOM association using `AsyncClientAssociation`.
2. Obtain a `PDataWriter` via `send_pdata(pc_id)`.
3. Call `write_all` with a buffer larger than the negotiated Max PDU Length (e.g., 10 MB image data with default 16 KB Max PDU).
4. The `.await` never returns.

## Expected Behavior

`write_all` should complete after internally fragmenting the data into multiple P-DATA-TF PDUs and transmitting them all.

## Actual Behavior

The task hangs forever at `pdata.write_all(&object_data).await`. Adding a `tokio::time::timeout` confirms the future is stalled with no network activity.

## Environment

- `dicom-rs` version: `dicom-ul` latest (crates.io / main)
- `dicom-rs` feature flags: `async`
- Async runtime: `tokio` (multi-thread)
- OS: macOS (Apple Silicon) / Linux
- Peer: Any PACS/SCP (tested against multiple independent implementations)

## Minimal Reproduction Code

```rust
use dicom_ul::association::ClientAssociationOptions;
use tokio::io::AsyncWriteExt;

#[tokio::main]
async fn main() -&gt; Result&lt;(), Box&lt;dyn std::error::Error&gt;&gt; {
    let mut scu = ClientAssociationOptions::new()
        .calling_ae_title("TEST_SCU")
        .called_ae_title("TEST_SCP")
        .establish_async("127.0.0.1:11112")
        .await?;

    let pc_id = scu.presentation_contexts()[0].id;
    let mut pdata = scu.send_pdata(pc_id);

    // Any data larger than negotiated Max PDU Length triggers the hang
    let large_object = vec![0u8; 10 * 1024 * 1024]; // 10 MB

    // This line never returns
    pdata.write_all(&large_object).await?;
    pdata.finish().await?;

    Ok(())
}

### PR DESCRIPTION
**Pull request by naterichman**
* The previous implementation was returning `Poll::Pending` even when the underlying stream wrote some data. Since the underlying stream is responsible for setting up the waker, but only does so when it returns `Poll::Pending`, this means the `poll_write` was never being woken up and would hang forever
* The new implementation loops over calls to the underlying stream for partial writes until the stream either writes all, or switches to `Poll::Pending`, in which case the rest of the original logic works
* Added a test using the code which reproduced this issue

Closes #780

### CODE REVIEW

File: ul/src/association/pdata.rs [hunk 1]
```diff
@@ -415,9 +416,13 @@ pub mod non_blocking {
     /// ```
     #[must_use]
     pub struct AsyncPDataWriter<W: AsyncWrite + Unpin> {
+        // Buffer for accumulating P-Data fragments
         buffer: Vec<u8>,
+        // Underlying stream
         stream: W,
         max_pdu_length: u32,
+        // State machine tracking whether we're currently writing to the
+        // underlying stream and how much of the buffer we've written
```
Enet4, line 433: You can use doc-comments (`///`) for these, though it isn't much of a big deal since the fields are private.

ul/src/association/pdata.rs
```diff
@@ -513,48 +518,70 @@ pub mod non_blocking {
...
+                        loop {
+                            let this: &mut AsyncPDataWriter<W> = self.as_mut().get_mut();
+                            match Pin::new(&mut this.stream).poll_write(cx, &this.buffer[written..]) {
+                                Poll::Ready(Ok(n)) => {
+                                    // Underlying writer wrote something.
+                                    written += n;
+                                    // If we wrote the whole buffer, we can try
+                                    // to write the rest immediately.  It is
+                                    // important we don't return
+                                    // `Poll::Pending`` here, since the
```
Enet4:
```suggestion
                                    // `Poll::Pending` here, since the
```

Enet4 (commented):
Thank you for this PR @naterichman! Tests in my dev machine and in CI seem to pass after the change, but considering https://github.com/Enet4/dicom-rs/issues/780#issuecomment-4639139008 I wonder if there are more edge cases still uncovered. If @Kinny2000 is still available, we can do a temporary commit with trace logs and try to figure out how it gets stuck.

pgimeno4d:
~~Seems to work. **[Edit: before the last commit, that is]**~~ [Edit2: Not actually, see next comment]

This is the AsyncWriter crafted by the AI (modified by me to take a mutable reference to a buffer instead of an owned buffer):

```rust
    /// A mock AsyncWrite that only accepts `chunk_size` bytes per poll_write call,
    /// forcing the AsyncPDataWriter to handle partial writes via the Writing state.
    #[cfg(feature = "async")]
    struct PartialWriter<'b> {
        buf: &'b mut Vec<u8>,
        chunk_size: usize,
    }

    #[cfg(feature = "async")]
    impl tokio::io::AsyncWrite for PartialWriter<'_> {
        fn poll_write(
            mut self: std::pin::Pin<&mut Self>,
            _cx: &mut std::task::Context<'_>,
            buf: &[u8],
        ) -> std::task::Poll<std::io::Result<usize>> {
            let n = buf.len().min(self.chunk_size);
            self.buf.extend_from_slice(&buf[..n]);
            std::task::Poll::Ready(Ok(n))
        }

        fn poll_flush(
            self: std::pin::Pin<&mut Self>,
            _cx: &mut std::task::Context<'_>,
        ) -> std::task::Poll<std::io::Result<()>> {
            std::task::Poll::Ready(Ok(()))
        }

        fn poll_shutdown(
            self: std::pin::Pin<&mut Self>,
            _cx: &mut std::task::Context<'_>,
        ) -> std::task::Poll<std::io::Result<()>> {
            std::task::Poll::Ready(Ok(()))
        }
    }
```

It seems trivial enough, so maybe it could be included in the tests like this:

```diff
@@ -872,9 +907,13 @@ mod tests {
         let my_data: Vec<_> = (0..2500).map(|x: u32| x as u8).collect();
 
         let mut buf = Vec::new();
         {
+            let stream = PartialWriter {
+                buf: &mut buf,
+                chunk_size: 8,
+            };
             let mut writer =
-                AsyncPDataWriter::new(&mut buf, presentation_context_id, MINIMUM_PDU_SIZE);
+                AsyncPDataWriter::new(stream, presentation_context_id, MINIMUM_PDU_SIZE);
             writer.write_all(&my_data).await.unwrap();
             writer.finish().await.unwrap();
         }
```

pgimeno4d:
Actually, disregard that. When I modified `PartialWriter` so it doesn't always return `Poll::Ready` but returns `Poll::Pending` and `Poll::Ready` alternatively, and a waker that immediately wakes it up when pending, the test freezes and hogs all CPU and RAM in the system until something kills it. For more than one minute my system remained unresponsive, with not even the mouse moving, so beware of doing that.

I asked the AI to write a version that uses `tokio::timer::sleep` to avoid freezing my system for a while, and the test just freezes, but the system keeps being functional:

```rust
#[cfg(feature = "async")]
struct PartialWriter<'a> {
    buf: &'a mut Vec<u8>,
    chunk_size: usize,
    sleep: Option<std::pin::Pin<Box<tokio::time::Sleep>>>,
}

#[cfg(feature = "async")]
impl<'a> PartialWriter<'a> {
    fn new(buf: &'a mut Vec<u8>, chunk_size: usize) -> Self {
        Self { buf, chunk_size, sleep: None }
    }
}

#[cfg(feature = "async")]
impl<'a> tokio::io::AsyncWrite for PartialWriter<'a> {
    fn poll_write(
        mut self: std::pin::Pin<&mut Self>,
        cx: &mut std::task::Context<'_>,
        buf: &[u8],
    ) -> std::task::Poll<std::io::Result<usize>> {
        if let Some(sleep) = self.sleep.as_mut() {
            match sleep.as_mut().poll(cx) {
                std::task::Poll::Pending => return std::task::Poll::Pending,
                std::task::Poll::Ready(()) => {
                    self.sleep = None;
                }
            }
        } else {
            self.sleep = Some(Box::pin(tokio::time::sleep(
                std::time::Duration::from_millis(1),
            )));
            let sleep = self.sleep.as_mut().unwrap();
            match sleep.as_mut().poll(cx) {
                std::task::Poll::Pending => return std::task::Poll::Pending,
                std::task::Poll::Ready(()) => {
                    self.sleep = None;
                }
            }
        }

        let n = buf.len().min(self.chunk_size);
        self.buf.extend_from_slice(&buf[..n]);
        std::task::Poll::Ready(Ok(n))
    }

    fn poll_flush(
        self: std::pin::Pin<&mut Self>,
        _cx: &mut std::task::Context<'_>,
    ) -> std::task::Poll<std::io::Result<()>> {
        std::task::Poll::Ready(Ok(()))
    }

    fn poll_shutdown(
        self: std::pin::Pin<&mut Self>,
        _cx: &mut std::task::Context<'_>,
    ) -> std::task::Poll<std::io::Result<()>> {
        std::task::Poll::Ready(Ok(()))
    }
}
```
This happens either with or without the second commit (after fixing the missing semicolon).

naterichman:
@Kinny2000 will you try this again? I added a few more fixes I caught with tests and/or review from claude

* Handle explicit 0-write -- This will now return an error instead of potentially spinning forever
* Fix state of buffer in `Writing` state -- this caused data to be sent multiple times over the wire, which would lead to invalid PDUs

Those are the only two that I could affect what you have.  Let me know if these fix the stability issues you're seeing.

pgimeno4d:
With the latest changes, the tests no longer freeze when using the above PartialWriter.

Kinny2000:
> Thank you for this PR @naterichman! Tests in my dev machine and in CI seem to pass after the change, but considering [#780 (comment)](https://github.com/Enet4/dicom-rs/issues/780#issuecomment-4639139008) I wonder if there are more edge cases still uncovered. If @Kinny2000 is still available, we can do a temporary commit with trace logs and try to figure out how it gets stuck.

I tried the new version, it's better now,  but sometimes it still hangs up.

> @Kinny2000 will you try this again? I added a few more fixes I caught with tests and/or review from claude
> 
> * Handle explicit 0-write -- This will now return an error instead of potentially spinning forever
> * Fix state of buffer in `Writing` state -- this caused data to be sent multiple times over the wire, which would lead to invalid PDUs
> 
> Those are the only two that I could affect what you have. Let me know if these fix the stability issues you're seeing.

I tried this version，it seems stable enough。

Kinny2000:
Yes,it's stable enough now.

ul/src/association/pdata.rs
```diff
@@ -429,7 +442,7 @@ pub mod non_blocking {
         /// Construct a new P-Data value writer.
         ///
         /// `max_pdu_length` is the maximum value of the PDU-length property.
-        pub(crate) fn new(stream: W, presentation_context_id: u8, max_pdu_length: u32) -> Self {
+        pub fn new(stream: W, presentation_context_id: u8, max_pdu_length: u32) -> Self {
```
Enet4: The `new` functions in the other `PData*` types have been kept private, I would prefer it that way for consistency. I think it's fine that we move the tests from `tests/pdata` on to the end of module, so that they can access the constructor at pub(crate) level.

Enet4 (commented):
This looks very promising! There is just one API concern that I would like to resolve before we merge.

Enet4 (approved):
Sounds good. Thank you all for your collaboration!

PR #751, opened 2026-03-20
JPEG 2000 support via the hayro-jpeg2000

bowenxuuu:
Hello @Enet4 , 

I'm trying to integrate the hayro-jpeg2000 crate into DICOM-rs, but encountered several issues: the final rendered image does not match the expected outcome, and I have no idea where the problem lies. 
Could you please give me some guidance?

This image is decoded using `openjpeg-sys`.
openjpeg-sys

This image is decoded using `hayro-jpeg2000`.
hayro-jpeg2000

PR #787, opened 2026-06-14
JPEG 2000 support via the hayro-jpeg2000

### LINKED ISSUE #730: `hayro-jpeg2000` A memory-safe, pure-Rust JPEG 2000 decoder.
@Enet4 Hi, 
[hayro-jpeg2000](https://crates.io/crates/hayro-jpeg2000) crate supports accelerating the decoding process via SIMD.

This crate got picked up by GNOME as its jpeg2000 implementation, and now jpeg2000 decoding in GNOME is enabled by default in the latest beta release: https://gitlab.gnome.org/GNOME/glycin/-/commit/8892184ac485b493bb6aac7450b805496b921e44

> Use hayro-jpeg2000 loader instead of jpeg2k and make it part of the glycin-image-rs loader, enabled by default.

bowenxuuu:
follow up #751

fix #730


bowenxuuu:
decode by `openjpeg-sys`


decode by `hayro-jpeg2000`



@LaurenzV 
The colors decoded by hayro-jpeg2000 appear incorrect, but I haven't yet figured out the root cause.

transfer-syntax-registry/src/adapters/jpeg2k.rs
```diff
@@ -63,33 +71,59 @@ impl PixelDataReader for Jpeg2000Adapter {
...
+            debug_assert_eq!(dst.as_ptr() as usize % align_of::<u16>(), 0);
+
+            // dst is aligned for u16, decode directly into dst_u16
+            let (prefix, dst_u16, suffix) = unsafe { dst.align_to_mut::<u16>() };
+            if prefix.is_empty() && suffix.is_empty() {
+                image
+                    .decode_u16_into(dst_u16, &mut DecoderContext::default())
+                    .expect("hayro-jpeg2000 decoder failure");
+            }
+        }
```
bowenxuuu, line 141: I use `unsafe` here to re-interpret `dst: &mut Vec<u8>` as `dst_u16: &mut Vec<u16>`. This reuses the existing buffer to avoid extra memory allocation and data copying. I'd like to hear your thoughts. @Enet4

bowenxuuu:
https://github.com/LaurenzV/hayro/blob/20888608e8682f57a23632924fd87d2858fca359/hayro-jpeg2000/src/lib.rs#L550-L551

After I changed `T::from_scaled_f32()` to `T::from_native_f32()`, everything works as expected.

This image decode by `T::from_native_f32()` of `hayro-jpeg2000`.

LaurenzV:
Is this image available somewhere so I can try it myself?

transfer-syntax-registry/src/adapters/jpeg2k.rs
LaurenzV: Couldn't you use bytemuck for this?

bowenxuuu:
Here's the DICOM image.

[img.zip](https://github.com/user-attachments/files/28922561/img.zip)

LaurenzV:
Okay, so I get it now. The method I added always scales to 16-bit, but what you want is something that always preserves the native bit depth. I think it's best if I instead expose an API for the raw f32 samples, then this should be the same as you are currently doing.

LaurenzV:
Could you try this one instead? https://github.com/LaurenzV/hayro/pull/1243

I think the code should then look something like this:
```rs
#[cfg(feature = "hayro-jpeg2000")]
{
    use hayro_jpeg2000::{DecodeSettings, DecoderContext, Image};

    let image = Image::new(&frame_data, &DecodeSettings::default())
        .whatever_context("hayro-jpeg2000 read image failure")?;

    let mut ctx = DecoderContext::default();
    let decoded = image
        .decode(&mut ctx)
        .whatever_context("hayro-jpeg2000 decoder failure")?;

    let components = decoded.components();

    for (component_i, component) in components.iter().enumerate() {
        if component_i >= samples_per_pixel as usize {
            warn!(
                "JPEG 2000 image has more components than expected ({} >= {})",
                component_i, samples_per_pixel
            );
            break;
        }

        for (i, sample) in component.samples().iter().enumerate() {
            let offset = base_offset
                + i * samples_per_pixel as usize * bytes_per_sample as usize
                + component_i * bytes_per_sample as usize;

            let sample = sample.round() as u32;
            dst[offset..offset + bytes_per_sample as usize]
                .copy_from_slice(&sample.to_le_bytes()[..bytes_per_sample as usize]);
        }
    }
}
```

bowenxuuu:
@LaurenzV Thanks, it works well now.

LaurenzV:
Should be included in version 0.4.0 now. :)

bowenxuuu:
@Enet4 Hello

PR #789, opened 2026-06-14
Feature-gate TLS support in dicom-storescu and dicom-storescp

Enet4:
This PR makes it possible to build the tools `storescu` and `storescp` without TLS support, and makes this the default behavior. To make it possible to do DICOM over TLS in `dicom-storescu` and `dicom-storescp` (`--tls` and other options), enable Cargo feature **"tls"**. Library crates (other than `dicom-app-common`) are unaffected.

### Summary

- add Cargo feature "tls" to dicom-app-common and affected tool packages
   - add also "tls" to dicom-ul as an alias to "sync-tls"
- [app-common] change TLS options to be hidden when Cargo feature "tls" is disabled
- [storescu] [storescp] only include rustls, rustls-native-certs, and tokio-rustls when "tls" feature is enabled
- [storescu] [storescp] add feature gating on TLS functionality
- [ci] include "tls" feature in the test project step where extra feature are included

### DIFF
--- a/app-common/src/lib.rs
+++ b/app-common/src/lib.rs
@@ -1,8 +1,12 @@
 use clap::Args;
+#[cfg(feature = "tls")]
 use rustls::{ClientConfig, ServerConfig, SupportedProtocolVersion, pki_types::{CertificateDer, CertificateRevocationListDer, PrivateKeyDer, pem::PemObject}, server::WebPkiClientVerifier};
+#[cfg(feature = "tls")]
 use tracing::{debug, info};
-use std::{path::PathBuf, sync::Arc};
-use snafu::{Snafu, ResultExt};
+use std::path::PathBuf;
+#[cfg(feature = "tls")]
+use std::sync::Arc;
+use snafu::prelude::*;
 
 #[derive(Snafu, Debug)]
 pub enum MissingPemObject {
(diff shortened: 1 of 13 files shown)


app-common/src/lib.rs
```diff
@@ -16,62 +20,79 @@ pub enum MissingPemObject {
...
+    /// the application was not built with TLS support
+    #[cfg(not(feature = "tls"))]
+    TlsSupportNotAvailable,
+}
 
 #[derive(Args, Debug)]
 pub struct TlsOptions {
     /// Enables mTLS (TLS for DICOM connections)
     #[arg(long = "tls", default_value = "false")]
+    #[arg(hide(cfg!(not(feature = "tls"))))]
```
naterichman, line 50: Can the whole struct just be put behind the `tls`?

storescp/src/main.rs
naterichman, line 58: And then this could be: 

```suggestion
   #[cfg(feature = "tls")]
   #[command(flatten, next_help_heading = "TLS Options")]
tls: TlsOptions,
```

naterichman (commented):
I understand the reason you did it like you did is to allow returning an error of "TLS support not available", but I'm wondering if that's necessary.

app-common/src/lib.rs
Enet4: I think I tried to put the `hide` property at struct level and that didn't work. I did not want to make `TlsOptions` disappear because the stub helps fulfill the feature gating and indicate that TLS support is available behind a Cargo feature. The alternative is to add the same `#[arg(hide(...))]` on every `tls` options group field of each application, so I preferred to do this once per `TlsOptions` field, which is harder to forget. Maybe there is a better way?

storescp/src/main.rs
Enet4: But if I do that, adding `--tls` as an option will result in an "unexpected argument '--tls' found" error message, which is different from telling users that the application needs to be built with TLS support. Same for any TLS option provided.

app-common/src/lib.rs
naterichman: Makes sense, I figured thats why you did it this way, just checking!

Enet4:
Thank you for taking a look, @naterichman! I answered to the other points inline. I feel that the default message provided by clap is really problematic, as it leaves users completely clueless about why `--tls` isn't working while the rest of the documentation would state that it should work.

naterichman (approved):

PR #791, opened 2026-06-15
feature flag: openjpeg-sys-threads

bowenxuuu:
When the `openjpeg-sys-threads` feature is enabled, `openjpeg-sys` will be activated automatically now.

### DIFF
--- a/transfer-syntax-registry/Cargo.toml
+++ b/transfer-syntax-registry/Cargo.toml
@@ -54,7 +54,7 @@ charls = ["dep:charls"]
 charls-vcpkg = ["charls?/vcpkg"]
 
 # build OpenJPEG with multithreading,
-# implies "rayon"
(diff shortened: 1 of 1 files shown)


transfer-syntax-registry/Cargo.toml
```diff
@@ -55,7 +55,7 @@ charls-vcpkg = ["charls?/vcpkg"]
 
 # build OpenJPEG with multithreading,
 # implies "rayon"
```
Enet4:
```suggestion
# implies "openjpeg-sys" and "rayon"
```

Enet4 (commented):
The intent was for `openjpeg-sys-threads` to always need to be combined with `openjpeg-sys`, but we can do as you suggest.

Let's update the comments accordingly.

Enet4 (approved):
Thanks. :+1:

PR #792, opened 2026-06-19
[tools] Add connection timeout options via CLI flags to DIMSE Tools

karanrane96:
- [app-common] add ConnectionOptions struct with --read-timeout and --write-timeout flags, shared across
  all tools
  - [echoscu] add dicom-app-common dependency; add --connect-timeout, --read-timeout, --write-timeout flags
  - [findscu] add dicom-app-common dependency; add --connect-timeout, --read-timeout, --write-timeout flags
  - [storescu] add --connect-timeout, --read-timeout, --write-timeout flags; wire timeouts in both sync and
  async paths
  - [storescp] add --read-timeout, --write-timeout flags (no --connect-timeout — server does not initiate
  TCP connections)

**Notes:** 
 - --connect-timeout sets a deadline on TCP connection establishment (TcpStream::connect_timeout),
  preventing indefinite hangs when the remote host is unreachable
  - --read-timeout / --write-timeout bound socket I/O, preventing hangs on stalled associations
  - All timeout values are in whole seconds


app-common/src/lib.rs
```diff
@@ -199,6 +199,18 @@ pub fn show_cipher_suites() {
...
+pub struct ConnectionOptions {
+    /// Read timeout for the underlying TCP socket in seconds
+    #[arg(long = "read-timeout", value_name = "SECS")]
+    pub read_timeout: Option<u64>,
+
+    /// Write timeout for the underlying TCP socket in seconds
+    #[arg(long = "write-timeout", value_name = "SECS")]
+    pub write_timeout: Option<u64>,
+}
+
```
Enet4, line 227: These timeouts can be `std::time::Duration`s, with the right value parser (don't forget to `use std::time::Duration`). This is semantically more accurate, and allows us to extend the parser to support different precision levels (millisecond, second, minute) in the future.

Same concern in other places using `u64` instead of `Duration` (e.g. echoscu/srcmain.rs:38).

```suggestion
/// Connection timeout options shared across DIMSE tools
#[derive(Args, Debug, Default)]
pub struct ConnectionOptions {
    /// Read timeout for the underlying TCP socket in seconds
    #[arg(
        long = "read-timeout",
        value_name = "SECS",
        value_parser(parse_duration_sec)
    )]
    pub read_timeout: Option<Duration>,

    /// Write timeout for the underlying TCP socket in seconds
    #[arg(
        long = "write-timeout",
        value_name = "SECS",
        value_parser(parse_duration_sec)
    )]
    pub write_timeout: Option<Duration>,
}

fn parse_duration_sec(arg: &str) -> Result<Duration, std::num::ParseIntError> {
    let seconds = arg.parse()?;
    Ok(std::time::Duration::from_secs(seconds))
}
```

Enet4 (commented):
Thank you for working on this. Please see if you can attend to the concern in the comments or whether you would prefer someone else to take over.

app-common/src/lib.rs
karanrane96: Good catch, fixed in latest commit. Switched `ConnectionOptions::read_timeout/write_timeout` to `Option<Duration>` with a `parse_duration_sec value_parser` as suggested. Also caught `connect_timeout` inechoscu/findscu/storescu had the same `u64` issue — made `parse_duration_sec pub` in app-common and reused it there too, so all four timeout flags share one parser now.

Enet4 (approved):
Looks good. There's room for future work which can be worked on in subsequent PRs by anyone interested:

- add relevant timeout options for movescu
- find a way to group connect-timeout with the other Connection Options
- implement support for time measure suffixes (ms, s, m, h) in timeout options

Thanks! :+1:

PR #794, opened 2026-06-20
Revise dicom-app-common library

Enet4:
Just a few clean-ups and bug fixes here and there.

- Bring support for `PeerCert::Ignore` when setting up a TLS client, via a server certificate verifier that skips verification.
   - A warning is always raised in clients when this option is set, because one should rather add trusted certificates on this end.
- Rename `TLSProtocolVersion` to `TlsProtocolVersion`
- Fix system roots flag option, which had `ArgAction::SetFalse` when it should have been enabled on `--system-roots`
- Rename a few error type variants and split others
   - remove anti-pattern of Io and Rustls error wrappers, replace them with more informative variants
- make error type non-exhaustive
- enabled use of value delimiter in suitable args, so that comma-separated values would actually work
- Adjust documentation

- [storescp] [storescu] adjust log filters: add `dicom_app_common=debug` in verbose mode

### DIFF
--- a/app-common/src/lib.rs
+++ b/app-common/src/lib.rs
@@ -68,12 +68,12 @@ pub struct TlsOptions {
     pub crypto_provider: CryptoProvider,
 
     /// List of cipher suites to use. If not specified, the default cipher suites for the selected crypto provider will be used.
-    #[arg(long, value_name = "cipher1,...", use_value_delimiter(true))]
+    #[arg(long, value_name = "cipher1,...", value_delimiter(','))]
     #[arg(hide(cfg!(not(feature = "tls"))))]
     pub cipher_suites: Option<Vec<String>>,
 
     /// TLS protocol versions to enable
-    #[arg(long, value_enum, value_name = "version,...", use_value_delimiter(true), default_values_t = vec![TlsProtocolVersion::TLS1_2, TlsProtocolVersion::TLS1_3])]
+    #[arg(long, value_enum, value_name = "version,...", value_delimiter(','), default_values_t = vec![TlsProtocolVersion::TLS1_2, TlsProtocolVersion::TLS1_3])]
     #[arg(hide(cfg!(not(feature = "tls"))))]
     pub protocol_versions: Vec<TlsProtocolVersion>,
 
@@ -88,12 +88,12 @@ pub struct TlsOptions {
     pub cert: Option<PathBuf>,
 
     /// Path to additional CA certificates (comma separated) in PEM format to add to the root store
-    #[arg(long, value_name = "/path/to/cert.pem,...", use_value_delimiter(true))]
+    #[arg(long, value_name = "/path/to/cert.pem,...", value_delimiter(','))]
     #[arg(hide(cfg!(not(feature = "tls"))))]
     pub add_certs: Option<Vec<PathBuf>>,
 
     /// Add Certificate Revocation Lists (CRLs) to the server's certificate verifier
-    #[arg(long, value_name = "/path/to/crl.pem,...", use_value_delimiter(true))]
+    #[arg(long, value_name = "/path/to/crl.pem,...", value_delimiter(','))]
     #[arg(hide(cfg!(not(feature = "tls"))))]
     pub add_crls: Option<Vec<PathBuf>>,
 
@@ -164,7 +164,7 @@ pub enum TlsProtocolVersion {
 }
 
 /// Peer certificate handling options
-/// cl
+///
 /// Defines how the TLS connection should handle peer certificates.
 #[derive(clap::ValueEnum, Clone, Debug, Eq, Hash, PartialEq)]
 pub enum PeerCertOption {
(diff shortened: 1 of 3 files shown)


qarmin:
Can be findscu/src/query.rs and movescu/src/query.rs unified in app-common? This file is duplicated in this two places

app-common/src/lib.rs
```diff
@@ -56,14 +68,14 @@ pub struct TlsOptions {
     pub crypto_provider: CryptoProvider,
 
     /// List of cipher suites to use. If not specified, the default cipher suites for the selected crypto provider will be used.
-    #[arg(long, value_name = "cipher1,...")]
+    #[arg(long, value_name = "cipher1,...", use_value_delimiter(true))]
```
naterichman: Looks like `use_value_delimiter` is [deprecated](https://docs.rs/clap_builder/4.6.0/src/clap_builder/builder/arg.rs.html#1646) in favor of `value_delimiter`
```suggestion
    #[arg(long, value_name = "cipher1,...", value_delimiter(','))]
```

app-common/src/lib.rs
```diff
@@ -56,14 +68,14 @@ pub struct TlsOptions {
...
     #[arg(hide(cfg!(not(feature = "tls"))))]
     pub cipher_suites: Option<Vec<String>>,
 
     /// TLS protocol versions to enable
-    #[arg(long, value_enum, value_name = "version,...", default_values_t = vec![TLSProtocolVersion::TLS1_2, TLSProtocolVersion::TLS1_3])]
+    #[arg(long, value_enum, value_name = "version,...", use_value_delimiter(true), default_values_t = vec![TlsProtocolVersion::TLS1_2, TlsProtocolVersion::TLS1_3])]
```
naterichman:
```suggestion
    #[arg(long, value_enum, value_name = "version,...", value_delimiter(','), default_values_t = vec![TlsProtocolVersion::TLS1_2, TlsProtocolVersion::TLS1_3])]
```

app-common/src/lib.rs
```diff
@@ -76,25 +88,24 @@ pub struct TlsOptions {
     pub cert: Option<PathBuf>,
 
     /// Path to additional CA certificates (comma separated) in PEM format to add to the root store
-    #[arg(long, value_name = "/path/to/cert.pem,...")]
+    #[arg(long, value_name = "/path/to/cert.pem,...", use_value_delimiter(true))]
```
naterichman:
```suggestion
    #[arg(long, value_name = "/path/to/cert.pem,...", value_delimiter(','))]
```

app-common/src/lib.rs
```diff
@@ -76,25 +88,24 @@ pub struct TlsOptions {
...
     #[arg(hide(cfg!(not(feature = "tls"))))]
     pub add_certs: Option<Vec<PathBuf>>,
 
     /// Add Certificate Revocation Lists (CRLs) to the server's certificate verifier
-    #[arg(long, value_name = "/path/to/crl.pem,...")]
+    #[arg(long, value_name = "/path/to/crl.pem,...", use_value_delimiter(true))]
```
naterichman:
```suggestion
    #[arg(long, value_name = "/path/to/crl.pem,...", value_delimiter(','))]
```

app-common/src/lib.rs
```diff
@@ -123,34 +135,38 @@ pub struct TlsAcceptorOptions {
...
     TLS1_2,
+    /// TLS v1.3
     TLS1_3,
 }
 
 /// Peer certificate handling options
-/// 
-/// Defines how the TLS connection should handle peer certificates
-#[derive(clap::ValueEnum, Clone, Debug)]
+/// cl
```
naterichman:
```suggestion
```

app-common/src/lib.rs
```diff
@@ -245,17 +264,29 @@ impl TlsOptions{
         let provider = match self.crypto_provider {
             CryptoProvider::AwsLC => rustls::crypto::aws_lc_rs::default_provider(),
         };
+
+        if self.peer_cert == PeerCertOption::Ignore {
```
naterichman: Can this be moved into the block actually constructing the peer cert verifier?

naterichman (changes requested):

naterichman:
> Can be findscu/src/query.rs and movescu/src/query.rs unified in app-common? This file is duplicated in this two places

I don't think these belong in `app-common`, but they should be de-duplicated at some point.  I have draft PR (#681) I need to get back to which adds structs for all the defined dimse messages. That should help de-duplicate somewhat.

After that, I think a few of us have been in conversations around adding a higher level module for dimse (see #781), [issue discussion here](https://github.com/Enet4/dicom-rs/issues/714#issuecomment-3664462480)

Enet4:
Thank you for the feedback!

Regarding the `query` module, I would not be opposed to moving it to dicom-app-common, but there may be an even better place for it. As it turns out, describing small DICOM objects with this syntax is pretty useful, even outside DIMSE! Let us defer to the milestone after 0.10, though I agree that it should have a public API in its own right somewhere.

PR #797, opened 2026-06-22
Fix invalid array declaration in encode_tag

qarmin:
Previously array of 2 elements were created instead of 4 element array


Enet4 (approved):
That's a curious catch. It might be that we aren't using those functions at all, since we have alternatives to write the full element header in one go. Thanks!

PR #798, opened 2026-06-22
Fix DICOM UL AE title field length validation

atlex00:
The A-ASSOCIATE writer encoded the Called/Calling AE titles and then called `resize(16, b' ')`. Invalid values slipped through: empty and all-space titles were written as the forbidden all-spaces field, and titles longer than 16 bytes were silently truncated.

DICOM PS3.8 §9.3.2 defines both as fixed 16-byte fields where the all-spaces value is not allowed. This adds a shared writer helper that pads valid shorter titles to 16 bytes and returns `WriteError::InvalidFixedSizeTextField` for empty, all-space, or overlong titles, used for both A-ASSOCIATE-RQ and A-ASSOCIATE-AC.

### DIFF
--- a/ul/src/pdu/writer.rs
+++ b/ul/src/pdu/writer.rs
@@ -9,6 +9,8 @@ pub type Error = crate::pdu::WriteError;
 
 pub type Result<T> = std::result::Result<T, WriteError>;
 
+const AE_TITLE_FIELD_LENGTH: usize = 16;
+
 #[derive(Debug, Snafu)]
 pub enum WriteChunkError {
     #[snafu(display("Failed to build chunk"))]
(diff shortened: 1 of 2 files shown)


atlex00:
Rebasing

ul/src/pdu/writer.rs
```diff
@@ -66,6 +68,46 @@ where
...
+
+    if bytes.len() > AE_TITLE_FIELD_LENGTH {
+        return InvalidFixedSizeTextFieldSnafu {
+            field,
+            length: AE_TITLE_FIELD_LENGTH,
+            actual_length: bytes.len(),
+            reason: "exceeds fixed field length",
+        }
+        .fail();
+    }
```
Enet4: These two can be rewritten to use the `ensure!` macro. I would also switch the order of the two checks so that the process can skip checking the bytes of the AE title when it is already known to be too large. After formatting:

```suggestion
    ensure!(
        bytes.len() <= AE_TITLE_FIELD_LENGTH,
        InvalidFixedSizeTextFieldSnafu {
            field,
            length: AE_TITLE_FIELD_LENGTH,
            actual_length: bytes.len(),
            reason: "exceeds fixed field length",
        }
    );

    ensure!(
        !validate_value || bytes.iter().any(|&byte| byte != b' '),
        InvalidFixedSizeTextFieldSnafu {
            field,
            length: AE_TITLE_FIELD_LENGTH,
            actual_length: bytes.len(),
            reason: "must contain at least one non-space character",
        }
    );
```

Enet4 (commented):
Thank you for working on this. The writer really ought to ensure that the output stays compliant when faced with bad input such as an invalid AE title.

I have only a small suggestion to implement before we proceed.

ul/src/pdu/writer.rs
atlex00: Hi, thanks for the feedback. Your suggestion is much more readable, so I've gone ahead and applied it. Also, the order about `ensure!` makes sense to me.

Enet4 (approved):
No more concerns here from my side. Thank you!

PR #799, opened 2026-06-22
Validate fixed PDU lengths and encoded item length overflow

atlex00:
`read_pdu` did not enforce the required PDU-length value of 4 for A-ASSOCIATE-RJ, A-RELEASE-RQ, A-RELEASE-RP, and A-ABORT PDUs. Declared lengths below 4 already failed field-length validation, but lengths above 4 could be accepted: the decoder parsed the defined four bytes and discarded the additional bytes included in the declared PDU body. The chunk writers also converted encoded payload lengths with truncating `as u16`/`as u32` casts, so payloads exceeding the corresponding length field could produce a truncated length value while the complete payload was written.

DICOM PS3.8 specifies a PDU-length of `00000004H` for these PDUs (§9.3.4, §9.3.6, §9.3.7, §9.3.8) and defines variable PDU and item lengths as the number of bytes in their following content. This change rejects any other declared length for those fixed PDUs, replaces the length casts with checked conversions that report overflow, and adds the corresponding read and write error variants.

### DIFF
--- a/storescu/src/main.rs
+++ b/storescu/src/main.rs
@@ -53,6 +53,9 @@ struct App {
         value_parser(clap::value_parser!(u32).range(1018..))
     )]
     max_pdu_length: u32,
+    /// Reject trailing bytes in fixed-length PDUs
+    #[arg(long)]
+    reject_trailing_fixed_pdu_bytes: bool,
     /// fail if not all DICOM files can be transferred
     #[arg(long = "fail-first")]
     fail_first: bool,
@@ -187,6 +190,7 @@ pub(crate) fn get_scu_options<'a>(
     calling_ae_title: String,
     called_ae_title: Option<String>,
     max_pdu_length: u32,
+    reject_trailing_fixed_pdu_bytes: bool,
     username: Option<String>,
     password: Option<String>,
     kerberos_service_ticket: Option<String>,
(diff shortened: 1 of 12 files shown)


atlex00:
Rebasing.

storescu/src/main.rs
```diff
@@ -53,6 +53,9 @@ struct App {
         value_parser(clap::value_parser!(u32).range(1018..))
     )]
     max_pdu_length: u32,
+    /// Reject trailing bytes in fixed-length PDUs
+    #[arg(long)]
+    reject_trailing_fixed_pdu_bytes: bool,
```
Enet4, line 58: This option could be shortened a bit. How about this?

```suggestion
    #[arg(long = "no-trailing-fixed-pdu-bytes")]
    reject_trailing_fixed_pdu_bytes: bool,
```

ul/src/association/client.rs
```diff
@@ -1481,12 +1506,7 @@ impl<'a> ClientAssociationOptions<'a> {
                                 error!("Received TLS response to non-TLS request!");
                                 return super::TlsNotSupportedSnafu.fail();
                             }
-                            // if let rustls::Error::InappropriateMessage{..} = err {
-                            //     error!("Recieved TLS response to non-TLS request!");
-                            //     return super::TlsNotSupportedSnafu.fail()
-                            // }
-                            // Recieved a valid TLS message, means the server expects TLS
-                            return super::TlsNotSupportedSnafu.fail();
+                            return Err(e);
```
Enet4, line 1509: I understand that this fixes a false positive in which a client would erroneously claim to be communicating with a TLS server? We can incorporate that, but this either needs to be documented differently in the PR, or split into a separate one.

Enet4 (changes requested):
Thank you for your pull request. Can you look into the two concerns inline?

PR #800, opened 2026-06-23
Added fuzz targets for parser, json, core and root justfile, to simplify running fuzzers

qarmin:


justfile
```diff
@@ -0,0 +1,71 @@
+fuzz crate target timeout="60":
+    cd {{crate}} && cargo +nightly fuzz run {{target}} -- -max_total_time={{timeout}}
```
Enet4: Can we make the toolchain configurable? One might have a specific nightly installed.

Enet4 (commented):
This is nice, and it has uncovered some bugs already. Would you be able to add a few improvements?

A small paragraph on how to run these tests in CONTRIBUTING.md would be nice. Something like this:

> Fuzz tests are available.
> For convenience, they can be run using [just](https://just.systems).
> 
> ```sh
> just fuzz-list
> just fuzz-all
> ```

Also, configurable Rust toolchain (inline comment).

ul/src/association/pdata.rs
```diff
@@ -98,7 +98,7 @@ where
     /// Construct a new P-Data value writer.
     ///
     /// `max_pdu_length` is the maximum value of the PDU-length property.
-    pub(crate) fn new(stream: W, presentation_context_id: u8, max_pdu_length: u32) -> Self {
+    pub fn new(stream: W, presentation_context_id: u8, max_pdu_length: u32) -> Self {
```
Enet4: Well, we're exposing the PData writer to the public API with this, though users are not expected to have direct access to this constructor. `PDataReader::new` is pub already, probably by mistake. I can accept this visibility change and think of an API narrowing for 0.11.0.

CONTRIBUTING.md
```diff
@@ -71,39 +50,43 @@ at the root of the repository will build all crates:
...
 
 ```sh
-cargo fmt
+just fuzz-list
+just fuzz-all
 ```
 
+Corpus directories can be seeded with your own `.dcm` files using `just apply_dcm_corpus <dir>`.
+
+A few targets that need a C/C++ toolchain (GDCM, CharLS), needs to be run seperately with `just fuzz-external`.
```
Enet4:
```suggestion
A few targets that need a C/C++ toolchain (GDCM, CharLS) need to be run separately with `just fuzz-external`.
```

PR #803, opened 2026-06-30
[json] Correctly serialize VR AT

pgimeno4d:
The standard requires the JSON to contain "GGGGEEEE" instead of "(GGGG,EEEE)", per [PS3.18 (2025d) F.2.3](https://dicom.nema.org/medical/dicom/2025d/output/html/part18.html):

"Attributes with a Value Representation of AT shall be restricted to eight character uppercase hexadecimal representation of a DICOM Tag"

### DIFF
--- a/json/src/ser/value.rs
+++ b/json/src/ser/value.rs
@@ -10,7 +10,7 @@ use crate::{INFINITY, NAN, NEG_INFINITY};
 /// which should always be encoded as strings.
 ///
 /// Should be used for the value representations
-/// AE, AS, AT, CS, DA, DT, LO, LT, SH, ST, TM, UC, UI, UR, and UT.
+/// AE, AS, CS, DA, DT, LO, LT, SH, ST, TM, UC, UI, UR, and UT.
 /// Can also be used for the value representations
 /// DS, IS, SV, and UV.
 ///
@@ -211,6 +211,7 @@ impl<'a> From<&'a str> for PersonNameDef<'a> {
 /// Should only be used for the value representation AT.
 ///
 /// [1]: dicom_core::PrimitiveValue
+#[derive(Debug, Clone)]
 pub struct AsAttributeTags<'a>(&'a PrimitiveValue);
 
 impl<'a> From<&'a PrimitiveValue> for AsAttributeTags<'a> {
--- a/json/src/ser/value.rs
+++ b/json/src/ser/value.rs
@@ -2,7 +2,7 @@
 
 use dicom_core::PrimitiveValue;
 use serde::Serialize;
-use serde::ser::SerializeSeq;
+use serde::ser::{Error, SerializeSeq};
 
 use crate::{INFINITY, NAN, NEG_INFINITY};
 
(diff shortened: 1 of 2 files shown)


json/src/ser/value.rs
```diff
@@ -205,10 +205,41 @@ impl<'a> From<&'a str> for PersonNameDef<'a> {
     }
 }
 
+/// Wrapper type for [primitive values][1]
+/// which should be encoded as attribute tags (group,element).
+///
+/// Should only be used for the value representation AT.
+///
+/// [1]: dicom_core::PrimitiveValue
+pub struct AsAttributeTags<'a>(&'a PrimitiveValue);
```
ingwinlu, line 215: Add ` #[derive(Debug, Clone)]`

json/src/ser/value.rs
ingwinlu: Remove AT

json/src/ser/value.rs
```diff
@@ -205,10 +205,41 @@ impl<'a> From<&'a str> for PersonNameDef<'a> {
...
+            ));
+        };
+        serializer.collect_seq(
+            tags.iter()
+                .map(|tag| format!("{:04X}{:04X}", tag.group(), tag.element())),
+        )
+    }
+}
+
 #[cfg(test)]
```
ingwinlu, line 240: could we add a roundtrip test for VR:AT?

ingwinlu (changes requested):

json/src/ser/value.rs
pgimeno4d: There are no round-trip tests at all at the moment. I guess that that'd be best left to lib.rs. Such a task exceeds a bit the scope of the problem that this PR intended to address.

json/src/ser/value.rs
```diff
@@ -205,10 +205,42 @@ impl<'a> From<&'a str> for PersonNameDef<'a> {
...
+        S: serde::Serializer,
+    {
+        let PrimitiveValue::Tags(tags) = self.0 else {
+            return Err(S::Error::custom(
+                "AsAttributeTags created with a non-Tags value",
+            ));
+        };
+        serializer.collect_seq(
+            tags.iter()
+                .map(|tag| format!("{:04X}{:04X}", tag.group(), tag.element())),
```
Enet4, line 235: I was wondering about the allocation, but it appears that implementing `Serialize` from `Display` isn't trivial: https://github.com/serde-rs/serde/issues/1316 The `Serialize` impl for `DicomJson<Tag>` has the same issue.

Let's proceed with this and revise if the circumstances change.

json/src/ser/value.rs
Enet4: A round-trip for a known object may indeed be nice to have at root crate level. Even better would be a round trip through fuzzy testing. One way or another, it won't be a blocker here. :)

Enet4 (approved):
Thank you for spotting this! Problems with AT values can be easy to miss.

PR #804, opened 2026-06-30
[json] Remove padding when serializing IS, DS

pgimeno4d:
When values with a value representation of DS (decimal string) or IS (integer string) come from a DICOM file, if their length is odd, a padding space is appended. This is reflected in their `PrimitiveValue::Strs()` value.

Since the serialization took the raw value contained in `PrimitiveValue::Strs()`, the padding spaces were not eliminated when converting to JSON. However, JSON requires that no such padding is present. Therefore, this PR inserts the "cooked" value instead via `PrimitiveValue.to_multi_str()`.

The unit test did not detect that because the value was created without padding spaces, therefore it has been adjusted to have one, to make this issue apparent.

### DIFF
--- a/json/src/ser/value.rs
+++ b/json/src/ser/value.rs
@@ -66,8 +66,8 @@ impl Serialize for AsNumbers<'_> {
             PrimitiveValue::Time(_) => panic!("wrong impl: cannot encode Time as numbers"),
             PrimitiveValue::Tags(_) => panic!("wrong impl: cannot encode Tags as numbers"),
             // strings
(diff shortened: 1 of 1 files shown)


ingwinlu:
I suggest adding a test for Strs as well to pin the behaviour there too. Rest looks good.

pgimeno4d:
Done. I've switched the test to use `dicom_values!` so that the specific `PrimitiveValue` variant used is explicit.

ingwinlu (approved):

Enet4 (approved):
Great! Thank you both!

PR #805, opened 2026-07-01
Fix de/serialization of PN attributes

pgimeno4d:
Serialization did not take into account PN values with ideographic and/or phonetic fields.

Deserialization did not allow for a missing Alphabetic value, yet it's allowed by the standard. PS3.5 section 6.2, entry PN in table 6.2-1:
'Any component group may be absent, including the first component group. In this case, the person name may start with one or more "=" delimiters.'

### DIFF
--- a/json/src/ser/value.rs
+++ b/json/src/ser/value.rs
@@ -195,13 +195,23 @@ impl Serialize for AsPersonNames<'_> {
 /// Should only used for the value representation PN.
 #[derive(Debug, Clone, Serialize)]
 pub struct PersonNameDef<'a> {
-    #[serde(rename = "Alphabetic")]
+    #[serde(rename = "Alphabetic", skip_serializing_if = "str::is_empty")]
     alphabetic: &'a str,
+    #[serde(rename = "Ideographic", skip_serializing_if = "str::is_empty")]
+    ideographic: &'a str,
+    #[serde(rename = "Phonetic", skip_serializing_if = "str::is_empty")]
+    phonetic: &'a str,
 }
 
 impl<'a> From<&'a str> for PersonNameDef<'a> {
     fn from(value: &'a str) -> Self {
-        PersonNameDef { alphabetic: value }
+        let mut parts = value.split('=');
+
+        PersonNameDef {
+            alphabetic: parts.next().unwrap_or(""),
+            ideographic: parts.next().unwrap_or(""),
+            phonetic: parts.next().unwrap_or(""),
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


json/src/ser/value.rs
```diff
@@ -275,4 +285,53 @@ mod tests {
...
+
+    #[test]
+    fn serialize_names_with_ideographic_and_phonetic() {
+        let v = dicom_value!(
+            Strs,
+            [
+                "House^Gregory^^M.D.",
+                "Wang^XiaoDong=\u{738b}^\u{5c0f}\u{4e1c}=",
+                "Orl\u{e9}ans de Gallia^Charlotte^H\u{e9}l\u{e8}ne==\
+                \u{30aa}\u{30eb}\u{30ec}\u{30a2}\u{30f3}\u{30fb}\u{30c7}\
```
Enet4: Do you anticipate problems if we write these directly in UTF-8 without defining each scalar value with `/u{...}`?

pgimeno4d:
This appears to be a breaking change, since `PersonNameDef` is public.

Enet4 (commented):
> This appears to be a breaking change, since `PersonNameDef` is public.

Uh, not quite! The two `value` modules are not public, so even though this struct has `pub` in declarations, they are not exposed to library consumers at all. :)

json/src/ser/value.rs
pgimeno4d: I'm not very fond of depending on the source file's encoding, as it has caused trouble in other languages, but I'm OK with it if you greenlight it.

pgimeno4d: This commit contains that change:
https://github.com/pgimeno4d/dicom-rs/commit/2c5b1763a7ac389ee9ed135464caef3fd57b77de

My editor supports it acceptably, although since the text has a high amount of double-width characters, the column numbers don't really correspond with the visual line length. Other editors might take the fixed width too literally and use the same width for single- and double-width characters. And some other editors might not recognize the text as UTF-8 and use a single-byte character set configuration.

Enet4: It's true that this could be a footgun in other languages, but thankfully, per the [Rust reference](https://doc.rust-lang.org/reference/input-format.html), source code files are required to be encoded in UTF-8. :)

I would appreciate that commit to be inserted in favor of clarity and readability over compatibility with subpar (or misconfigured) text editors, as we'll also not encounter that many cases of characters outside the ASCII range.

Enet4 (approved):
:+1: Thank you!

Issue #73, opened 2020-09-30
Refactor documentation to use intra-doc links


Enet4:
While intra-doc links are not fully stabilized, they are [close to stabilization](https://blog.rust-lang.org/inside-rust/2020/09/17/stabilizing-intra-doc-links.html), already work in docs.rs, and are easier to write and less error prone than working with HTML links directly. I believe it is time for DICOM-rs to embrace them.

PR #806, opened 2026-07-07
Add intra-doc links to core crate

Luni-4:
This PR addresses part of #73 by introducing intra-doc links to the public API of the `core` crate. 

Additional changes:
- Converted obsolete comment blocks  (/* */) to idiomatic documentation comments (///).
- Refactored various comments to improve clarity and readability.

### DIFF
--- a/core/src/dictionary/data_element.rs
+++ b/core/src/dictionary/data_element.rs
@@ -146,7 +146,7 @@ impl FromStr for TagRange {
 /// It is used by element dictionary entries to describe circumstances
 /// in which the real VR may depend on context.
 /// As an example, the _Pixel Data_ attribute
-/// can have a value representation of either [`OB`](VR::OB) or [`OW`](VR::OW).
+/// can have a value representation of either [`VR::OB`] or [`VR::OW`].
 #[derive(Debug, Copy, Clone, Eq, Hash, PartialEq)]
 #[non_exhaustive]
 pub enum VirtualVr {
@@ -162,19 +162,19 @@ pub enum VirtualVr {
     /// (represented by a _Pixel Representation_ value of `1`),
     /// then values with this virtual VR
     /// should be interpreted as signed 16 bit integers
-    /// ([`SS`](VR::SS)),
+    /// ([`VR::SS`]),
     /// otherwise they should be interpreted as unsigned 16 bit integers
-    /// ([`US`](VR::US)).
+    /// ([`VR::US`]).
     Xs,
     /// Represents overlay data sample values.
     ///
-    /// It can be either [`OB`](VR::OB) or [`OW`](VR::OW).
+    /// It can be either [`VR::OB`] or [`VR::OW`].
     Ox,
     /// Represents pixel data sample value.
     ///
-    /// It can be either [`OB`](VR::OB) or [`OW`](VR::OW).
+    /// It can be either [`VR::OB`] or [`VR::OW`].
     Px,
-    /// Represents LUT data, which can be [`US`](VR::US) or [`OW`](VR::OW)
+    /// Represents LUT data, which can be [`VR::US`] or [`VR::OW`].
     Lt,
 }
 
@@ -198,10 +198,10 @@ impl VirtualVr {
     /// making a relaxed conversion if it cannot be
     /// accurately resolved without context.
     ///
-    /// - [`Xs`](VirtualVr::Xs) is relaxed to [`US`](VR::US)
-    /// - [`Ox`](VirtualVr::Ox) is relaxed to [`OW`](VR::OW)
-    /// - [`Px`](VirtualVr::Px) is relaxed to [`OW`](VR::OW)
-    /// - [`Lt`](VirtualVr::Lt) is relaxed to [`OW`](VR::OW)
+    /// - [`VirtualVr::Xs`] is relaxed to [`VR::US`]
+    /// - [`VirtualVr::Ox`] is relaxed to [`VR::OW`]
+    /// - [`VirtualVr::Px`] is relaxed to [`VR::OW`]
+    /// - [`VirtualVr::Lt`] is relaxed to [`VR::OW`]
     ///
     /// This method is ill-advised for uses where
     /// the corresponding attribute is important.
@@ -238,27 +238,27 @@ enum ParseSelectorErrorInner {
 /// to retrieve a record containing additional information about a data element,
 /// in one of the following ways:
 ///
-/// - By DICOM tag, via [`by_tag`][1];
-/// - By its keyword (also known as alias) via [`by_name`][2];
+/// - By DICOM tag, via [`by_tag`];
+/// - By its keyword (also known as alias) via [`by_name`];
 /// - By an expression which may either be a keyword
 ///   or a tag printed in one of its standard forms,
-///   using [`by_expr`][3].
+///   using [`by_expr`].
 ///
-/// These methods will return `None`
+/// These methods will return [`None`]
 /// when the tag or name is not recognized by the dictionary.
 ///
 /// In addition,
 /// the data element dictionary provides
 /// built-in DICOM tag and selector (path) parsers for convenience.
-/// [`parse_tag`][4] converts an arbitrary expression to a tag,
-/// whereas [`parse_selector`][5] produces an [attribute selector][6].
+/// [`parse_tag`] converts an arbitrary expression to a tag,
+/// whereas [`parse_selector`] produces an [AttributeSelector].
 ///
-/// [1]: DataDictionary::by_tag
-/// [2]: DataDictionary::by_name
-/// [3]: DataDictionary::by_expr
-/// [4]: DataDictionary::parse_tag
-/// [5]: DataDictionary::parse_selector
-/// [6]: crate::ops::AttributeSelector
+/// [`by_tag`]: Self::by_tag
+/// [`by_name`]: Self::by_name
+/// [`by_expr`]: Self::by_expr
+/// [`parse_tag`]: Self::parse_tag
+/// [`parse_selector`]: Self::parse_selector
+/// [`AttributeSelector`]: crate::ops::AttributeSelector
 pub trait DataDictionary {
     /// The type of the dictionary entry.
     type Entry: DataDictionaryEntry;
@@ -277,14 +277,15 @@ pub trait DataDictionary {
     /// slightly more expensive than by DICOM tag.
     /// If the parameter provided is a string literal
     /// (e.g. `"StudyInstanceUID"`),
-    /// then it may be better to use [`by_tag`][1]
+    /// then it may be better to use [`by_tag`]
     /// with a known tag constant
-    /// (such as [`tags::STUDY_INSTANCE_UID`][2]
-    /// from the [`dicom-dictionary-std`][3] crate).
+    /// (such as [`tags::STUDY_INSTANCE_UID`]
+    /// from the [`dicom-dictionary-std`] crate).
     ///
-    /// [1]: DataDictionary::by_tag
-    /// [2]: https://docs.rs/dicom-dictionary-std/0.5.0/dicom_dictionary_std/tags/constant.STUDY_INSTANCE_UID.html
-    /// [3]: https://docs.rs/dicom-dictionary-std/0.5.0
+    ///
+    /// [`by_tag`]: Self::by_tag
+    /// [`tags::STUDY_INSTANCE_UID`]: https://docs.rs/dicom-dictionary-std/latest/dicom_dictionary_std/tags/constant.STUDY_INSTANCE_UID.html
+    /// [`dicom-dictionary-std`]: https://docs.rs/dicom-dictionary-std/latest/dicom_dictionary_std/
     fn by_name(&self, name: &str) -> Option<&Self::Entry>;
 
     /// Fetch an entry by its alias or by DICOM tag expression.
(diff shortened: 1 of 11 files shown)


core/src/dictionary/data_element.rs
```diff
@@ -277,12 +271,11 @@ pub trait DataDictionary {
     /// slightly more expensive than by DICOM tag.
     /// If the parameter provided is a string literal
     /// (e.g. `"StudyInstanceUID"`),
-    /// then it may be better to use [`by_tag`][1]
+    /// then it may be better to use [`DataDictionary`]
```
Enet4: This changes the intended meaning. `by_tag` is the method referred to here.

core/src/dictionary/mod.rs
```diff
@@ -1,6 +1,7 @@
 //! This module contains the concept of a DICOM data dictionary.
 //!
-//! The standard data dictionary is available in the [`dicom-dictionary-std`] crate.
+//! The standard data dictionary is available in the
+//! [dicom-dictionary-std](https://github.com/Enet4/dicom-rs/tree/master/dictionary-std).
```
Enet4: Mentions to non-dependent crates are usually made via docs.rs.


```suggestion
//! [dicom-dictionary-std](https://docs.rs/dicom-dictionary-std).
```

core/src/value/range.rs
```diff
@@ -174,11 +175,11 @@ pub trait AsRange {
 
     /// Returns the earliest possible value from a partial precision structure.
     /// Missing components default to 1 (days, months) or 0 (hours, minutes, ...)
-    /// If structure contains invalid combination of `DateComponent`s, it fails.
+    /// If structure contains invalid combination of [`crate::value::partial::DateComponent`]s, it fails.
```
Enet4: This is one of the various places where the fully qualified name is inelegant.


```suggestion
    /// If the structure contains an invalid combination of [`DateComponent`]s, it fails.
    ///
    /// [`DateComponent`]: crate::value::partial::DateComponent
```

core/src/ops.rs
```diff
@@ -431,17 +422,16 @@ impl<'a> IntoIterator for &'a AttributeSelector {
...
+    /// [`AttributeSelectorStep::Nested`],
+    /// and terminated by one item guaranteed to be a
+    /// [`AttributeSelectorStep::Tag`].
     fn into_iter(self) -> Self::IntoIter {
         self.0.iter()
     }
 }
 
-/// Creates an attribute selector for just a [`tag`](AttributeSelectorStep::Tag).
+/// Creates an attribute selector for just a [`AttributeSelectorStep::Tag`].
```
Enet4: This is another place where the conceptual meaning is diffused. The text speaks of a DICOM tag, and the link is just additional context. I don't think that there is much value in replacing this one.

core/src/ops.rs
```diff
@@ -199,7 +198,7 @@ impl std::fmt::Display for AttributeSelectorStep {
 ///
 /// In most cases, you might only wish to select an attribute
 /// that is sitting at the root of the data set.
-/// This can be done by converting a [DICOM tag](crate::Tag) via [`From<Tag>`]:
+/// This can be done by converting a [`crate::Tag`] via [`From<Tag>`]:
```
Enet4: Intra-doc links do not have to be transparent all the time. In here, the subsequent `From<Tag>` already clarifies that the input is `crate::Tag`.

core/src/dictionary/data_element.rs
```diff
@@ -335,21 +328,19 @@ pub trait DataDictionary {
         })
     }
 
-    /// Parse a string as an [attribute selector][1].
+    /// Parse a string as an [crate::ops::attribute selector].
```
Enet4, line 339: This one is broken. We might as well keep it how it was before.

core/src/dictionary/data_element.rs
```diff
@@ -358,7 +349,7 @@ pub trait DataDictionary {
     ///   _Patient Age_
     /// - `0040A168[0].CodeValue`:
     ///   _Code Value_ in first item of _Concept Code Sequence_
-    /// - `SequenceOfUltrasoundRegions.RegionSpatialFormat`:
+    /// - `[`SequenceOfUltrasoundRegions.RegionSpatialFormat`]`:
```
Enet4: This line represents an example of an attribute selector, not a Rust construct.

core/src/header.rs
```diff
@@ -909,7 +897,7 @@ impl VR {
         }
     }
 
-    /// Retrieve a copy of this VR's byte representation.
+    /// Retrieve a copy of this [`VR`]'s byte representation.
```
Enet4, line 906: This method is already for `VR`, so there is no point in bringing a reference back to the same type.


```suggestion
    /// Retrieve a copy of this VR's byte representation.
```

core/src/dictionary/data_element.rs
```diff
@@ -238,27 +238,21 @@ enum ParseSelectorErrorInner {
...
-///
-/// [1]: DataDictionary::by_tag
-/// [2]: DataDictionary::by_name
-/// [3]: DataDictionary::by_expr
-/// [4]: DataDictionary::parse_tag
-/// [5]: DataDictionary::parse_selector
-/// [6]: crate::ops::AttributeSelector
+/// [`DataDictionary::parse_tag`] converts an arbitrary expression to a tag,
+/// whereas [`DataDictionary::parse_selector`] produces
+/// an [`crate::ops::AttributeSelector`].
```
Enet4: Let's avoid fully qualifying types in documentation like this, as it is inelegant. For what it's worth, this doc section was already using intra-doc comments, so no changes were necessary.

core/src/value/mod.rs
```diff
@@ -122,11 +122,11 @@ impl<I> Value<I> {
     /// from a list of items and length.
     ///
     /// This function will define the pixel data fragment type parameter `P`
-    /// to the `Value` type's default ([`InMemFragment`]),
+    /// to the [`Value`] type's default ([`InMemFragment`]),
```
Enet4, line 125: Value is already the type for which we are implementing this function. Creating intra-doc links to Self can mislead users into thinking that it's a different type.

core/src/value/partial.rs
```diff
@@ -609,10 +608,8 @@ impl DicomTime {
         }
     }
 
-    /**
-     * Constructs a new `DicomTime` from an hour, minute, second, second fraction
-     * and fraction precision value (1-6). Function used for parsing only.
-     */
+    // Constructs a new [`DicomTime`] from an hour, minute, second, second fraction
+    // and fraction precision value (1-6). Function used for parsing only.
```
Enet4: A correct conversion would use three forward slashes instead of just two.

```suggestion
    /// Constructs a new [`DicomTime`] from an hour, minute, second, second fraction
    /// and fraction precision value (1-6). Function used for parsing only.
```

By the way, doc-comment blocks of the form `/** */` are still pretty much supported and not disallowed by this project, though indeed there have been [regressions](https://github.com/rust-lang/rust/issues/92872) associated with its use in the past.

core/src/value/partial.rs
```diff
@@ -1104,12 +1093,12 @@ impl PreciseDateTime {
     }
 }
 
-/// The partial ordering for `PreciseDateTime`
+/// The partial ordering for [`PreciseDateTime`]
 /// is defined by the partial ordering of matching variants
-/// (`Naive` with `Naive`, `TimeZone` with `TimeZone`).
+/// ([`Self::Naive`] with [`Self::Naive`], [`Self::TimeZone`] with [`Self::TimeZone`]).
```
Enet4: Likewise, this would be more readable without the constant `Self::` propping up.

```suggestion
/// ([`Naive`][1] with [`Naive`][1], [`TimeZone`][2] with [`TimeZone`][2]).
///
/// [1]: Self::Naive
/// [2]: Self::TimeZone
```

Enet4 (changes requested):
Thank you for looking into this. There were indeed a few places left with the old intra-reference syntax and other places in which  linking is beneficial, but I can also see a bunch of places where this was done incorrectly or pretty much overdone. :) The inline comments are not exhaustive, but please take a look at them to understand how to reiterate from here.

core/src/dictionary/mod.rs
Luni-4: I have a general concern about this change. When a new version is released, the link will always point to the latest version of `dicom-dictionary-std`, rather than the version that was actually used. This could lead to discrepancies.

One simple option would be to add a note in the comment clarifying that the link always points to the latest version. That's the simplest solution I could think of. What do you think?

core/src/dictionary/data_element.rs
```diff
@@ -277,12 +271,11 @@ pub trait DataDictionary {
...
     /// (e.g. `"StudyInstanceUID"`),
-    /// then it may be better to use [`by_tag`][1]
+    /// then it may be better to use [`DataDictionary`]
     /// with a known tag constant
     /// (such as [`tags::STUDY_INSTANCE_UID`][2]
     /// from the [`dicom-dictionary-std`][3] crate).
     ///
-    /// [1]: DataDictionary::by_tag
     /// [2]: https://docs.rs/dicom-dictionary-std/0.5.0/dicom_dictionary_std/tags/constant.STUDY_INSTANCE_UID.html
     /// [3]: https://docs.rs/dicom-dictionary-std/0.5.0
```
Luni-4: A [similar](https://github.com/Enet4/dicom-rs/pull/806#discussion_r3557046517) problem happens here, but in this case there are fixed tags

core/src/dictionary/data_element.rs
Luni-4: We can simply use `AttributeSelector` here, my fault

core/src/ops.rs
Luni-4: I don't fully understand the difference here. If we keep the previous version, the name `tag` is displayed, but it always links to `AttributeSelectorStep::Tag`. In the new version, the reference is simply made explicit.

Is the goal to remove the intra-doc for grammatical reasons? Personally, I think being explicit makes the context clearer at a glance, but I'm happy to revert the change if it's considered a negative choice.

core/src/dictionary/mod.rs
Enet4: We can point to specific versions in docs.rs, the main problem is the extra burden of having to update these links on each new release. So yes, a small note that it points at the latest version would be fine here.

core/src/value/mod.rs
Luni-4: Sometimes, when you're looking at the methods of a struct, it's useful to have a direct link in the documentation that points back to the main struct. That's why I changed it to an intra-doc link.

That said, I'm happy to revert the change if you think it's unnecessary. Let me know.

core/src/dictionary/data_element.rs
Enet4: Indeed, it is much too easy to forget to update these links when they are fixed to specific versions.

core/src/value/partial.rs
Luni-4: In this case, I used a regular comment (`//`) because the function is `pub(crate)`. The same applies to the function on the line below.

Luni-4:
Thanks a lot, @Enet4, for your deep review!

I think I've addressed all of your review comments:

* Updated the links to point to the _latest_ documentation version. I also updated some DICOM specification links so that they point to the _current_ version. This way, the documentation won't need to be updated whenever a new release is published, as it will always point to the most up-to-date information.
* Replaced doc comments with regular comments (`//`) for `pub(crate)` functions and within some tests. If you are interested in documenting private items with `cargo doc --document-private-items` I can revert these changes.
* Used a cleaner and more readable notation for intra-doc links by displaying only the item name instead of `Self::Something`.

I haven't addressed this comment (https://github.com/Enet4/dicom-rs/pull/806#discussion_r3554949385) and this one (https://github.com/Enet4/dicom-rs/pull/806#discussion_r3554919168), because I think it's useful to provide a link back to the main struct from its associated methods. I've seen this pattern used in other crates as well. Personally, I find it convenient to have a direct link back to the main struct. However, if you think these intra-doc links are unnecessary, just let me know and I'll be happy to remove them.

At the end of the review process, I will squash all commits into a single one

core/src/dictionary/data_element.rs
```diff
@@ -162,19 +162,19 @@ pub enum VirtualVr {
...
     Xs,
     /// Represents overlay data sample values.
     ///
-    /// It can be either [`OB`](VR::OB) or [`OW`](VR::OW).
+    /// It can be either [`VR::OB`] or [`VR::OW`].
     Ox,
     /// Represents pixel data sample value.
     ///
-    /// It can be either [`OB`](VR::OB) or [`OW`](VR::OW).
+    /// It can be either [`VR::OB`] or [`VR::OW`]).
```
Enet4:
```suggestion
    /// It can be either [`VR::OB`] or [`VR::OW`].
```

Enet4 (commented):
The reported errors are unrelated and should be resolved after a rebase. Did some semver checks and found no breakages. Just one more tiny fix and we're ready to merge!

---

It would be fine to keep comments in `pub(crate)` elements as doc-comments (if we ever consider making them part of the public API, it would mean less changes to make), but I won't make it a blocker.

Luni-4:
> It would be fine to keep comments in `pub(crate)` elements as doc-comments (if we ever consider making them part of the public API, it would mean less changes to make), but I won't make it a blocker.

You're right, reverted back my changes. I also squashed all changes into a single commit

Enet4 (approved):
Great! Thank you for following up on this, and welcome to the list of DICOM-rs contributors. :+1:

Luni-4:
@Enet4 

Thanks to you for the review!

PR #808, opened 2026-07-10
[app-common] Add documentation for missing pem object

Luni-4:
This PR documents `MissigPemObject` structure

### DIFF
--- a/app-common/src/lib.rs
+++ b/app-common/src/lib.rs
@@ -12,10 +12,16 @@ use std::sync::Arc;
 #[cfg(feature = "tls")]
 use tracing::debug;
 
+/// A required PEM object is missing from the provided data.
+///
+/// Indicates that a PEM file is expected to contain both an X.509 certificate
+/// and a private key, but one or more of the required objects are missing.
 #[derive(Snafu, Debug)]
 pub enum MissingPemObject {
+    /// No certificate found in the PEM data.
     #[snafu(display("Missing Certificate"))]
     Certificate,
+    /// No private key found in the PEM data.
     #[snafu(display("Missing Private Key"))]
     PrivateKey,
 }


Enet4:
So, I was a bit hesitant last time I looked at this, because it was surprising to find `MissingPemObject` as an error type that is not _exactly_ used like an error. Semantically, it represents which kind of PEM object is missing. The corresponding error variant which uses it is `TlsError::Config`.

Changing this to better reflect the semantics of the type would be a breaking change though, so we might as well admit this as a documentation and error display improvement (it still is, because the error delegates to this new Display impl), and maybe refactor these types later in v0.11.0.

Luni-4:
Yeah, `MissingPemObject` is used as a description of the `TlsError::Config` variant, so by transitivity it should be treated as an error. However, we can view it more as **metadata** associated to the error.

From a documentation perspective, we might change the description in this way:

```rust
/// A required PEM object is missing from the provided data. 
/// 
/// Indicates that a PEM file is expected to contain both an X.509 certificate and a private key, 
/// but one or more of the required objects are missing.
```

What do you think?

Enet4:
> Yeah, `MissingPemObject` is used as a description of the `TlsError::Config` variant, so by transitivity it should be treated as an error. However, we can view it more as **metadata** associated to the error.

I mean, an error's field can be used in its display impl without it being an error. In this case it so happens that `MissingPemObject` has a display impl which amounts to a suitable error message.

What further adds to the discrepancy is that when we have a field in an error type which is _also_ an error, it is usually a `source`! And in this case it isn't. :)

> 
> From a documentation perspective, we might change the description in this way:
> 
> ```rust
> /// A required PEM object is missing from the provided data. 
> /// 
> /// Indicates that a PEM file is expected to contain both an X.509 certificate and a private key, 
> /// but one or more of the required objects are missing.
> ```
> 
> What do you think?

For the time being, I think one doc-block or the other is fine, and admittedly better than what we have upstream. Feel free to do update the PR or signal to proceed.

Luni-4:
> I mean, an error's field can be used in its display impl without it being an error. In this case it so happens that `MissingPemObject` has a display impl which amounts to a suitable error message.
> 
> What further adds to the discrepancy is that when we have a field in an error type which is _also_ an error, it is usually a `source`! And in this case it isn't. :)

Thank you for the clarification, now I can see the discrepancy more clearly. I agree that we should refactor these types in v0.11.0, where we can introduce breaking changes freely.

> For the time being, I think one doc-block or the other is fine, and admittedly better than what we have upstream. Feel free to do update the PR or signal to proceed.

Yep, while waiting for the breaking change, we can maintain this new documentation. I'll update this PR.

Enet4 (approved):
:+1: Thanks.

PR #809, opened 2026-07-10
Make module functions private in binary crates

Luni-4:
Since this issue is still open (https://github.com/rust-lang/rust/issues/74970), it's better to change all `pub` modifiers in binary crates to `pub(crate)`. This change also helped uncover an unused enum variant.


Enet4 (approved):
Looks OK. :+1: We may well try to follow that approach for binary crates from now on. Appreciated!

PR #820, opened 2026-08-06
Support overlays

### LINKED ISSUE #818: Add option to merge overlay into pixeldata
Some DICOM files include overlay pixeldata (stored in `(60xx,3000)`). Currently that pixeldata is completely ignored when using `decode_pixel_data_frame`. 
The desirable behaviour is that we can "burn" overlays into the extracted image. The color & transparency that is used for rendering the overlay should/must be configurable. Maybe it's easier to just extract the pixeldata as a separate image and leave the rendering to the API consumer.

Below are some samples, using our own viewer (mRay):

[testfiles.zip](https://github.com/user-attachments/files/30743567/testfiles.zip)

## `1.3.12.2.1107.5.1.4.1005.30000012052906451134300000405` (Patient protocol, all pixeldata as overlay):
No overlay:
Image
Overlay:
Image

## `1.3.12.2.1107.5.1.4.65094.30000025032406503517900000761`(Measurement, overlay to pixeldata):
No overlay:
Image
Overlay:
Image

For reference our option to control the visualization of the overlays looks like this:
Image

feliwir:
This PR adds 2 new functions to the `PixelDecoder` trait:
- `decode_overlays`: This decodes all OverlayPlanes present in the file as vec
- `decode_overlay`: This decodes a single OverlayPlane by it's index

There's currently one default implementation in `overlay.rs`, which is used by all PixelDecoder implementations
Additionally i added support to the `toimage` crate, with the following options:
- `overlays`: A toggle to enable overlay burnins (defaults to false)
- `overlay-color`: A repeating argument which can be used to change the default overlay color (default is white)
- `overlay-file`: If overlays stored in a separate file (presentation state) these can be passed here

This closes #818

### DIFF
--- a/pixeldata/src/attribute.rs
+++ b/pixeldata/src/attribute.rs
@@ -33,6 +33,15 @@ pub enum AttributeName {
     LutDescriptor,
     LutData,
     LutExplanation,
+    OverlayRows,
+    OverlayColumns,
+    OverlayType,
+    OverlayOrigin,
+    OverlayBitsAllocated,
+    OverlayBitPosition,
+    OverlayData,
+    NumberOfFramesInOverlay,
+    ImageFrameOrigin,
 }
 
 impl std::fmt::Display for AttributeName {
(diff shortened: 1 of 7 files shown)


pixeldata/src/lib.rs
```diff
@@ -283,6 +285,21 @@ enum InnerError {
         nr_frames: u32,
         backtrace: Backtrace,
     },
+    #[snafu(display("Unsupported overlay plane configuration in group {group:#06X}: {reason}"))]
+    UnsupportedOverlay {
+        group: u16,
+        reason: String,
```
Enet4: Looks like an anti-pattern: the field `reason` here could stand as an error cause in its own right.

It is better to create an `OverlayError` possibly defined at the `overlay` module. Then this variant would contain a source error type and we'd no longer need any stringly errors.

pixeldata/src/attribute.rs
```diff
@@ -769,6 +778,196 @@ pub fn voi_lut_sequence<D: DataDictionary + Clone>(
...
+) -> Result<u16> {
+    retrieve_required_u16(obj, Tag(group, 0x0102), AttributeName::OverlayBitPosition)
+}
+
+/// Get the Number of Frames in Overlay (60xx,0015) of the overlay plane
+/// in the given group,
+/// returning 1 if it is not present
+pub fn number_of_frames_in_overlay<D: DataDictionary + Clone>(
+    obj: &FileDicomObject<InMemDicomObject<D>>,
+    group: u16,
```
Enet4, line 861: I started wondering how pedantic we should be about checking the `group` input, but since this module isn't private, I think we do not need hard restrictions. These checks even exist in the `overlay` module. If anything, there could be room for a `debug_assert!` at this level, if you wish.

Enet4 (commented):
Thank you for working on this, it looks solid and extensive! I wasn't even expecting support for legacy overlay planes embedded in the pixel data, but we might as well keep it. 😅

I would only like for the overlay error reason field to be refactored (more details inline) before we get this upstream.

pixeldata/src/lib.rs
feliwir: I've created a dedicated OverlayError type as you suggested :+1:

pixeldata/src/attribute.rs
feliwir: I'd leave this untouched if that's alright for you

Enet4 (approved):
OK, looks good! Thank you! :+1:

PR #824, opened 2026-08-24
[json] Restrict binary value deserialization to compatible VRs

### LINKED ISSUE #813: Fuzzer: Non stable json reading/writing of InlineBinary values
Code
```rust
use dicom_object::InMemDicomObject;
use std::process::ExitCode;

static DS_INLINE_BINARY: &str = include_str!("../inputs/ds-inline-binary.json");

fn run(s: &str) -> Result<(), String> {
    let Ok(obj) = dicom_json::from_str::<InMemDicomObject>(s) else {
        return Ok(());
    };
    let json = dicom_json::to_string(&obj).map_err(|e| e.to_string())?;

    let Ok(obj2) = dicom_json::from_str::<InMemDicomObject>(&json) else {
        return Ok(());
    };
    let json2 = dicom_json::to_string(&obj2).map_err(|e| e.to_string())?;

    if json != json2 {
        return Err(format!(
            "second JSON serialization is not identical to the first:\n  {json}\n  {json2}"
        ));
    }
    Ok(())
}

fn main() -> ExitCode {
    match run(DS_INLINE_BINARY) {
        Ok(()) => {
            println!("REPRO-OK");
            ExitCode::SUCCESS
        }
        Err(e) => {
            println!("REPRO-BROKEN: {e}");
            ExitCode::FAILURE
        }
    }
}
```
causes
```none
REPRO-BROKEN: second JSON serialization is not identical to the first:
  {"00020001":{"vr":"DS","Value":[0,1]}}
  {"00020001":{"vr":"DS","Value":["0","1"]}}
```
for
```json
{"00020001":{"vr":"DS","InlineBinary":"AAE="}}
```
Problably Inline Binary should be ignore for Vrs different than `VR::OB, VR::OD, VR::OF, VR::OL, VR::OV, VR::OW, VR::UN`

[ds-inline-binary.json](https://github.com/user-attachments/files/30167232/ds-inline-binary.json)

Enet4:
Should resolve #813.

- reject InlineBinary with a VR that is not on the list of enclosing Attribute VRs in PS3.19 A.1.5
- expand final match clauses in `DataElementVisitor::vist_map` for clarity
- fix check for conflicting Value and InlineBinary fields
- extend test coverage

### DIFF
--- a/json/src/de/mod.rs
+++ b/json/src/de/mod.rs
@@ -186,7 +186,7 @@ where
                     inline_binary = Some(val);
                 }
                 "BulkDataURI" => {
-                    if values.is_some() {
+                    if value.is_some() {
                         return Err(A::Error::custom("\"BulkDataURI\" conflicts with \"Value\""));
                     }
 
@@ -638,6 +638,18 @@ mod tests {
                 "vr": "OW",
                 "InlineBinary": "AAE=",
                 "BulkDataURI": "http://localhost:8042/dicom-web/instances/1.2.3.4.5/pixeldata"
+            }
+        });
+
+        let res: Result<InMemDicomObject, _> = super::from_value(serialized);
+        let e = res.unwrap_err();
+        assert_eq!(e.classify(), serde_json::error::Category::Data);
+
+        let serialized = serde_json::json!({
+            "00081010": {
+                "vr": "SH",
+                "Value": [ "RUSTATION" ],
+                "BulkDataURI": "http://localhost:8042/dicom-web/instances/1.2.3.4.5/station_name"
             }
         });
 
--- a/json/src/de/mod.rs
+++ b/json/src/de/mod.rs
@@ -170,7 +170,7 @@ where
                     value = Some(map.next_value()?);
                 }
                 "InlineBinary" => {
-                    if values.is_some() {
+                    if value.is_some() {
                         return Err(A::Error::custom(
                             "\"InlineBinary\" conflicts with \"Value\"",
                         ));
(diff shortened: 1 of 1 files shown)


json/src/de/mod.rs
```diff
@@ -170,7 +170,7 @@ where
                     value = Some(map.next_value()?);
                 }
                 "InlineBinary" => {
-                    if values.is_some() {
+                    if value.is_some() {
```
userepo, line 173: @Enet4, while you're in there, it might be worth a look at the `"BulkDataURI"` arm a few lines below, since it has the same shape. `values` is still `None` at that point too, so the check never fires...

json/src/de/mod.rs
```diff
@@ -594,4 +604,45 @@ mod tests {
...
+                "vr": "OW",
+                "InlineBinary": "AAE=",
+                "BulkDataURI": "http://localhost:8042/dicom-web/instances/1.2.3.4.5/pixeldata"
+            }
+        });
+
+        let res: Result<InMemDicomObject, _> = super::from_value(serialized);
+        let e = res.unwrap_err();
+        assert_eq!(e.classify(), serde_json::error::Category::Data);
+    }
```
Enet4, line 659:
```suggestion

        let serialized = serde_json::json!({
            "00081010": {
                "vr": "SH",
                "Value": [ "RUSTATION" ],
                "BulkDataURI": "http://localhost:8042/dicom-web/instances/1.2.3.4.5/station_name"
            }
        });

        let res: Result<InMemDicomObject, _> = super::from_value(serialized);
        let e = res.unwrap_err();
        assert_eq!(e.classify(), serde_json::error::Category::Data);
    }
```

json/src/de/mod.rs
Enet4:
```suggestion
                    if value.is_some() {
```

json/src/de/mod.rs
Enet4: Thank you for your review, @userepo! That is a good catch. Updating right away.

PR #826, opened 2026-08-27
[pixeldata] Fix rendering for files with a window width smaller than 1

abustany:
On a file with the following tags:
```
(0028,1050) DS [0.4565323565]                           #  12, 1 WindowCenter
(0028,1051) DS [0.9130647131]                           #  12, 1 WindowWidth
(0028,1052) DS [-0.0002442002587]                       #  16, 1 RescaleIntercept
(0028,1053) DS [0.0002442002587]                        #  16, 1 RescaleSlope
```

Before:
out

After:
out2

### DIFF
--- a/pixeldata/src/transform.rs
+++ b/pixeldata/src/transform.rs
@@ -83,20 +83,38 @@ pub struct WindowLevelTransform {
 impl WindowLevelTransform {
     /// Create a new window level transformation.
     ///
-    /// The width of the given `window_level` is automatically clamped
+    /// The width of the given `window_level` is automatically adjusted
     /// if it is incompatible with the given LUT function:
-    /// it muse be `>= 0` if the function is [`LinearExact`](VoiLutFunction::LinearExact),
+    /// it must be `>= 0` if the function is [`LinearExact`](VoiLutFunction::LinearExact),
     /// and `>= 1` in other functions.
+    ///
+    /// The `LINEAR` function is defined in terms of `width - 1`
+    /// (see [PS3.3 C.11.2.1.2.1][1]),
+    /// so it degenerates into a hard black/white threshold
+    /// once the width is clamped up to 1.
+    /// Objects which describe a window narrower than one unit
+    /// (typically because the modality rescale maps the samples
+    /// onto a small floating point range)
+    /// are therefore rendered with the `LINEAR_EXACT` function instead,
+    /// which is well defined for any non-negative width.
+    ///
+    /// [1]: https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.2.html#sect_C.11.2.1.2.1
     #[inline]
     pub fn new(voi_lut_function: VoiLutFunction, window_level: WindowLevel) -> Self {
+        let (voi_lut_function, width) = match voi_lut_function {
+            VoiLutFunction::LinearExact => (voi_lut_function, window_level.width.max(0.)),
+            VoiLutFunction::Linear if window_level.width < 1. => {
+                (VoiLutFunction::LinearExact, window_level.width.max(0.))
+            }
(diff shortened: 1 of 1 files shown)


Enet4:
Thank you for the pull request. I was originally assuming this to be a bug, but with a deeper inspection, I understand that this makes an adjustment to the VOI LUT function from LINEAR to LINEAR_EXACT when the width is incompatible with the default LINEAR VOI LUT function, is this the case? At this point we may be dealing with a compatibility concern rather than a bug, since we're dealing with non-compliant inputs. Is there a precedent for this behavior in other tools?

abustany:
The [standard](https://dicom.nema.org/medical/dicom/current/output/html/part03.html#sect_C.11.2.1.2.1) says:

>  Window Width (0028,1051) shall always be greater than or equal to 1.

so practically, this file is not compliant...

The two other viewers I know of are:

- Cornerstone3D, which has the same bug (I'm carrying a downstream patch I need to upstream at some point)
- Horos, which doesn't care about LINEAR and always does LINEAR_EXACT

Enet4:
Thank you for the clarification. This is not at all a blocker, though I find it important to recognize when a non-conformity with DICOM exists and from which side it occurs.

abustany:
Yup fully agree! We recently had to ingest a large amount of older files, and bumped into various of those "interesting" cases...

pixeldata/src/transform.rs
```diff
@@ -83,29 +83,47 @@ pub struct WindowLevelTransform {
...
+    ///
+    /// [1]: https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.2.html#sect_C.11.2.1.2.1
     #[inline]
     pub fn new(voi_lut_function: VoiLutFunction, window_level: WindowLevel) -> Self {
+        let (voi_lut_function, width) = match voi_lut_function {
+            VoiLutFunction::LinearExact => (voi_lut_function, window_level.width.max(0.)),
+            VoiLutFunction::Linear if window_level.width < 1. => {
+                (VoiLutFunction::LinearExact, window_level.width.max(0.))
+            }
+            VoiLutFunction::Linear => (voi_lut_function, window_level.width),
```
userepo, line 109: @abustany, I think this arm introduces a regression: a NaN window width used to be harmless and now breaks rendering.

On master every width went through `.max(1.)`, and `f64::max` returns the non-NaN operand, so NaN silently became `1.0`. With this change `NaN < 1.` is `false`, the NaN falls through to this catch-all arm, and it's stored as-is.

From there:

- debug builds panic on `debug_assert!(ww >= 1.)` in `window_level_linear`
- release builds fail LUT construction with `CreateLutError`

So a file that renders today would stop rendering. 
```rust
VoiLutFunction::Linear => (voi_lut_function, window_level.width.max(1.)),
```

abustany: hmm that's a good catch, thanks for the review. I must admit though that I don't know how we'd end up with a NaN value in there?

If I'm not mistaken, restoring `.max(1.)` at line 109 in the hunk above would preserve the old behavior?

Enet4: NaNs could occur from the ingestion of invalid input, either programmatically through the library core+object API or by loading a file with illegal values for Window Width (the VR DS requires numbers to be finite, but they are not inspected during DICOM parsing). On the latter case I have plans to reduce the risk of introducing NaNs at this level by adding methods to retrieve _finite_ floating point values, so we get a recoverable error instead of a NaN.

Enet4: With #831, a window width of NaN from a DICOM object would error before it reaches this spot. However, arbitrary public API usage of `WindowLevelTransform` could bring NaNs, so we might as well keep the safeguard since we recognized it. Afterwards it should be ready to merge.

PR #830, opened 2026-09-10
[transfer-syntax-registry] Set Photometric Interpretation to YBR_FULL_422 when encoding JPEG Baseline

### LINKED ISSUE #829: JPEG Baseline encoder labels its YCbCr output PhotometricInterpretation = RGB
**Version:** dicom-rs 0.10.0; the same code is on `master`
([`transfer-syntax-registry/src/adapters/jpeg.rs`](https://github.com/Enet4/dicom-rs/blob/master/transfer-syntax-registry/src/adapters/jpeg.rs)).

## What happens

Encoding a 3-sample-per-pixel object to JPEG Baseline (1.2.840.10008.1.2.4.50) produces a JFIF stream whose components are YCbCr, but the encoder sets (0028,0004) Photometric Interpretation to `RGB`:

```rust
} else if samples_per_pixel == 3 {
    // set Photometric Interpretation to RGB
    // if it was not already set to RGB
    if pmi != Some("RGB") {
        changes.push(AttributeOp::new(
            Tag(0x0028, 0x0004),
            AttributeAction::SetStr("RGB".into()),
        ));
    }
}
```

The frame is handed to `jpeg-encoder` as `ColorType::Rgb`, which converts to YCbCr and writes a JFIF header:

```rust
let color_type = match samples_per_pixel {
    1 => ColorType::Luma,
    3 => ColorType::Rgb,
    _ => whatever!("Unsupported samples per pixel: {}", samples_per_pixel),
};
let mut encoder = jpeg_encoder::Encoder::new(&mut *dst, quality);
encoder.set_progressive(false);
encoder.encode(&frame_data, cols, rows, color_type)
```

So the label contradicts the codestream. PS3.5 Section 8.2.1 has Photometric Interpretation describe the color space of the *encoded* data, and for JPEG Baseline carrying YCbCr that value is `YBR_FULL_422`.

## Reproducing it

```rust
use dicom_object::open_file;
use dicom_pixeldata::Transcode;
use dicom_transfer_syntax_registry::{TransferSyntaxIndex, TransferSyntaxRegistry};

let mut obj = open_file("rgb_secondary_capture.dcm")?;   // RGB, 8 bits, uncompressed
let ts = TransferSyntaxRegistry
    .get("1.2.840.10008.1.2.4.50")
    .unwrap();
obj.transcode(ts)?;
obj.write_to_file("out.dcm")?;
```

## Suggested fix

In the encoder's attribute changes, write `YBR_FULL_422` rather than `RGB` when `samples_per_pixel == 3`, since that is what the stream holds:

```rust
} else if samples_per_pixel == 3 {
    // jpeg-encoder converts RGB input to YCbCr and subsamples the chroma
    if pmi != Some("YBR_FULL_422") {
        changes.push(AttributeOp::new(
            Tag(0x0028, 0x0004),
            AttributeAction::SetStr("YBR_FULL_422".into()),
        ));
    }
}
```

The alternative, if keeping `RGB` is preferred, would be to stop the color conversion in the encoder, `jpeg-encoder` can be given the components untransformed, and to write an Adobe APP14 marker with transform 0 so readers know. Either is fine; the two just have to agree.

userepo:
Fixes #829.

- When encoding a 3-sample-per-pixel object to JPEG Baseline, the frame is handed to `jpeg-encoder` as `ColorType::Rgb`, which converts the samples to YCbCr and writes a JFIF stream. The attribute changes returned by `encode_frame` nevertheless asked for (0028,0004) Photometric Interpretation to be `RGB`. Request `YBR_FULL_422` instead, since that is the color space of the encoded data.
- `RGB` is not an invalid value here — Table 8.2.1-1 in PS3.5 Section 8.2.1 lists both `YBR_FULL_422` and `RGB` for the 3-sample row of JPEG Baseline. What the same section does require is that the Data Elements describing the Pixel Data encoding be consistent with the characteristics of the compressed data stream, and a `RGB` tag over a YCbCr codestream is not.

### DIFF
--- a/pixeldata/src/transcode.rs
+++ b/pixeldata/src/transcode.rs
@@ -525,6 +525,16 @@ mod tests {
         // check transfer syntax
         assert_eq!(obj.meta().transfer_syntax(), JPEG_BASELINE.uid());
 
+        // check that the photometric interpretation
+        // describes the encoded data, which is YCbCr
+        assert_eq!(
+            obj.get(tags::PHOTOMETRIC_INTERPRETATION)
+                .expect("Photometric Interpretation should be present")
+                .to_str()
+                .unwrap(),
+            "YBR_FULL_422",
+        );
+
         // check that the pixel data is encapsulated
         // and has the expected number of fragments
         let pixel_data = obj.get(tags::PIXEL_DATA).unwrap();
--- a/transfer-syntax-registry/src/adapters/jpeg.rs
+++ b/transfer-syntax-registry/src/adapters/jpeg.rs
@@ -341,12 +341,14 @@ impl PixelDataWriter for JpegAdapter {
                 ));
             }
         } else if samples_per_pixel == 3 {
-            // set Photometric Interpretation to RGB
-            // if it was not already set to RGB
-            if pmi != Some("RGB") {
+            // set Photometric Interpretation to YBR_FULL_422:
+            // `jpeg-encoder` converts the RGB samples to YCbCr,
+            // and the tag must be consistent with the compressed stream
+            // (PS3.5 Section 8.2.1)
+            if pmi != Some("YBR_FULL_422") {
                 changes.push(AttributeOp::new(
                     Tag(0x0028, 0x0004),
-                    AttributeAction::SetStr("RGB".into()),
+                    AttributeAction::SetStr("YBR_FULL_422".into()),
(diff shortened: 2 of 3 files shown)


Enet4 (approved):
Looks OK. Thank you! :+1:

